PR #1, opened 2015-07-01
Review cf 20150726

vext01:
Stuff pointed by cfbolz and some other bits.

The commit "Use 10e-9..." is bogus and was fixed later.


iterations_runners/iterations_runner.java
```diff
@@ -0,0 +1,49 @@
+// XXX nanotime is actually CLOCK_MONOTONIC on linux, not MONOTONIC_RAW!
```
cfbolz: is this XXX still relevant?

PR #6, opened 2015-07-03
Portautils integration

vext01:
Integrates portautils into krun. Namely:

 * Add the ability to include an audit in the output (e.g. system info, package list...)
 * Check CPU is running in performance mode.
 * Wait for CPU to cool after each run.


krun/cpu.py
```diff
@@ -0,0 +1,117 @@
...
+class CPUNotThrottledError(Exception):
+    pass
+
+def check_cpu_throttled():
+    print("Checking CPU is throttled")
+    p = detect_platform()
+    if p == PLATFORM_LINUX:
+        check_cpu_throttled_linux()
+    else:
+        raise NotImplementedError("I don't know how to check your CPU")
```
ltratt: Given this error will be seen by end users, wouldn't it be better to do something like
```python
  sys.stderr.write("No support for checking CPU on this platform. Aborting.")
  sys.exit(0)
```

krun/cpu.py
```diff
@@ -0,0 +1,117 @@
...
+
+def check_cpu_throttled_linux():
+    with open(LINUX_GOV_FILE, "r") as fh:
+        v = fh.read().strip()
+
+    if v != "performance":
+        raise CPUNotThrottledError(("Expected 'performance' got '%s'. "
+                                    "Use cpufreq-set from the "
+                                    "cpufrequtils package") % v)
```
ltratt: Given this error will be seen by end users, wouldn't it be better to do something like
```python
  sys.stderr.write("Expected ... Use cpufreq-set ...")
  sys.exit(1)
```

(where ... is the full message from the exception).

krun/detect.py
```diff
@@ -0,0 +1,29 @@
+# Tools for detecting platform and OS.
+
+import sys
+import os
+
+
+# Note the subtle difference between OS and platform
```
ltratt: This is an unhelpful comment, since it doesn't say why there's a difference.

krun/cpu.py
vext01: OK, but we should exit non-zero.

sdaudit
```diff
@@ -0,0 +1,79 @@
...
+DFL_OUTFILE = "sdaudit.txt"
+
+from krun.detect import detect_os, OS_DEBIAN
+
+
+def run_cmd(cmd):
+    p = Popen(cmd, shell=True, stdout=PIPE)
+    stdout, stderr = p.communicate()
+    rc = p.wait()
+    assert(rc == 0)
```
ltratt: `assert` isn't a function.

ltratt:
I think this is going in the right direction, but it's a very #ifdefy in spirit which makes it hard to understand what's going on.

One suggestion would be to have a file called platform.py (or, if you want to, perhaps a package) with a structure along the lines of:

```python
class Platform(object): # abstract class
  def check_cpu_isnt_throttled(self): raise NotImplemented()
  def set_base_cpu_temp(self): raise NotImplemented()
  def wait_until_cpu_cool(self): raise NotImplemented()

class Linux(Platform):
  def check_cpu_isnt_throttled(self):
    <do linux stuff>
  def set_base_cpu_temp(self):
    os.sleep(1)
    self.base_cpu_temp = self.read_linux_cpu_temp()
  def wait_until_cpu_cool(self):
    os.sleep(1)
    while self.read_linux_cpu_temp() > self.base_cpu_temp:
      os.sleep(1)
```

and so on. Then, one can get a platform object and pass it around.

I know this might seem a bit icky in one way -- the platform class will have some fairly unrelated stuff in it -- but it will make porting to other platforms much easier. It also means that one can do things like have a Debian subclass of Linux if one wants to print out specific messages just for that Linux variant.

krun/cpu.py
```diff
@@ -0,0 +1,117 @@
...
+        time.sleep(BasePlatform.CPU_TEMP_MANDATORY_WAIT)
+        msg_shown = False
+        while True:
+            cool, reason = self.has_cpu_cooled()
+            if cool:
+                break
+
+            # if we get here, CPU is too hot!
+            if not msg_shown:
+                print("CPU is running hot.")
```
ltratt: I don't see a problem with the CPU being temporarily hot -- presumably the problem is if it's in meltdown mode and/or after several seconds (30? 60? dunno...) it hasn't cooled down? We might want to be careful about scaring the user until we're really sure something's gone wrong.

krun/cpu.py
```diff
@@ -0,0 +1,117 @@
...
+                    (self.initial_readings[i], readings[i])
+                return (False, reason)  # one or more sensor too hot
+        return (True, None)
+
+    def check_cpu_throttled(self):
+        with open(LinuxPlatform.CPU_GOV_FILE, "r") as fh:
+            v = fh.read().strip()
+
+        if v != "performance":
+            fatal(("Expected 'performance' got '%s'. "
```
ltratt: Double brackets are unnecessary I think?

krun/cpu.py
```diff
@@ -0,0 +1,117 @@
...
+    def collect_audit(self):
+        LinuxPlatform.collect_audit(self)
+        self.audit["packages"] = collect_cmd_output("dpkg-query -l")
+        self.audit["debian_version"] = collect_cmd_output("cat /etc/debian_version")
+
+def platform():
+    if os.path.exists("/etc/debian_version"):
+        return DebianLinuxPlatform()
+    else:
+        fatal("I don't have support for your platform")
```
ltratt: Can we make this a little less... personal :) Something like ```fatal("Platform '%s' currently unsupported" % platform_name)``` perhaps?

krun/platform.py
```diff
@@ -0,0 +1,129 @@
...
+        time.sleep(BasePlatform.CPU_TEMP_MANDATORY_WAIT)
+        msg_shown = False
+        trys = 0
+        while True:
+            cool, reason = self.has_cpu_cooled()
+            if cool:
+                break
+
+            # if we get here, CPU is too hot!
+            if not msg_shown:
```
ltratt: Shouldn't this message only be shown if trys is exceeded?

krun/platform.py
```diff
@@ -0,0 +1,129 @@
...
+    def _read_zone(self, zone):
+        fn = os.path.join(LinuxPlatform.THERMAL_BASE, zone, "temp")
+        with open(fn, "r") as fh:
+            return int(fh.read())
+
+    def take_cpu_temp_readings(self):
+        return [self._read_zone(z) for z in self.zones]
+
+    def has_cpu_cooled(self):
+        """returns tuple: cool * str_reason_if_false"""
```
ltratt: "*" -> ","?

vext01: As in cartesian product.

krun/platform.py
vext01: This message is shown when we start waiting for longer than a second. Intuition is that the user may thing krun has hung. Feedback let's them know this is supposed to happen. There is another message when we wait too long too.

krun/platform.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if self.initial_readings is None:
```
ltratt: This can't ever happen can it? Or, if it did, a base class messed things up badly. Maybe just "assert self.initial_readings is not None"

vext01: The constructor no longer makes a reading straight away. I thought it possible that the user may wish to instantiate the instance and set the base readings separately. Note initial_readings is born in the LinuxPlatform constructor. I figured that other data structures besides a list may be more appropriate in other platforms.

vext01: I should rename this field to base_readings anyway. Mixed terminology and all...

krun/platform.py
ltratt: I don't think they need a message immediately. In fact, it's likely to scare them. I agree if the pause was a minute or two they might get paranoid, but not if it's (say) 10-30s. For processes running multiple days, they'll be somewhat patient :)

krun/platform.py
ltratt: What is `cool' then? I think this isn't a very helpful docstring as it stands. At least, I still don't get it.

ltratt:
Can we be careful about pushing new stuff that isn't related to the original pull request? It's now really difficult to review new things coming in, as the diff is (in essence) now an entire file.

krun/platform.py
```diff
@@ -0,0 +1,151 @@
...
+        return collect_cmd_output("dmesg").split("\n")
+
+    def check_dmesg_for_changes(self):
+        new_dmesg = self._collect_dmesg_lines()
+        lines = [x for x in difflib.unified_diff(
+            self.current_dmesg, new_dmesg, "old", "new", lineterm="")]
+
+        if lines:
+            # dmesg changed!
+            print("Dmesg seems to have changed! Diff follows:\n")
```
ltratt: dmesg should be lower-case.

krun/platform.py
ltratt: I think the new non-scary variant is fine.

vext01: Not sure what you mean. So it's OK as is?

My two pence: When I am using someone else's software and it appears to hang for >10s, I am already hitting ctrl+c and running it under ktrace. I think the least we can do is inform the user. I would also like to see if this ever fires.

PR #7, opened 2015-07-06
Collect dmesg changes and print them at the end as well.

vext01:
Not paying much attention to which fd we print errors on at this point, as I will be changing everything to use the logging API next.

Cheers

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -5,8 +5,9 @@ import os
 import sys
 import difflib
 from collections import OrderedDict
+from krun import ABS_TIME_FORMAT
 from krun.util import fatal, collect_cmd_output
-
+from time import localtime
 
 class BasePlatform(object):
     CPU_TEMP_MANDATORY_WAIT = 1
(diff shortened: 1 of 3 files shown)


vext01:
Example output: https://gist.github.com/vext01/1dc50fcfd41ceb1882e5

krun/platform.py
```diff
@@ -21,21 +21,49 @@ def __init__(self):
         # errors and warnings in the Linux dmesg. If that happens, we
         # really want to know about it!
         self.current_dmesg = self._collect_dmesg_lines()
+        self.current_dmesg_time = time.time()
```
ltratt: Why use time() here but localtime() later.

ltratt: I'm also not sure `current_dmesg_time` is a good name: shouldn't it be something like `last_dmesg_time`?

ltratt:
The commit title is wrong ("aswell" instead of "as well") -- a force push of the commit should fix this.

ltratt:
Are all the dmesg changes printed at the end of a run too? [If things scroll off the top of the screen, people won't notice. I also wonder if problems like this should be emailed to a specified address(es) so that people don't need to continually keep an eye on the console?]

krun/platform.py
vext01: former: I'm using it to convert to local time, but I see a workaround. (time api in python is awful).

latter: as you wish

vext01:
aswell -> as well: this really doesn't matter. If you absolutely insist, one we agree the changes are ok,  I will merge by and and amend the commit. We should never force push.

vext01:
dmesgs are printed at the end. yes.

PR #8, opened 2015-07-06
Use std lib logging API for printing runner messages, warnings and errors.

vext01:
Note I was able to remove some messages altogether since the logger is already printing the current time.

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -7,17 +7,17 @@ usage: runner.py <config_file.krun>
 """
 
 import os, sys, json, time
+import logging
 from collections import deque
 import datetime
 import resource
 from subprocess import Popen, PIPE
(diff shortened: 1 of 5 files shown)


krun.py
```diff
@@ -319,14 +296,49 @@ def main():
...
+                                  job.key)
         one_exec_scheduled = True
 
     sched.run() # does the benchmarking
 
-    print("Time now is %s" % datetime.datetime.now().strftime(ABS_TIME_FORMAT))
+    info("Time now is %s" % datetime.datetime.now().strftime(ABS_TIME_FORMAT))
+
+def setup_logging():
+
```
ltratt: Extraneous blank line.

PR #10, opened 2015-07-07
Tidy up a bit.

vext01:
Misc tidying and a couple of bug fixes. Most of this found by static analyser.

On a related note: I've been using this particular analyser (flake8 + vim plugin) for a year or so on my own projects. It reports many other cosmetic changes (long lines, whitespace, etc.) that we could apply, however I have restrained since we don't really have a coding style in place.

Since we are now in the business of improving how we do software engineering, Is there any interest in adopting a coding style? PEP8 is the standard for Python code, and most static analysers and IDEs support it out the box. I would be willing to give it a shot.

If there is interest in this, we should not, however, enforce this on external projects.

Cheers

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -11,7 +11,6 @@ import logging
 from collections import deque
 import datetime
 import resource
-from subprocess import Popen, PIPE
 from logging import warn, info, error, debug
 
 import krun.util as util
@@ -34,7 +33,7 @@ def dump_json(config_file, out_file, all_results, audit):
     with open(config_file, "r") as f:
         config_text = f.read()
 
-    to_write = {"config" : config_text, "data" : all_results, "audit": audit}
+    to_write = {"config": config_text, "data": all_results, "audit": audit}
 
     with open(out_file, "w") as f:
         f.write(json.dumps(to_write, indent=1, sort_keys=True))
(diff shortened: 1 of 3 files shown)


ltratt:
I am all for using widely adopted standards (provided they're not completely obnoxious).

krun.py
```diff
@@ -232,7 +229,7 @@ def run(self):
 
         info("Done: Results dumped to %s" % self.out_file)
         if errors:
-            warn("%s ERRORS OCCURRED! READ THE LOG!%s" % (ANSI_RED, ANSI_RESET))
+            warn("Errors occurred! Read the scrollback!")
```
ltratt: What happens if errors exceed the scrollback? I thought the point was that we dumped these things to a log file and/or emailed them to a given address too?

vext01: We've not discussed that, but I don't object. In fact I think it's a good idea.

Let's dump whatever goes via the logging API to a date-stamped file and mail errors as they occur? The latter should be controlled by a config option. Sounds good?

vext01: If you agree, merge this and I will make a new pull (eventually, I want to spend time on other tasks for a bit now).

ltratt: Agreed!

vext01:
On Tue, Jul 07, 2015 at 05:19:36AM -0700, Laurence Tratt wrote:
> Merged #10.

Thanks.

-- 
Best Regards
Edd Barrett

http://www.theunixzoo.co.uk

vext01:
RE: standards.

PEP8 is indeed widely adopted. It is mostly pretty good. It has a couple of warts, but not many. We should ask the others what they think.

If you get bored have a play with this:
https://github.com/nvie/vim-flake8

PR #12, opened 2015-07-08
Log to file. Mail errors and the completion message.

### LINKED ISSUE #11: Log to file and email errors.
As discussed with @ltratt , it should be possible to log all benchmarking output to a file and to (optionally) email errors to a configurable address.

vext01:
* Log messages to file.
* Mail errors as they occur.
  * But send no more than 5 errors per hour.
* Mail us when benchmarking is complete.
* Capture stderr of iterations_runner so we can include stderr in error reports.
* Prompt to proceed if mail addresses are set.
  * Prevents accidental spamming.

Fixes #11

Quite a big change, so I expect you will find a few bits.

## Example Mails

```
Date: Wed, 08 Jul 2015 11:18:28 +0100                                                                                                                           
From: noreply@bencher3.dcs.kcl.ac.uk                                                                                                                            
To: vext01@localhost                                                                                                                                            
Subject: [krun] Benchmark failure: binarytrees:PyPy:default-python                                                                                              
                                                                                                                                                                
Message from krun running on bencher3.dcs.kcl.ac.uk:                                                                                                            
                                                                                                                                                                
########################################################################                                                                                        
Benchmark didn't emit a parsable list on stdout.                                                                                                                
stdout:                                                                                                                                                         
                                                                                                                                                                
stderr:                                                                                                                                                         
  File "<builtin>/app_main.py", line 75, in run_toplevel                                                                                                        
  File "/home/vext01/research/warmup_experiment/krun/iterations_runners/iterations_runner.py", line 31, in <module>                                             
    bench_mod = imp.load_source(bench_mod_name, benchmark)                                                                                                      
IOError: [Errno 2] No such file or directory                                                                                                                    
                                                                                                                                                                
########################################################################                                                                                        
```

```
Date: Wed, 08 Jul 2015 11:18:28 +0100                                                                                                                           
From: noreply@bencher3.dcs.kcl.ac.uk                                                                                                                            
To: vext01@localhost                                                                                                                                            
Subject: [krun] dmesg changed                                                                                                                                   
                                                                                                                                                                
Message from krun running on bencher3.dcs.kcl.ac.uk:                                                                                                            
                                                                                                                                                                
########################################################################                                                                                        
dmesg seems to have changed! Diff follows:                                                                                                                      
--- 2015-07-08 11:18:27                                                                                                                                         
+++ 2015-07-08 11:18:28                                                                                                                                         
@@ -888,3 +888,4 @@                                                                                                                                             
 [2511438.416231] d8[8381]: segfault at 10 ip 0000000000496da7 sp 00007fffffffe200 error 4 in d8[400000+12b7000]                                                
 [2511482.538337] d8[8537]: segfault at 10 ip 0000000000496da7 sp 00007fffffffe200 error 4 in d8[400000+12b7000]                                                
 [2542600.838511] Clock: inserting leap second 23:59:60 UTC                                                                                                     
+lkkljsflkjdaslkfjdalkfj                                                                                                                                        
########################################################################                                                                                        
```


krun.py
```diff
@@ -14,13 +14,34 @@
...
 from krun.platform import platform
 from krun import ABS_TIME_FORMAT, UNKNOWN_TIME_DELTA, UNKNOWN_ABS_TIME
+from krun.mail import Mailer
 
 BENCH_DRYRUN = os.environ.get("BENCH_DRYRUN", False)
 
 HERE = os.path.abspath(os.getcwd())
 
+
+COLOURS = True
```
ltratt: move line 27 to after the import, to be clearer about the logic?

krun.py
```diff
@@ -14,13 +14,34 @@
...
+    COLOURS = False
+
+COLOUR_FORMATTER = colorlog.ColoredFormatter(
+    "%(log_color)s[%(asctime)s %(levelname)s] %(message)s%(reset)s",
+    ABS_TIME_FORMAT)
+PLAIN_FORMATTER = logging.Formatter(
+    '[%(asctime)s: %(levelname)s] %(message)s',
+    ABS_TIME_FORMAT)
+
+if COLOURS:
```
ltratt: The COLOURS and CONSOLE_FORMATTER variables seem redundant in pairs. Remove COLOURS and merge this into the try/except maybe?

krun.py
```diff
@@ -92,20 +113,24 @@ def run(self):
...
             iterations_results = eval(stdout) # we should get a list of floats
         except SyntaxError:
-            error("=ERROR=" * 8)
-            error("*error: benchmark didn't error a parsable list.")
-            error("We got:\n---\n%s\n---\n" % stdout)
-            error("To see the invokation set the BENCH_DEBUG env and run again")
-            error("=ERROR=" * 8)
-
+            rule = 50 * "-"
+            err_s = "Benchmark didn't emit a parsable list on stdout.\n"
```
ltratt: Use a triple quoted string rather than +=.

krun.py
```diff
@@ -229,9 +256,12 @@ def run(self):
 
         info("Done: Results dumped to %s" % self.out_file)
         if errors:
-            warn("Errors occurred! Read the scrollback!")
+            warn("Errors occurred! Read the log!")
```
ltratt: Too! Many! Explanation! Marks! How about "Errors occurred -- read the log!"

krun.py
```diff
@@ -277,9 +307,27 @@ def main():
     config = util.read_config(config_file)
     out_file = util.output_name(config_file)
 
+    mail_recipients = config.get("MAIL_TO", [])
```
ltratt: I think MAIL_TO should be optional, so if it doesn't have a value, don't send mails. [Though I suppose an empty list is equivalent...]

krun/mail.py
```diff
@@ -0,0 +1,86 @@
+from collections import deque
+from email.mime.text import MIMEText
+from datetime import timedelta, datetime
+import smtplib
+import socket
+import textwrap
+
+
+class Mailer(object):
+
```
ltratt: Extraneous blank line.

krun/mail.py
```diff
@@ -0,0 +1,86 @@
...
+    FROM_USER = "noreply"
+    SMTP_HOST = "localhost"
+    WRAP_THRESHOLD = 72
+    RULE = WRAP_THRESHOLD * "#"
+
+    QUOTA_THRESHOLD_TEMPLATE = \
```
ltratt: Why are these class level constants? Surely they're better at the module level?

krun/mail.py
```diff
@@ -0,0 +1,86 @@
...
+    def set_recipients(self, recipients):
+        self.recipients = recipients
+
+    def _wrap_para(self, txt):
+        return "\n".join(textwrap.wrap(txt, Mailer.WRAP_THRESHOLD))
+
+    def _one_hour_ago(self):
+        return datetime.now() - timedelta(hours=1)
+
+    def _clear_old(self):
```
ltratt: I don't understand the logic here (I'm not saying it's wrong, but it is obscure). At least a comment is needed.

krun/mail.py
```diff
@@ -0,0 +1,86 @@
...
+        one_hour_ago = self._one_hour_ago()
+        while True:
+            try:
+                t = self.last_mail_times.popleft()
+            except IndexError:
+                break  # nothing in deque
+            if t > one_hour_ago:
+                self.last_mail_times.appendleft(t)  # put back
```
ltratt: If I understand this correctly (and I might not...), this is still going to send *all* mails, but just throttle them to 5 an hour. If so, I think this is undesireable -- I could still wake up to 50 mails. I think it should say something like "after 5 mails in an hour, I will throw away all further mails". After all, after 5 errors, one probably has the idea that the experiment is screwed... The complete number of errors is still in the log on the server right?

krun.py
vext01: PLAIN_FORMATTER is used unconditionally for logging to a file later (see line 373), otherwise I would have done exactly as you suggested.

krun.py
vext01: I did start out with a triple quoted string, but to achieve the desired effect I would have to mess up the indentation of the file (the triple quoted text would have to be up against the left margin). What do you think?

krun.py
vext01: It is indeed optional. If the directive is missing in the config file, then it defaults to `[]`, which means mail no one. See mailer.py:55.

krun/mail.py
vext01: Actually it is wrong. I misinterpreted your comment on the original bug report :)

krun/mail.py
vext01: You understand correctly.

I would go one step further and say "after 5 mails <del>in an hour</del>, I will throw away all further mails". 

What do you think?

vext01:
Branch updated:

 * Remove time from mail quota. There is a simple numeric threshold on the number of mails we will send per run.
 * Threshold now configurable via config file.

krun.py
```diff
@@ -277,9 +307,32 @@ def main():
...
+    if type(mail_recipients) is not list:
+        fatal("MAIL_TO config should be a list")
+
+    max_mails = config.get("MAX_MAILS", 5)
+
+    # You should only have mailing for real experiments.
+    # Prevent accidental sending of mail during benchmark implementation
```
ltratt: I don't get the point of this if clause. If the user specifies an email address, let's just trust them. Most users hate pressing multiple Y/N stuff, which also prevents automation.

vext01:
Eventually we should also use the sendmail executable and not smtp #15.

krun.py
vext01: Just trying to avoid accidentally spamming people. Will remove if you are sure.

ltratt: If you want to add a "don't send email switch" (-n or whatever), that would be better I think. But I don't think that's even necessary. At worst people are going to get 5 mails... and if they don't want any, they can simply not specify a mail address.

vext01:
* Move class constants to module scope.
 * Send a mail as benchmarks start.
 * Remove anti-spam question.

krun.py
```diff
@@ -277,9 +314,27 @@ def main():
...
+    if mail_recipients:
```
ltratt: This clause is still redundant. The user knows it's going to send mail N times, because they told it ito. Let's not repeat the sins of some Unix's dmesg ;)

ltratt:
OK, squash those three commits down (including the merge commit) and we'll be ready to merge.

Issue #14, opened 2015-07-28
Add sanity checks for VMs


vext01:
As discussed with Chris Seaton, some VMs need checks prior to experimentation.

E.g. for graal, we need to check `java.vm.name` for "JVMCI".

Similarly, for JRuby we should check `Truffle.graal?` is true.

From experience, it's easy to get these wrong.

vext01:
This is on hold. The version of graal I have seems to have a broken jvmci bootstrap. I cannot update as Oracle seem to have moved/restructured  their hg repo, so build.sh no longer works.

Have mailed Chris asking if they can update the build instructions.

vext01:
WIP work is in fix_oracle_stuff branch.

PR #18, opened 2015-07-29
Check things exist

vext01:
* check all VM executables exist.
 * check all neccessary benchmark files exist.

Avoids cases where benchmarks run for some time, then crash out later because of typos in the config file. Better to check sooner rather than later.

Was going to implement this as part of the fix for #14, but since this is blocked, decided to split this unit of work out.

Fixes #18.

OK?


vext01:
I've kept the commits self-contained, so no need to squash this time.

PR #23, opened 2015-08-06
Fix oracle VMs.

### LINKED ISSUE #5: vm_defs.py
It works but I'm unhappy with it:

 * Handling of vm arguments and environments is a mess.
 * Updating environment dicts with dict.update() may flatten an existing entry in the environment.

Some of the issues regarding the former are due to the fact that the heap limit is not known when the VMDef is instantiated. I don't want the user to have to pass that in explicitly. I guess the proper fix is to make a proper config file format which is not evaluated by the python interpreter. Then after the file is parsed, the VMDefs could be instantiated passing in whatever we want.

To properly handle the latter we would need to provide krun with information about the environment. E.g. to properly update PATH without flattening the existing $PATH, krun would need to know that it is a colon separated collection of paths, and that to append another path, it should append ":%s" % another_path

These should be fixed one day perhaps, but we have bigger fish to fry at the moment.

### LINKED ISSUE #14: Add sanity checks for VMs

vext01:
Includes sanity checks that run prior to real benchmarking. These are
checking that graal and truffle and enabled.

Environment handling also improved. (Sorry this is mingled in with the other change. In the beginning, fixing the oracle vms depended upon this, but I found a simpler way to invoke the VMs which didn't need the environment cleanups. Nevertheless, I am happy that handling has been simplified.)

Fix #14. Fix #5.

### DIFF
--- a/krun/vm_defs.py
+++ b/krun/vm_defs.py
@@ -24,28 +24,33 @@ class EnvChange(object):
     def __init__(self, var, val):
         self.var, self.val = var, val
 
+
+    @staticmethod
+    def apply_all(changes, env):
+        """Apply a collection of changes"""
+        for change in changes:
+            change.apply(env)
+
+    def apply(self, env):
+        raise NotImplementedError("abstract")
+
+
 class EnvChangeSet(EnvChange):
-    pass
+    def apply(self, env):
+        cur_val = env.get(self.var, None)
+        if cur_val is not None:
+            fatal("Environment %s is already defined" % self.var)
+        else:
+            env[self.var] = self.val
+
 
 class EnvChangeAppend(EnvChange):
-    pass
-
-
-def _apply_env_changes(env, changes):
-    for change in changes:
-        cur_val = env.get(change.var, None)
-        if isinstance(change, EnvChangeSet):
-            if cur_val is not None:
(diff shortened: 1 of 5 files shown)


krun/vm_defs.py
```diff
@@ -15,26 +19,63 @@
...
+        self.var, self.val = var, val
+
+class EnvChangeSet(EnvChange):
+    pass
+
+class EnvChangeAppend(EnvChange):
+    pass
+
+
+def _apply_env_changes(env, changes):
```
cfbolz: it looks to me like it would be much more natural to make these changes be methods on the EnvChange classes, instead of the isinstance checks here.

Also, I think there needs to be a docstring in some of these classes to explain what they are for.

vext01: I tend to agree. Will fix.

vext01:
Pushed a fix.

cfbolz:
looks good to me!

PR #25, opened 2015-08-10
Add missing sanity checks

vext01:
Looks like the actual implementations of the sanity checks went missing somewhere along the line. Here they are.


vext01:
"Looks good, go ahead" -- cfbolz

PR #27, opened 2015-08-12
Don't sanity check VMs that won't be used.

vext01:
Helps me, when I don't have a full set of VMs built.

Comment in the code explains in detail.


ltratt:
Looks good.

PR #32, opened 2015-08-24
Scaler and save power

### LINKED ISSUE #30: Eco friendly krun
We should make our best attempt to set the powersave governor after benchmarking is done.

Recommend using sudo to crank up/down the governor (if possible).

Discussed with @ltratt.

vext01:
* Check the CPU scaler is *not* pstate.
* Save power when done.

OK?

Fixes #30.


ltratt:
Maybe needs a bit of squashing?

vext01:
Clean squish. What do you think?

ltratt:
Yep, looks mergeable!

PR #33, opened 2015-08-25
Support for pinning benchmarks to an isolated CPU core.

vext01:
Took a few attempts to get right, due to Linux intricacies and bugs. Please read comments carefully.

Includes small refactorings where a method became large.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -140,6 +140,24 @@ class LinuxPlatform(BasePlatform):
         BasePlatform.__init__(self, mailer)
 
 
+    def _fatal_kernel_arg(self, arg, prefix, suffix):
+        """Bail out and inform user how to add a kernel argument"""
+
+        # This is generic Linux advice.
+        # If you can offer better distribution-specific advice, override this
+        # in a more specific Linux subclass.
+
+        if prefix != "":
+            prefix += "\n"
+
+        if suffix != "":
+            suffix += "\n"
+
+        fatal("%s"
+              "Set `%s` in the kernel arguments.\n"
+              "%s" % (prefix, arg, suffix))
+
+
     def bench_cmdline_adjust(self, args):
         """Adjusts benchmark invocation so as to pin to one CPU core"""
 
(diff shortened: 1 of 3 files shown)


krun/platform.py
```diff
@@ -191,9 +213,69 @@ def save_power(self):
...
+                  "Did you add `isolcpus=X` to the kernel arguments?\n"
+                  "To do this on Debian:\n"
+                  "  * Edit /etc/default/grub\n"
+                  "  * Add the argument to GRUB_CMDLINE_LINUX_DEFAULT\n"
+                  "  * Run `sudo update-grub`\n"
+                  "When the system comes up, check `ps -Pef`.")
 
-    def check_cpus_throttled(self):
-        """Checks the CPU is configured for high performance
+        if isol_cpu == 0:
```
ltratt: If it detects CPU 0 then throwing a fatal error seems a bit harsh. Can it not try again and see if another non-CPU0 is possible? Otherwise I wonder if on a 4 core machine about 1/4 of executions might hit this?

vext01: This will only happen if the user actively chose to include CPU0 in the set of cores to isolate. The user chooses which cores by passing special arguments to the kernel at boot time via the bootloader, so it's really in the user's hands to make sure they don't isolate CPU 0.

[If the odds of choosing CPU 0 on a 4 CPU machine are 1/4, this means the user chose to isolate all four CPUs, thus not allowing anywhere for any non-benchmarking processes to run.]

vext01: I've just noticed your wording of "about 1/4 of executions might hit this". All executions run on the same core. The core used is selected once prior to benchmarking.

ltratt: Maybe the error should tell them what the config option to exclude core 0 is?

vext01: We can do, but if they got this far, they already managed to isolate a CPU, just not the right one. They just do the same as before replacing `0` with another integer.

Thoughts?

vext01: In retrospect, it could have been another member of the team who set up the initial isolation. So I actually agree. Bear with me.

vext01:
Last commit fixes this and makes a general method for helping the user to add kernel arguments. Prevents code duplication.

I'm happy with the structure of the commits as-is. OK?

ltratt:
Maybe squash a bit? Otherwise seems ready for commit.

vext01:
Force push fixes typo. OK?

vext01:
Read previous comment about squashing.

ltratt:
The first two commits look sensibly distinct; the last two I'm less sure of. There again, I don't care *that* much. Your call.

vext01:
Ah right. Well, I kept the bug comment separate purposefully. The last commit changes code outside of what I touched in the three preceding commits (e.g. pstate checks), so I would rather not mix them.

ltratt:
OK.

PR #37, opened 2015-08-27
Resilience

vext01:
Make krun more resilient in the face of crashing benchmarks.

I'm happy with the commit structure.

OK?

### DIFF
--- a/iterations_runners/iterations_runner.java
+++ b/iterations_runners/iterations_runner.java
@@ -44,7 +44,11 @@ class IterationsRunner {
             double stopTime = IterationsRunner.JNI_clock_gettime_monotonic();
 
             double intvl = (stopTime - startTime);
-            System.out.print(intvl + ", ");
+            System.out.print(intvl);
+
+            if (i < iterations - 1) {
+                System.out.print(", ");
+            }
         }
         System.out.print("]");
 
--- a/iterations_runners/iterations_runner.js
+++ b/iterations_runners/iterations_runner.js
@@ -19,6 +19,8 @@ for (BM_i = 0; BM_i < BM_n_iters; BM_i++) {
 
 	var BM_intvl = BM_stop_time - BM_start_time;
 	print(BM_intvl);
-	print(", ")
+    if (BM_i < BM_n_iters - 1) {
+        print(", ")
+    }
 }
 print("]");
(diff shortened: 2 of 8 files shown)


vext01:
Squashed last two commits, as requested on irc.

vext01:
Ready.

vext01:
Not ready, we both posted an once.

vext01:
WRT eval. We can certainly parse the output, although: a) i don't think security matters much for a benchmark runner, and b)  bear in mind the config file reading is pretty much a big eval also.

I've wanted to re-write the way configuration works for some time, but it's not a small task and too many other things are higher priority.

cfbolz:
If it's literally a list of floats, just use ``json.loads(stdout)``

ltratt:
It's my fault for not spotting the `eval` earlier, but it has to be removed.

vext01:
What about the config file?
https://github.com/softdevteam/krun/blob/master/krun/util.py#L31

Let's at-least raise an issue for this and I can look at it after the pyhyp deadline. Juggling too many things at once.

ltratt:
The config file is fine as an `eval`, because the user should know what they're doing (and, if they don't, bad luck) -- and the flexibility can be actively useful. The problem with parsing random output from a program is that it might accidentally produce incorrect output which is `eval`able and changes the program state.

vext01:
Try this.

ltratt:
I think the last 3 commits should be squashed (it doesn't make sense to have individual commits which can't work on their own). Then I think we're probably ready to merge (although maybe it's worth removing the "eval syntax error" commit entirely?).

vext01:
I was unable to cleanly pull out the syntaxerror exception commit, as future commits depend upon it.

Squashed the JSON related commits into one, but since the indent fixes are on lines untouched by the json fix, I think it should remain separate. Would you agree?

ltratt:
Since you were editing that file anyway, I think touching up non-critical indents might as well be folded in.

vext01:
OK squashed. Note that I have gotten in trouble in the past (albeit outside of softdev) for mushing style changes with functional changes. One to bear in mind.

OK to merge?

ltratt:
Minor whitespace fixes of a few lines are pretty common IME, and I don't see a problem in bundling them up. Of course, there's an element of judgement in this.

Anyway, this looks ready for merging. Good stuff!

PR #39, opened 2015-08-27
Unbuffered stderr

vext01:
Print stderr as it arrives. As discussed on IRC.

OK?

### DIFF
--- a/iterations_runners/iterations_runner.rb
+++ b/iterations_runners/iterations_runner.rb
@@ -43,7 +43,7 @@ if __FILE__ == $0
         intvl = stop_time - start_time
         STDOUT.write String(intvl)
         if i < iters - 1 then
-            STDOUT.Write ", "
+            STDOUT.write ", "
         end
     end
     STDOUT.write "]"
--- a/krun/vm_defs.py
+++ b/krun/vm_defs.py
@@ -1,5 +1,6 @@
 import subprocess
 import os
+import select
 
 from logging import info, debug
 from krun import EntryPoint
@@ -10,6 +11,17 @@ ITERATIONS_RUNNER_DIR = os.path.abspath(os.path.join(DIR, "..", "iterations_runn
 BENCHMARKS_DIR = os.path.abspath(os.path.join(os.getcwd(), "benchmarks"))
 VM_SANITY_CHECKS_DIR = os.path.join(DIR, "..", "vm_sanity_checks")
 SANITY_CHECK_HEAP_KB = 1024 * 1024  # 1GB
+
+# Pipe buffer sizes vary. I've seen reports on the Internet ranging from a
+# page size (Linux pre-2.6.11) to 64K (Linux in 2015). Ideally we would
+# query the pipe for its capacity using F_GETPIPE_SZ, but this is a) not
+# portable between UNIXs even, and b) not exposed by Python's fcntl(). For
+# now, we use a "reasonable" buffer size. If it is larger than the pipe
+# capacity, then no harm done; if it is smaller, then we may do more reads
+# than are strictly necessary. In either case we are safe and correct.
+PIPE_BUF_SZ = 1024 * 16
+
+SELECT_TIMEOUT = 1.0
 
 
 # !!!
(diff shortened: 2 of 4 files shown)


iterations_runners/iterations_runner.rb
```diff
@@ -43,7 +43,7 @@ def assert(cond)
         intvl = stop_time - start_time
         STDOUT.write String(intvl)
         if i < iters - 1 then
-            STDOUT.Write ", "
+            STDOUT.write ", "
```
cfbolz: what is that difference?

krun/vm_defs.py
```diff
@@ -110,10 +110,30 @@ def _run_exec(self, args, heap_lim_k, bench_env_changes=None):
...
+            if line == "":
+                break  # EOF, child has exited
+
+            line = line.strip()
+            info(line)
+            stderr_lines.append(line)
+
+        stderr = "\n".join(stderr_lines)
+        stdout = p.stdout.read()
+        p.poll()  # sets returncode attr
```
cfbolz: I think the stdoud read and the poll should be switched. I wonder whether stderr.readline could also return "" if the process closes stderr.

krun/vm_defs.py
```diff
@@ -110,10 +110,30 @@ def _run_exec(self, args, heap_lim_k, bench_env_changes=None):
...
+
+        # Should not fire. The empty string that causes
+        # the above loop to exit is indicating EOF. Thus the
+        # child should be done.
+        assert p.returncode is not None
```
ltratt: I'm not sure this is correct -- you should almost certainly `wait` on the process before expecting it to definitely have a return code.

iterations_runners/iterations_runner.rb
vext01: lowercase `W`

krun/vm_defs.py
vext01: I guess the child could close it's stderr without exiting.

ltratt: Yes, and it will definitely happen at various points (child processes do all sorts of odd things).

krun/vm_defs.py
vext01: I understand you latter point, but the former about switching stuff around, can you explain a bit more? We need the readline() in a loop to that we get the lines as they appear. There's little we can do if stderr closes but the process keeps running. Perhaps we need to replace the poll() with something which will block.

ltratt: It doesn't matter in which order you do this: you have to interleave the reading of stdout and stderr. A process might block on (say) writing stdout waiting for stderr to be read. So you need to empty both of them concurrently.

cfbolz:
```
        p.poll()  # wait till process finishes
        stdout = p.stdout.read()
```

vext01: E.g. `Popen.wait()` https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait

ltratt: You have to empty stdout and stderr and then `wait`.

vext01:
Guys, I'm out of time, but I've pushed my effort thus far.

It works, but all of the stderr lines all arrive at once for some reason. This brings us back to where we started :(

vext01:
Laurie was right that this is fiddly.

vext01:
Take a look at the last commit. This seems to work really well. If it looks good, we can squash this lot into one commit.

Note that `fcntl()` is a UNIX-only interface, hence moving code that uses it into `platform.py`. Windows has its own platform specific mechanism. We don't care for windows for now, so meh.

Looks good?

vext01:
Fixed @ltratt comments, and a few other fixes.

Re: wait(). It can raise, but the documents don't say what. Reading the source I can see it at-least raises `OSError`. I've played it safe by catching a generic exception.

See comment about larger buffer size.

OK to squash?

ltratt:
I don't mind the `buffer` comment, but I don't think it's necessary (or, at least, not at that length). After all, it's simply a minor performance optimisation, and has nothing to do with correctness.

vext01:
Thinking about it, it doesn't matter if we don't over-approximate the buffer capacity. It just means we may do more reads than is necessary. Am I right?

ltratt:
Correct. 1KiB is too small; I think 16KiB is a decent compromise.

vext01:
OK, try this. I've reset the size and revised the comment reflecting the conversation we have had here.

OK to do the squisharoo?

ltratt:
Looks good to me. Squish away.

vext01:
Splat. Ready to merge.

PR #40, opened 2015-09-02
Fix ruby iterations runner progress reporting.

vext01:
A bug introduced in a recent change. Ruby runner was always reporting iteration 1/X.

### DIFF
--- a/iterations_runners/iterations_runner.rb
+++ b/iterations_runners/iterations_runner.rb
@@ -32,8 +32,7 @@ if __FILE__ == $0
     require("#{benchmark}")
 
     STDOUT.write "["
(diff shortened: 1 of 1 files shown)


ltratt:
In a sense, the real problem was that line 36 should have referenced `i` and not `krun_item`, right? Anyway, the patch is fine.

vext01:
Correct. I just preferred `krun_iter_num`. In some languages, I have used obscure names for the iterations runner code so as to avoid name clashes with benchmark code. This doesn't apply to ruby as it has decent scoping rules. I guess old(ish) habits die hard.

Let's merge this?

PR #43, opened 2015-09-28
A helper function to find the jvmci internal jdk8.

vext01:
OK?

### DIFF
--- a/krun/vm_defs.py
+++ b/krun/vm_defs.py
@@ -1,6 +1,7 @@
 import subprocess
 import os
 import select
+import fnmatch
 
 from logging import info, debug
 from krun import EntryPoint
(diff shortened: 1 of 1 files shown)


ltratt:
Can we give an idea of which directories? The comment is a bit vague...

vext01:
Made the function name more accurate and documented it more. OK?

krun/vm_defs.py
```diff
@@ -293,6 +294,30 @@ def check_benchmark_files(self, benchmark, entry_point):
...
+    or it could be:
+    jvmci/jdk1.8.0_66-internal/product/bin/java
+
+    This is a helper function to try and find the 'java' binary
+    inside this "moving" directory.
+
+    arguments:
+    base_dir -- base jvmci directory"""
+
+    for f in os.listdir(base_dir):
```
ltratt: Could this be `for f in fnmatch.filter(os.listdir(base_dir), "jdk1.8*internal"):`?

ltratt: Also `f` suggests "file"; maybe `l` for leaf or `p` for path?

vext01:
We can use `fnmatch.filter`, but we wouldn't use it in a loop. In doing so we can also check there is only one match, thus making this a bit more robust.

Try this. OK to squish?

ltratt:
I'm going to suggest we remove the pointless `assert`, squash, then merge. OK?

vext01:
Hey,

On Tue, Sep 29, 2015 at 03:31:25AM -0700, Laurence Tratt wrote:
> I'm going to suggest we remove the pointless `assert`, squash, then merge. OK?

The conditional is not exhaustive, so I would argue the assert is good
practice. (OK, so we should never have a list of length <0, but it doesn't
hurt and krun doesn't need to be high-performance.)

Equally I don't care much either. We are both nitpicking at this point :)

-- 
Best Regards
Edd Barrett

http://www.theunixzoo.co.uk

ltratt:
I am a fan of `assert`s in general, but I think we can go overboard. If the list has length less than 0, I'd be pretty worried ;) Let's kill it, squash, and then I can merge.

vext01:
splat.

vext01:
Don't forget the accompanying change here: https://github.com/softdevteam/warmup_experiment/pull/13

krun/vm_defs.py
```diff
@@ -293,6 +294,34 @@ def check_benchmark_files(self, benchmark, entry_point):
...
+    This is a helper function to try and find the 'java' binary
+    inside this "moving" directory.
+
+    arguments:
+    base_dir -- base jvmci directory"""
+
+    matches = fnmatch.filter(os.listdir(base_dir), 'jdk1.8.0*internal*')
+
+    if len(matches) == 1:
+            return os.path.join(base_dir, matches[0], "product", "bin", "java")
```
ltratt: Broken indentation?

ltratt:
I'm not sure if I missed the indentation before, but we should fix it before merging.

vext01:
go for it.

PR #44, opened 2015-09-29
fix krun if colorlog is not installed

cfbolz:
There was some code to keep krun working with ``colorlog`` not installed, but it was broken. This fixes it.


cfbolz:
oops, hold on, something else is wrong :-(

cfbolz:
ready for review, @vext01

vext01:
Looks good.

PR #45, opened 2015-10-01
Fix quadratic stderr

cfbolz:
it's not a huge problem, the quadraticness would only appear for really huge stderrs (that we don't have). But still. In this case it's probably slightly easier to look at the two commits independently, the first is a refactoring towards testability, the second one fixes the problem.

Note that the test of course passes in both cases, but in the quadratic implementation it would take a very long time.

### DIFF
--- a/krun/tests/test_process.py
+++ b/krun/tests/test_process.py
@@ -1,0 +1,25 @@
+from krun.vm_defs import print_stderr_linewise
+
+def test_quadratic():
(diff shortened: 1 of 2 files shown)


vext01:
Very nice. I didn't actually know about `send()` for a generator, so TIL.

Tried the test before and after the fix, and verified a huge performance improvement.

Thanks

PR #47, opened 2015-10-08
Trunc and compress

vext01:
Store results truncated to 6 DP, and store results bz2 compressed.

Tools that consume results will be tweaked separately.

Looks good?


ltratt:
Looks good!

PR #49, opened 2015-10-12
C support

vext01:
Add's C support and tidies up build system (a lot, including making Java optional).

OK?

### DIFF
--- a/iterations_runners/Makefile
+++ b/iterations_runners/Makefile
@@ -1,0 +1,28 @@
+PWD != pwd
+C_ITER_RUNNER_CFLAGS = -fPIC -Wall -Wextra -pedantic
+C_ITER_RUNNER_LDFLAGS =
+
+
+OS != uname -s
+ifeq ($(OS),Linux)
+	C_ITER_RUNNER_LDFLAGS += -ldl
+endif
+
+.PHONY: clean
+
+all: iterations_runner_c IterationsRunner.class BaseKrunEntry.class
+
+IterationsRunner.class BaseKrunEntry.class: iterations_runner.java
+	if [ "${ENABLE_JAVA}" = "1" ]; then \
+		${JAVAC} iterations_runner.java; \
+	fi
+
+iterations_runner_c: iterations_runner.c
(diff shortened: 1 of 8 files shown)


vext01:
(Sorry this is such a large change)

iterations_runners/Makefile
```diff
@@ -0,0 +1,26 @@
+HERE != pwd
```
ltratt: Can we use a better variable name? Or, preferably, simpy use `pwd` instead of `${HERE}` later?

ltratt:
This patch seems to use hard tabs everywhere, instead of 4 spaces. [Obviously in a Makefile, one has no choice. But everywhere else we should be using spaces.]

iterations_runners/Makefile
```diff
@@ -0,0 +1,26 @@
...
+C_ITER_RUNNER_CFLAGS = -fPIC -Wall -Wextra -pedantic -g
```
snim2: Compiling with `-g` probably isn't necessary for benchmarking.

vext01:
The iterations runner is in KNF.

ltratt:
Tabs are an editing disaster. Let's use 4 spaces, please.

iterations_runners/Makefile
ltratt: Good spot. IIRC, it actually causes minor slowdown (bigger files etc.). I have no idea whether it's a measurable slowdown, though.

vext01:
As you wish.

vext01:
Try this.

If OK, will do some minimal squashing.

ltratt:
Looks good to me!

snim2:
I think this is still compiling the iter-runner with `-g`?

vext01:
Nice catch. If the user wants symbols, they add `-g` to `CFLAGS`. E.g. `gmake CFLAGS=-g`.

snim2:
OK, lgtm. Do you want to squash any commits before I merge?

vext01:
Splat. I propose this commit structure.

vext01:
Thanks all!

PR #50, opened 2015-10-12
Add a "VM definition" for native code.

vext01:
Add a "VM definition" for native code benchmarks, e.g. C.

Required a little boilerplate hacking and fix a bad abstract signature while we are here.

OK?

### DIFF
--- a/krun/vm_defs.py
+++ b/krun/vm_defs.py
@@ -101,13 +101,21 @@ class BaseVMDef(object):
 
         self.platform = None  # Set later
 
+    def _get_benchmark_path(self, benchmark, entry_point, sanity_check=False):
+        if not sanity_check:
+            return os.path.join(BENCHMARKS_DIR, benchmark, entry_point.subdir,
+                                entry_point.target)
+        else:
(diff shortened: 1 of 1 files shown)


snim2:
Looks OK to me. 

This isn't in this PR, but reading through the surrounding code I was initially confused by `NotImplementedError("abstract")`, which I would usually write as something like `NotImplementedError("method must be overridden in subclass")` (if I've understood this correctly).

vext01:
It's an idiom I've picked up from PyPy.

In a sense the message is moot, as the line should never be executed ;)

I've just done some searching on the net, and apparently the "official" way is to use the `abc` module and `@abstractmethod`. Shall we try it?

snim2:
That sounds like a good idea to me. I see the relevant PEP (3119) was accepted in 2007, so I would guess that this exists in the PyPy stdlib too.

vext01:
OK, let's merge this, then I'll make a new PR doing the abstract stuff.

vext01:
thanks!

PR #51, opened 2015-10-12
Use PEP 3119-style abstract methods.

vext01:
Look OK?

Note the Python 2 style syntax. Looks like Python 3 introduced kwargs in classdefs.


snim2:
lgtm

Issue #55, opened 2015-10-14
nice -20 benchmarks


vext01:
run benchmarks highest priority. Helps on platforms with out CPU isolation.

vext01:
This is fixed.

I think you need to put "Fixes #55" int he PR to have it auto closed.

PR #64, opened 2015-10-19
Run benchmarks at highest priority. 

snim2:
Resolves Issue #55.


vext01:
Looks good.

Initially I was worried that we might end up nicing taskset, however `platform.bench_cmdline_adjust()` *prepends* arguments, so I think we are OK. Do you agree?

May I suggest a constant for the nice value used?

ltratt:
Given that `nice` values won't change, I'm not sure a constant is that useful.

vext01:
I tend to define constants instead of magic numbers. Up to you Sarah.

snim2:
I agree a constant is not strictly necessary, but it is more consistent with the original code. Will rebase...

vext01:
LGTM

Issue #53, opened 2015-10-14
Run benchmarks as a new user


vext01:
Run benchmarks as a new user, i.e. '_krun'. Make sure the user has a clean home dir and shell environment.

snim2:
New user should have a known (and constant) environment size

vext01:
Sarah have you started work on this?

vext01:
I mat have trodden on your toes...

snim2:
not yet...

vext01:
phew. I was confused by "ready to publish" above. Mind if I assign myself?

In the future I need to check if someone is already assigned.

snim2:
Sure

PR #66, opened 2015-10-22
Run as a different user.

vext01:
In doing so, move process priority change code to platform.py. It occurs to me that `nice` is a unix utility, and should we (god forbid) ever need this to work on windows, we will be forced to think about what to do thanks to `ABCMeta` abstract methods.

Also adds a sanity check for the user change.

A typical benchmark invocation looks like this now:
```
[2015-10-22 11:20:10 DEBUG] cmdline='sudo -u krun nice -20 taskset 0x2 env LD_LIBRARY_PATH=/home/vext01/research/warmup_experiment/krun/krun/../libkruntime work/pypy/pypy/goal/pypy-c /home/vext01/research/warmup_experiment/krun/iterations_runners/iterations_runner.py /home/vext01/research/warmup_experiment/benchmarks/nbody/python/bench.py 2 3'
[2015-10-22 11:20:10 DEBUG] env='{}'
```

Note the now empty environment (prior to`sudo`) and the use of `sudo` to switch user. `LD_LIBRARY_PATH` is needed so the VMs can pick up my implementation of `clock_gettime_monotonic()` in `libkruntime`.

Looks OK?

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -23,6 +23,8 @@ from krun.mail import Mailer
 BENCH_DRYRUN = os.environ.get("BENCH_DRYRUN", False)
 
 HERE = os.path.abspath(os.getcwd())
+DIR = os.path.abspath(os.path.dirname(__file__))
+MISC_SANITY_CHECK_DIR = os.path.join(DIR, "misc_sanity_checks")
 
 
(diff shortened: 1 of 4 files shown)


snim2:
This all looks sensible to me

snim2:
I think this Fixes #53

PR #68, opened 2015-10-22
Fix per VM environment.

vext01:
Fixes Java, as CLASSPATH was getting lost, and Graal as both CLASSPATH
and JAVA_CMD were getting lost.

OK?


snim2:
lgtm

Issue #41, opened 2015-09-28
Easier session resume


vext01:
When a benchmark session crashes, or when the system loses power, we often need to run again just what we need to.

Currently this is quite a manual process. The experimenter has to manually determine which work needs to be run, run them, then manually merge the new data into the existing data.

It would be better if krun could help us do this.

I imagine something like the following:

* user invokes krun
* krun realises a results file already exists
* krun says `partial results detected. Looks like we need to run: 4 executions of benchmark3 and 4 executions of benchmark 4, OK?`
* If user says yes, krun *appends* results to existing results.

This need not be an interactive process, there could be a `-check-results` and a `-resume` option on the CLI.

ltratt:
Is this implicitly the same as reboot mode? If so, should we merge it with that issue?

vext01:
No, it's separate. The session may crash in non-reboot mode, and we need a way to continue without over-writing results.

ltratt:
I should have said: is the code which picks up where we left off implicitly the same between the two modes?

vext01:
There will be sharing, yes.

snim2:
@vext01 this issue is still unassigned. Do you want me to take this one along with #54?

vext01:
Sure. Go ahead.

vext01:
I imagine we need to flesh out the working of this feature too. When I raised this, I did not have the reboot mechanism in mind.

snim2:
OK, well your initial issue here suggests that `krun` starts by looking for a results file, and if one exists, that it checked against the current configuration to see if the results are complete. 

If the results are not complete, and `-resume` has been passed as a CLI then the next execution can be started.

Each execution then needs to append to the JSON file, rather than writing the file out at the end of all the executions.

Also, there is no need for a file in `/var/log` or similar, to say which VM-benchmark pair was last executed, as the same information can be found in the results file.

In this scheme, #54 just needs to have a script in `rc.local` that starts `krun` with `-resume` switched on. So, I'm not sure whether `krun` needs to know whether it is running in reboot-mode or not, the file in `rc.local` just exists or doesn't.

Does that make sense?

vext01:
Yes, and I think this all sounds sensible. Go for it.

snim2:
Edited that slightly.

vext01:
This is separate, but I would also like a way to re-run a subset of experiments.

[...]

PR #69, opened 2015-10-22
Add basic travis config

snim2:
There are some tests in `krun/tests/` which can be run with py.test, and some of the work for Issue #41 (such as appending to a JSON file) are testable. 

This PR adds a travis config to run these tests automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: python
+
+python:
+  - "2.6"
+  - "2.7"
+
+script: py.test


snim2:
We need to turn travis on in the repo, and also the travis account needs to turn the repo on (if that makes sense). I'm not sure whether this github account already has a travis account?

vext01:
Are you happy to look into that?

snim2:
I can do all that, but I think I need the password to this GH account, otherwise everything will end up being owned by snim2.

vext01:
I don't think there is an account for the team. Just admins for the team?

snim2:
Hmm. OK, well I'll have a go, and if we aren't happy with the set up we can always change it later.

vext01:
Good luck!

snim2:
OK, this seems to be fine: https://travis-ci.org/softdevteam/krun

vext01:
Nice! Well done.

Issue #57, opened 2015-10-14
DOCUMENT KRUN!


vext01:
Krun is becoming complicated. We should document it.

snim2:
This needs a couple of small changes to close:
  * Check `examples/` works on BSD and make any necessary changes
  * Add documentation for BSD to `examples/README.md`
  * Move the information in `examples/README.md` to the main README

snim2:
Also:
  * Python must be python2.7 (see PR #74 )

snim2:
Document the `mk_graphs.py` script. Requires:

  * matplotlib
  * statsmodels

snim2:
For Linux, prefer `adduser` to `useradd`

vext01:
Are we happy with the level of documentation at present? Shall we close this?

snim2:
The original plan was to move this out of `krun/examples/README.md` and into `krun/README.md` once we have included some BSD notes. You can move it over now if you wish and put the other work in with Issue #87 ?

vext01:
Sounds sensible.

Issue #58, opened 2015-10-14
Remove `warmup_upon_iter` and `N_GRAPHS_PER_EXEC` support.


vext01:
Dealing with warmup and drawing graphs are not the responsibilities of krun

snim2:
Can this be closed? GH search doesn't come up with any results for those terms in this repo

vext01:
There may be some references elsewhere, but we can fix them as we find em.

snim2:
Thanks. I can't find anything with egrep either, so I think we are safe.

PR #71, opened 2015-10-26
Updated examples directory

snim2:
This is an updated version of the old `examples/` directory, which is a cut down version of the warmup benchmarks. The aim is to use this to test the work to resolve Issue #41 and `reboot mode`.

This isn't quite ready to merge, firstly it needs an `examples/README.md`. Secondly, running this on `bencher5` results in:

```bash
$ cd krun/examples/benchmarks
$ make
for i in richards nbody ; do \
	echo "Building java benchmark ${i}..."; \
	cd /home/snim2/krun/examples/benchmarks/${i}/java && \
	CLASSPATH=../../../krun/iterations_runners/ javac *.java; \
	done
Building java benchmark richards...
KrunEntry.java:2: error: cannot find symbol
class KrunEntry implements BaseKrunEntry {
                           ^
  symbol: class BaseKrunEntry
KrunEntry.java:4: error: cannot find symbol
      (new richards()).runIter(param);
           ^
  symbol:   class richards
  location: class KrunEntry
2 errors
Building java benchmark nbody...
KrunEntry.java:2: error: cannot find symbol
class KrunEntry implements BaseKrunEntry {
                           ^
  symbol: class BaseKrunEntry
1 error
Makefile:8: recipe for target 'all' failed
make: *** [all] Error 1
``` 

which looks like the `CLASSPATH` has been incorrectly set?

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,141 @@
+# krun example
+
+This directory contains a simple experiment using krun.
+This is a good starting point for setting up your own krun configuration.
+
+The example here contains two benchmark programs (*nbody* and *dummy*),
+executed on two VMs (*cPython* and a standard *JVM* such as HotSpot).
+Each benchmark is run for 5 iterations on the same VM, then the VM is
+restarted and the benchmark is re-run for another 5 iterations.
+We say that the experiment runs 2 *executions* and 5 *iterations* of
+each benchmark.
+
+This configuration can be found in the file `examples/example.krun`.
+
+## Step 1: prepare the benchmarking machine
+
+krun currently only runs on Unix-like environments.
+To run this example experiment, you need superuser rights to the machine you are
+using, e.g. on Linux you should be able to run `sudo`.
+
+You need to have the following installed:
+
+  * Python
+  * a Java SDK (version 7)
+  * GNU make, a C compiler and libc (e.g. `sudo apt-get install build-essential`)
+  * cpufrequtils (e.g. `sudo apt-get install cpufrequtils`)
+  * cffi (e.g. `sudo apt-get install python-cffi`)
+
+If you are using a Linux system, you will need to set some kernel arguments.
+If your Linux bootloader is Grub, you can follow these steps:
+
+  * Edit /etc/default/grub (e.g. `sudo gedit /etc/default/grub`)
+  * Add `isolcpus=X` to `GRUB_CMDLINE_LINUX_DEFAULT` (where `X` is an integer > 0)
+  * Add `intel_pstate=disable` to `GRUB_CMDLINE_LINUX_DEFAULT`
+  * Run `sudo update-grub`
+
+Create a new user called `krun`, with minimal permissions:
+
+```bash
+sudo useradd krun
+```
+
+## Step 2: Fetch the krun source
+
+```bash
+$ git clone https://github.com/softdevteam/krun.git
+$ cd krun
+```
+
+## Step 3: set `JAVA_HOME`
+
+```bash
+$ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64/
+```
+## Step 4: Build krun
+
+The krun Makefile honours the standard variables: `CC`, `CPPFLAGS`, `CFLAGS`
+and `LDFLAGS`. For example, if you wish to use `clang` rather than `gcc` you
+can append `CC=/bin/clang` to the options here:
+
+```bash
+$ pwd
+.../krun
+$ make JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/linux"' \
+    JAVA_LDFLAGS=-L${JAVA_HOME}/lib ENABLE_JAVA=1
+```
+
+## Step 5: Build the benchmarks
+
+```bash
+$ cd examples/benchmarks
+$ pwd
+.../krun/examples/benchmarks
+$ make
+```
+
+## Step 6: Run the example experiment
+
+```bash
+$ cd ../
+$ pwd
+.../krun/examples
+$ PYTHONPATH=../ ../krun.py example.krun
+```
+
+You should see a log scroll past, and results will be stored in the file:
+`../krun/examples/example_results.json.bz2`.
+
+## Creating your own experiments
+
+The configuration file `examples/example.krun` controls the experiment here.
+To create your own experiments, you can start by expanding on this example.
+The directory structure for the
+
+```
+experiment/
+    experiment.krun
+    benchmarks/
+        Makefile
+        benchmark_1/
+            language_1/
+            benchmark_file.lang1
+    ...
+```
+
+The `Makefile` (or similar build configuration) should perform any necessary
+compilation or preprocessor steps.
+Each benchmark should expose a function (or method) called `run_iter`
+which is the entry point to the benchmark.
+In the case of compiled languages, it often makes sense to add an extra
+class to an existing benchmark, in order to provide an entry point.
+
+The example here uses Java and cPython.
+The following VMs are currently supported:
+
+  * Standard Java SDK (such as Hotspot)
+  * GraalVM
+  * cPython
+  * Lua
+  * PHP
+  * Ruby
+  * JRuby
+  * Truffle-JRuby
+  * Javascript V8
+
+To add a new VM definition, add a new class to `krun/vm_defs.py` and a
+new iteration runner to the `iterations_runners` directory.
+
+The following platforms are currently supported:
+
+  * Generic Linux
+  * Debian-based Linux
+
+To add a new platform definition, add a new class to `krun/platform.py`.
+
+## Licenses
+
+The *nbody* benchmark comes from the Computer Language Benchmarks Game, which
+are published under a revised BSD license:
+
+  http://shootout.alioth.debian.org/license.php
--- a/examples/benchmarks/dummy/java/Dummy.java
+++ b/examples/benchmarks/dummy/java/Dummy.java
@@ -1,0 +1,22 @@
+public final class Dummy {
+
+    private static final int DELAY = 1000;  // milliseconds.
+
+    /** The benchmark itself. */
+    private static void dummy() {
+        try {
+            Thread.sleep(DELAY);
+        } catch (InterruptedException e) {
+            System.out.println("Benchmark was interrupted.");
+            System.out.println("Measurements may be inaccurate.");
+        }
+    }
+
+    /** Entry point to the benchmark.
+     * This method is called by krun and runs one iteration of the benchmark.
+     */
+    public static void runIter(int n) {
+        Dummy.dummy();
+    }
+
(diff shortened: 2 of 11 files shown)


vext01:
Parts of krun need to be compiled. There is a `Makefile` in the krun root directory. Since you need Java support, you will need to set some make variables on the command line. Loop at the `build-krun` target in the warmup experiment `Makefile` for an example.

vext01:
You will may need to set a `CLASSPATH` too.

snim2:
OK, this is now fixed, documented, and I have tested it on my laptop.

vext01:
This looks pretty good Sarah. Especially the much needed documentation. I'm going to add some comments. They will be pretty nit-picky -- feel free to accept or reject them as you see fit.

examples/README.md
```diff
@@ -0,0 +1,85 @@
+# krun example
+
+This directory contains a simple experiment using krun.
+This is a good starting point for setting up your own krun configuration.
+
+The example here contains two benchmark programs (*nbody* and *dummy*),
+executed on two VMs (*cPython* and a standard *JVM* such as HotSpot).
+Each benchmark is run for 5 iterations on the same VM, then the VM is
+restarted and the benchmark is re-run for another 5 iterations.
+We say that the benchmarks are *executed* twice, with 5 *iterations*.
```
vext01: Maybe use a different number of iterations and executions, so as to help distinguish them?

examples/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+This configuration can be found in the file `examples/example.krun`.
+
+## Step 0: prepare the benchmarking machine
+
+To run this example experiment, you need `sudo` rights to the machine you are
```
vext01: Can we be more generic WRT sudo here. On different platforms we may use different tools. E.g. on solaris `pfexec` and on OpenBSD `doas`.

examples/README.md
```diff
@@ -0,0 +1,85 @@
...
+```bash
+$ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64/
+```
+## Step 3: Build krun
+
+```bash
+$ pwd
+.../krun
+$ make JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/linux"' \
+    JAVA_LDFLAGS=-L${JAVA_HOME}/lib ENABLE_JAVA=1
```
vext01: You can also set the `CC` make variable to choose a C compiler. Worth mentioning?

examples/benchmarks/dummy/java/Dummy.java
```diff
@@ -0,0 +1,22 @@
+public final class Dummy {
+
+    private static final int DELAY = 1000;  // milliseconds.
+
+    /** The benchmark itself. */
```
vext01: `**` intentional?

examples/benchmarks/nbody/java/nbody.java
```diff
@@ -0,0 +1,175 @@
+/* The Computer Language Benchmarks Game
```
vext01: Did you take this directly from the language shootout? It looks different to the one in the warmup experiment. Not sure which one we should use...

examples/benchmarks/nbody/java/nbody.java
```diff
@@ -0,0 +1,175 @@
...
+   modified slightly by Chad Whipkey
+
+*/
+
+public final class nbody {
+    public static void runIter(int n) {
+        //int n = Integer.parseInt(args[0]);
+
+        NBodySystem bodies = new NBodySystem();
+        //System.out.printf("%.9f\n", bodies.energy());
```
vext01: Laurie doesn't like commented code.

examples/example.krun
```diff
@@ -0,0 +1,52 @@
...
+# Maximum number of error emails to send per-run
+#MAX_MAILS = 2
+
+DIR = os.getcwd()
+JKRUNTIME_DIR = os.path.join(DIR, "krun", "libkruntime", "")
+
+HEAP_LIMIT = 2097152  # K == 2Gb
+
+# Variant name -> EntryPoint
+VARIANTS = {
```
vext01: Should we describe entry points (the file/class in which to start) in more detail? We may also want to mention the names of the functions in the entry point that are expected to exist?

examples/example.krun
```diff
@@ -0,0 +1,52 @@
...
+}
+
+# list of "bench:vm:variant"
+SKIP=[
+    #"*:CPython:*",
+    #"*:Java:*",
+]
+
+N_EXECUTIONS = 2  # Number of fresh processes.
+N_GRAPHS_PER_BENCH = 2
```
vext01: See #58 regarding this option.

vext01:
OK, done.

ltratt:
There are definite tabs & spaces issues in some of this stuff. We should use 4 spaces unless there's a compelling reason not to.

examples/README.md
ltratt: I think it's OK in the documentation to just mention `sudo` since that's the dominant way of doing such things. Other platforms which don't use `sudo` will work out what to replace it with.

examples/README.md
snim2: ? 5 iterations, 2 executions

examples/benchmarks/nbody/java/nbody.java
snim2: This was from warmup (hence the mixed tabs and spaces).

vext01:
Those are likely present in the original benchmarks. I leave it to you guys to decide whether to fix them. Diff distance vs. readability.

examples/benchmarks/dummy/java/Dummy.java
snim2: Yes, Javadoc. There should be a class-level bit of Javadoc too.

examples/benchmarks/nbody/java/nbody.java
vext01: Maybe from an earlier version? Compare to https://github.com/softdevteam/warmup_experiment/blob/master/benchmarks/nbody/java/nbody.java and it looks different.

examples/benchmarks/dummy/java/Dummy.java
vext01: Ah well, I wouldn't bother adding it.

examples/README.md
vext01: Yeah. Fine.

examples/example.krun
snim2: sure, are there any other names apart from `run_iter`?

examples/example.krun
snim2: What does: `BENCHMARKS = {'dummy':1000, ...}` do? Is it necessary here?

ltratt:
Readability is generally more important (there are lines next to each other with mixed tabs/spaces, which means it looks pretty bad).

snim2:
These latest changes need a good squashing, but hopefully they address the issues in the comments.

The `README.md` is quite extensive now. Would it be better to put this information in `krun/README` and close Issue #57 ?

ltratt:
I really think the `advance` method needs reformatting. Checking again, I'm not even sure if tabs are the problem -- but the formatting is all off, whatever the reason is.

examples/example.krun
vext01: I've just checked, they are all `run_iter`. Note that Java has a special requirement that the benchmark entry point class subclasses `BaseKrunEntry`

examples/README.md
```diff
@@ -0,0 +1,140 @@
...
+  * GNU make, a C compiler and libc (e.g. `sudo apt-get install build-essential`)
+  * cpufrequtils (e.g. `sudo apt-get install cpufrequtils`)
+  * cffi (e.g. `sudo apt-get install python-cffi`)
+
+You also need to set up some kernel arguments.
+To do this on a Debian-based system (including Ubuntu):
+
+  * Edit /etc/default/grub (e.g. `sudo gedit /etc/default/grub`)
+  * Add `isolcpus=X` to `GRUB_CMDLINE_LINUX_DEFAULT` (where `X` is an integer > 0)
+  * Run `sudo update-grub`
```
vext01: Currently we also require intel pstate to be disabled.

examples/README.md
```diff
@@ -0,0 +1,140 @@
...
+            benchmark_file.lang1
+    ...
+```
+
+The `Makefile` (or similar build configuration) should perform any necessary
+compilation or preprocessor steps.
+Each benchmark should expose a function (or method) called `run_iter` which
+is the entry point to the benchmark.
+In the case of compiled languages, it often makes sense to add an extra
+class to an existing benchmark, in order to provide an entry point.
```
vext01: Wrap lines here?

examples/README.md
vext01: These are all linux specific details of course.

snim2: Well, I've stated "Debian-based systems"; I don't know how anything other than grub works!

vext01:
Looks good Sarah! Anything to add @ltratt?

snim2:
@vext01 you are happy to have this documentation in the example/ directory?

examples/README.md
```diff
@@ -0,0 +1,141 @@
...
+executed on two VMs (*cPython* and a standard *JVM* such as HotSpot).
+Each benchmark is run for 5 iterations on the same VM, then the VM is
+restarted and the benchmark is re-run for another 5 iterations.
+We say that the experiment runs 2 *executions* and 5 *iterations* of
+each benchmark.
+
+This configuration can be found in the file `examples/example.krun`.
+
+## Step 0: prepare the benchmarking machine
+
```
ltratt: Step 0? Step 1?!

Anyway, I think we should say something like "on Linux, you can optionally set up some kernel arguments to obtain more reliable results". On non-Linuxes, either there are no equivalent options, or there are different ones. We don't want to make it seem it's Linux only though.

vext01:
@snim2 You could even move it up to the main `README.md` if you liked. I'm not fussy though. Up to you.

examples/README.md
snim2: OK, but the options aren't optional on Linux, krun bails if the environment isn't set up like this :)

snim2:
@vext01 OK, I think that would be sensible to close #57 but I'll raise a new PR for it

snim2:
Ready to squash?

ltratt:
Perhaps my only comment is that we might want to make it a little clearer that some stuff is Linux specific and some isn't. [We'll soon have this running on OpenBSD.] I don't think we have to tell the non-Linux people what to do, as such, but we should make it clear that they're not excluded.

snim2:
@ltratt OK. When we have BSD support I'll check that these examples work on the BSD machine, and raise a new PR to make any necessary changes and move the documentation across to the main README.

ltratt:
I'm happy with that approach.

snim2:
OK, squashed

PR #74, opened 2015-10-27
Resume and reboot modes

### LINKED ISSUE #41: Easier session resume

### LINKED ISSUE #54: Reboot after each execution.
Reboot the system after every execution. Will require a systemd hook.

snim2:
This PR adds two new command-line switches to krun: `--resume` and `--reboot`.

In resume-mode krun will look for an existing set of results. If one is found, krun first checks that the current platform is (approximately) the same as the platform detailed in the results file. If this test passes, the schedule is built and executions which have already been run are removed from the job queue. Old results are added to the current job scheduler, which means the JSon results file can be dumped (rather than appended to), as before.

Under reboot-mode, every time an execution has finished, krun runs a reboot command which is defined with the platform definition (actually, krun currently just prints the command out, to ease testing, this needs to be fixed before merging).

Json-related code has been refactored into `krun/util.py`. A very basic test suite has been added to `krun/tests/`. The documentation in `examples/README.md` has been updated.

Different platforms have different conventions for starting a program on boot. An example `rc.local` file has been added to `etc/`.

Fixes #41
Fixes #54


snim2:
That Travis build failed because the `bz2` module does not contain a context manager in Python 2.6. Is Python2.6 needed?

ltratt:
I think we can safely force the use of python2.7, on the assumption that it will be installed everywhere that we want to run.

vext01:
The design sounds sane. Now I will inspect the code.

etc/rc.local
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh -e
+#
+# rc.local
+#
+# This script is executed at the end of each multiuser runlevel.
+#
+
+nohup sudo -H -u krun python krun.py --resume --reboot /home/krun/krun/examples/example.krun
```
vext01: Do we really want nohup? That means all krun output is going to disk twice.

examples/README.md
```diff
@@ -86,6 +86,30 @@ $ PYTHONPATH=../ ../krun.py example.krun
 You should see a log scroll past, and results will be stored in the file:
 `../krun/examples/example_results.json.bz2`.
 
+## Running in reboot and resume modes
+
+krun can resume an interrupted benchmark by passing in the `--resume` flag:
```
vext01: Say something about the granularity of this feature? I.e. executions.

etc/rc.local
snim2: I'm not sure, but the script needs to `exit 0` to work with the init framework (I need to test this later today)

examples/README.md
```diff
@@ -86,6 +86,30 @@ $ PYTHONPATH=../ ../krun.py example.krun
...
+
+You may wish to use this facility to reboot after every execution.
+To do this, you can pass in both the `--resume` and `--reboot` flags when
+you start krun:
+
+```bash
+$ PYTHONPATH=../ ../krun.py --resume --reboot example.krun
+```
+
+You will also need to ensure that krun is restarted once the machine has
```
vext01: Wrap this?

examples/README.md
```diff
@@ -86,6 +86,30 @@ $ PYTHONPATH=../ ../krun.py example.krun
...
+rebooted.
+You can do this by hand, or by using the boot configuration provided by
+your OS.
+The `krun/etc` directory contains an `rc.local` file which goes with the
```
vext01: Can we rename it `rc.local.linux` as an example `rc.local` will follow for OpenBSD too.

krun.py
```diff
@@ -6,16 +6,16 @@
 usage: runner.py <config_file.krun>
 """
 
-import os, sys, json, time
+import argparse
+import os, sys, time
```
vext01: I've been loosely trying to keep the krun source PEP8 (it's the first softdev project to do so). PEP8 (stupidly) says one import per line. I know the old line was wrong in the regard, but since we are changing it, we may as well comply.

krun.py
```diff
@@ -140,13 +125,16 @@ class ScheduleEmpty(Exception):
 class ExecutionScheduler(object):
     """Represents our entire benchmarking session"""
 
-    def __init__(self, config_file, log_filename, out_file, mailer, platform):
+    def __init__(self, config_file, log_filename, out_file, mailer, platform,
```
vext01: The number of args to this constructor has bugged me in the past. Perhaps we should wrap these into a class? Thoughts?

snim2: Like a `Configuration` class? That sounds like a good idea.

krun.py
```diff
@@ -172,6 +160,15 @@ def jobs_until_eta_known(self):
...
         self.work_deque.append(job)
 
+    def remove_job_by_key(self, key):
+        job_to_remove = None
+        for job in self.work_deque:
+            if job.key == key:
+                job_to_remove = job
+                break
+        if job_to_remove:
+            self.work_deque.remove(job_to_remove)
```
vext01: Should there be an error case if the job was not found? You can use an `else` on the `for` loop. One of Python's curious features...

krun.py
```diff
@@ -203,13 +200,41 @@ def get_overall_time_estimate_formatter(self):
     def add_eta_info(self, key, exec_time):
         self.eta_estimates[key].append(exec_time)
 
+    def build_schedule(self, config, current):
```
vext01: Complete nitpick, but can we give `current` a more descriptive name? I had to search around to find its purpose.

snim2: I would love one, but I couldn't think of anything sensible. Maybe `previous_schedule`, `completed_jobs`?

krun.py
```diff
@@ -252,13 +277,17 @@ def run(self):
...
-            dump_results(self.config_file, self.out_file, self.results,
-                      self.platform.audit)
+            util.dump_results(self.config_file, self.out_file, self.results,
+                              self.platform.audit)
 
             self.jobs_done += 1
             self.platform.wait_until_cpu_cool()
             self.platform.check_dmesg_for_changes()
 
+            # FIX BEFORE MERGE
```
vext01: You could hook into `BENCH_DRYRUN` to skip reboot for testing purposes. Actually I keep meaning to rename this variable to `KRUN_DRYRUN`.

krun.py
```diff
@@ -380,26 +388,45 @@ def sanity_check_user_change(platform):
...
+    parser = create_arg_parser()
+    args = parser.parse_args()
+
+    if not args.config.endswith(".krun"):
+        usage(parser)
 
-    if not config_file.endswith(".krun"):
-        usage()
+    try:
+        if os.stat(args.config).st_size <= 0:
```
vext01: Do we need this check. I think it would bomb out later when it finds config variables missing.

krun.py
```diff
@@ -413,32 +440,32 @@ def main():
...
 
-    log_filename = attach_log_file(config_file)
+    # If the user has asked for resume-mode, the current platform must
+    # be an identical machine to the current one.
+    current = None
+    if args.resume:
+        if os.path.isfile(out_file):
+            current = util.read_results(out_file)
+            if not util.audits_same_platform(platform.audit, current["audit"]):
+                msg = """You have asked krun to resume an interrupted benchmark.
```
vext01: You can use `+` operator so as to make the indentation look better. Or move it out into a global, where we are already at indent level 0. I prefer the former, as the latter ruins locality.

krun/platform.py
```diff
@@ -194,7 +200,6 @@ def adjust_env_cmd(self, env_dct):
             args.append("%s=%s" % t)
         return args
 
-
```
vext01: Two lines between classes in PEP8.

krun/platform.py
```diff
@@ -467,6 +472,11 @@ def collect_audit(self):
     def change_user_args(self, user="root"):
         return [self.CHANGE_USER_CMD, "-u", user]
 
+    def get_reboot_cmd(self):
+        cmd = self.change_user_args()
+        cmd.append(self.REBOOT)
+        return cmd
+
```
vext01: And again here.

krun/platform.py
```diff
@@ -491,7 +501,6 @@ def _fatal_kernel_arg(self, arg, prefix="", suffix=""):
               "  * Run `sudo update-grub`\n"
               "%s" % (prefix, arg, suffix))
 
-
```
vext01: I *think* two lines between functions in PEP8.

krun/tests/quick.krun
```diff
@@ -0,0 +1,49 @@
...
+    'nbody': 1000,
+}
+
+# list of "bench:vm:variant"
+SKIP=[
+    #"*:CPython:*",
+    #"*:Java:*",
+]
+
+N_EXECUTIONS = 1  # Number of fresh processes.
```
vext01: DId you mean to add a results file?

krun/tests/test_util.py
```diff
@@ -0,0 +1,130 @@
...
+    assert config["SKIP"] == []
+    assert config["MAIL_TO"] == []
+    assert config["ITERATIONS_ALL_VMS"] == 5
+    assert config["HEAP_LIMIT"] == 2097152
+
+def test_output_name():
+    assert output_name(".krun") == "_results.json.bz2"
+    assert output_name("example.krun") == "example_results.json.bz2"
+
+def test_log_name():
```
vext01: Ah, how do you deal with log files in reboot mode? Originally the log file was timestamped, so do we end up with multiple log filenames each time krun is started afresh.

krun/util.py
```diff
@@ -78,3 +83,65 @@ def run_shell_cmd(cmd, failure_fatal=True):
...
+    """Check whether two platform audits are from identical machines.
+    A machine audit is a dictionary with the following keys:
+
+        * cpuinfo (Linux only)
+        * packages (Debian-based systems only)
+        * debian_version (Debian-based systems only)
+        * uname (all platforms)
+        * dmesg (all platforms)
+
+    Platform information may be Unicode.
```
vext01: It feels like this should go into the `platform.py` framework somehow...

krun/tests/test_util.py
snim2: I haven't touched log files at all. I can see that concatenating log files might make sense in reboot mode, but if you resume after a long gap I'm not so sure.

krun.py
vext01: I hesitate to call it that as we already have a dict called `config' floating about. What do you think?

krun.py
vext01: Ah, I thought it was the previous results.

krun.py
snim2: I think if we have the class then we need to use it everywhere and not have the dict, rather than having both.

krun.py
snim2: Hmm. Well, you aren't wrong, it's everything from the old results file, which includes the audit etc. This is another case where it might make more sense to have a class?

krun/tests/test_util.py
vext01: Indeed. So you would either need to write the original log file name out to somewhere that krun can pick up after a reboot, or remove the timestamp from the filename. Neither really appeal to me.

How about this:
* When starting  a new session, krun touches the config file, thus updating the mtime on the file. The current timestamp is used for the config file.
*  When resuming the date on the config file mtime is used to determine the log timestamp, thus allowing us to append to the existing log file.

etc/rc.local
vext01: Usually `nohup` will write stdout to `nohup.out` you see.

krun.py
vext01: That could work, but let's consider that a separate issue?

krun.py
vext01: Sorry, I used the wrong language. That is exactly what I thought it was. The results file contains metadata too. So why not call it `current_result_json`?

krun.py
vext01: OK ignore this. IRC conversation ruled out PEP8.

krun.py
snim2: Grepping through the code, I see that this is the only ENV VAR we get. I think now we have a command line parser it would be sensible (and easier for documentation, etc.) to make this a command-line argument.

krun.py
snim2: I think you're right, but this check means that we can give a more meaningful error message to the user, so I am minded to keep it.

krun/util.py
snim2: One way to do that would be to make this a method of the base class of platforms. However, I think platforms and results files are separate things, and I think we would be violating SRP if we do that. What I would suggest instead is to make a class that represents results files, which can compare two audits, read and write the results files, and so on. That way the execution scheduler would just receive a results file object as an argument to `build_schedule` and extract the previously executed keys via a method on the results file. 

This makes a lot more sense to me, in terms of OOP design and it allows for easier testing, but I think it needs to be a separate PR.

ltratt: What is SRP?

snim2: Single Responsibility Principle

ltratt: OK (although those terms always suggest a rigidity of thought that makes me nervous!). I tend to agree in this case that splitting apart results and platforms sounds sensible.

krun.py
vext01: On Tue, Oct 27, 2015 at 04:14:23AM -0700, Sarah Mount wrote:
> >              self.jobs_done += 1
> >              self.platform.wait_until_cpu_cool()
> >              self.platform.check_dmesg_for_changes()
> >  
> > +            # FIX BEFORE MERGE
> 
> Grepping through the code, I see that this is the only ENV VAR we get. I think now we have a command line parser it would be sensible (and easier for documentation, etc.) to make this a command-line argument.

As I recall, there is also KRUN_DEBUG, which I use a lot.

I don't mind these being CLI args.

-- 
Best Regards
Edd Barrett

http://www.theunixzoo.co.uk

snim2:
The latest changes address most of the comments, but dealing with log files and testing `etc/rc.local.linux` is still to go.

snim2:
Logging done, `etc/rc.local.linux` still to be tested.

snim2:
This now works on my machine, with some caveats:

  * I haven't found a way to re-start krun after boot whilst running as a non-root user. When I start krun on the command line I have to give a password to sudo. However, I think this is an issue with how the environment is set up rather than krun.
  * When I tried running this on bencher5 I got an error about cpufrequtils not being installed, but apt-get believes the package is there.

ltratt:
If you use `sudo` in `rc.local` you should be able to do `sudo -u krun <path to krun>` without requiring a password (most, though perhaps not all, `sudo` installs allow root to call `sudo` without a password).

snim2:
yes, but if I run krun from the command line as a non-root user it asks me for a sudo password. I don't think it is doing an `sudo -u krun` when it asks, because krun works whether or not I have a user called krun.

ltratt:
`/etc/rc.local` is run as root, so `sudo -u krun` in that file almost certainly won't require a password. [It is possible someone's set a really silly `sudo` config that requires a password from root, but I haven't seen such a setup yet.]

vext01:
If `rc.local` is running as root (maybe it is not under linux, no idea) then you can just use `su(1)` and bypass `sudo` altogether.

snim2:
Just to be clear, the file I have committed works fine.

vext01:
> yes, but if I run krun from the command line as a non-root user it asks me for a sudo password.

I've not set up no-password root-up. I've just been typing the password. This is probably why.

Re cpufrequtils, is the binary existing and in the path?

```
vext01@bencher3:~$ which cpufreq-info
/usr/bin/cpufreq-info
```

etc/rc.local.linux
```diff
@@ -0,0 +1,24 @@
...
+# This script is executed at the end of each multiuser runlevel.
+# This means that once krun has rebooted the machine, you will not
+# see a login prompt until all benchmarks have finished.
+#
+# Verify this script by running:
+#    sudo service rc.local start
+#
+
+# Re-direct STDOUT and STDERR to a log file.
+exec 2>/tmp/rc.local.log
```
vext01: I think we are getting close to being able to merge.

Perhaps `/tmp` isn't the best place to put the log file, as many systems clear `/tmp` on boot. How about `/var/log`?

examples/README.md
vext01: Don't forget to wrap these lines.

krun.py
```diff
@@ -172,6 +173,15 @@ def jobs_until_eta_known(self):
     def add_job(self, job):
         self.work_deque.append(job)
 
+    def remove_job_by_key(self, key):
+        for job in self.work_deque:
+            if job.key == key:
+                job_to_remove = job
+                break
+        else:
+            raise JobMissingFromConfig(key)
```
vext01: Minor nit. `JobMissingFromConfig` is a misleading name, as jobs don't exist in the config atall? I would go for the more generic `JobMissingError` or seomthing?

vext01:
Bar those few minor bits, I think this is looking in good shape. As long as you are happy that you have tested enough, I would then be happy to merge.

snim2:
OK, if there are no more comments I will start squashing

krun.py
```diff
@@ -203,13 +213,50 @@ def get_overall_time_estimate_formatter(self):
...
+                    self.results[key] = current_result_json['data'][key]
+
+    def run(self, dry_run=False):
         """Benchmark execution starts here"""
 
-        log_and_mail(self.mailer, info,
-                     "Benchmarking started",
-                     "Benchmarking started.\nLogging to %s" % self.log_path,
-                     bypass_limiter=True)
+        util.log_and_mail(self.mailer, info,
```
vext01: One final thing I remembered last night. We are going to get an email every time krun starts? We should only get an email saying benchmarking started if not in resume mode?

krun.py
```diff
@@ -203,18 +216,67 @@ def get_overall_time_estimate_formatter(self):
...
         """Benchmark execution starts here"""
 
-        log_and_mail(self.mailer, info,
-                     "Benchmarking started",
-                     "Benchmarking started.\nLogging to %s" % self.log_path,
-                     bypass_limiter=True)
+        if len(self) == 0:
+            debug("krun started with an empty queue of jobs")
+
+        if not resume:
```
vext01: Logic here is wrong?

vext01:
Sarah, if you are happy with my last commit, I think we can merge this.

Note however that we should address #83 and #84.

PR #86, opened 2015-11-02
Documenting how to consume Krun results files

### LINKED ISSUE #48: Decouple config filename from results filename and remove need for config file from tools which consume results.
Currently the result filename is derived from the config filename. E.g.:

```
warmup.krun -> warmup_results.json
```

Tools that consume results use a function in krun to discover the json filename.

I think it would be better if:

 * The results file was named similarly to the log files. e.g. `warmup_20151008.bz2`
 * Tools consuming results should not need the config file at all.

This is really just an annoyance and isn't all that high priority.

### LINKED ISSUE #76: mk_graphs.py appears to be broken
I think `mk_graphs.py` has fallen behind the changes in the rest of the repo. Running `mk_graphs.py` on a cut-down version of the examples config I get this:

```bash
$ ./mk_graphs.py examples/quick.krun 
  File "./mk_graphs.py", line 296, in <module>
    data_dct = json.load(fh)
  File "/usr/lib/python2.7/json/__init__.py", line 290, in load
    **kw)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 384, in raw_decode
    raise ValueError("No JSON object could be decoded")
ValueError: No JSON object could be decoded
```

Config file needed to reproduce:

```bash
import os
from krun.vm_defs import (PythonVMDef, JavaVMDef)
from krun import EntryPoint

# Who to mail
MAIL_TO = []

# Maximum number of error emails to send per-run
#MAX_MAILS = 2

DIR = os.getcwd()
JKRUNTIME_DIR = os.path.join(DIR, "krun", "libkruntime", "")

HEAP_LIMIT = 2097152  # K == 2Gb

# Variant name -> EntryPoint
VARIANTS = {
    "default-java": EntryPoint("KrunEntry", subdir="java"),
    "default-python": EntryPoint("bench.py", subdir="python"),
}

ITERATIONS_ALL_VMS = 1  # Small number for testing.

VMS = {
    'Java': {
        'vm_def': JavaVMDef('/usr/bin/java'),
        'variants': ['default-java'],
        'n_iterations': ITERATIONS_ALL_VMS,
    },
    'CPython': {
        'vm_def': PythonVMDef('/usr/bin/python2'),
        'variants': ['default-python'],
        'n_iterations': ITERATIONS_ALL_VMS,
    }
}


BENCHMARKS = {
    'dummy': 1000,
    'nbody': 1000,
}

# list of "bench:vm:variant"
SKIP=[
    #"*:CPython:*",
    #"*:Java:*",
]

N_EXECUTIONS = 1  # Number of fresh processes.
```

Benchmarks as in the `examples/` directory.

Related to Issue #48

### LINKED ISSUE #79: --dump-audit --dump-config
It is difficult to see what is in a results file. These CLI options would print the audit or config to STD out, saving the user the effort of loading the bzip / json.

snim2:
This PR adds documentation and example scripts to show the user how to consume a results file produced by Krun. The `mk_graphs` script has been fixed and moved into `examples/`. A simpler script which just iterates over the results and creates `.csv` files has also been added.

Fixes #48
Fixes #76
Fixes #79

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -120,6 +120,32 @@ $ PYTHONPATH=../ ../krun.py example.krun
 You should see a log scroll past, and results will be stored in the file:
 `../krun/examples/example_results.json.bz2`.
 
+## Using a Krun results file
+
+Krun generates a bzipped Json file containing results of all executions.
+This directory (`examples/`) contains two Python scripts which show
+how to consume results in Krun format:
+
+  * `results2csv.py` which converts the results data to `.csv` files
+  * `chart_results.py` which shows a number of interactive charts
+
+`results2csv.py` only requires the Python standard library.
+`chart_results.py` requires maptplotlib and statsmodels v0.6 or higher.
+
+The structure of the Json results is as follows:
+
+```python
+{
+    'audit': '',  # A unicode object containing platform information
+    'config': '', # A unicode object containing your Krun configuration
+    'data': {     # A dict object containing timing results
+        u'bmark:VM:variant': [  # A list of lists of timing results
+            [ ... ], ...        # One list per execution
+        ]
+  }
+}
+```
+
 ## Testing your configurations
 
 It is often useful to test a configuration file, without actually
--- a/examples/README.md
+++ b/examples/README.md
@@ -122,17 +122,8 @@ You should see a log scroll past, and results will be stored in the file:
 
 ## Using a Krun results file
 
-Krun generates a bzipped Json file containing results of all executions.
-This directory (`examples/`) contains two Python scripts which show
-how to consume results in Krun format:
-
-  * `results2csv.py` which converts the results data to `.csv` files
-  * `chart_results.py` which shows a number of interactive charts
-
-`results2csv.py` only requires the Python standard library.
-`chart_results.py` requires maptplotlib and statsmodels v0.6 or higher.
-
-The structure of the Json results is as follows:
+Krun generates a bzipped JSON file containing results of all executions.
+The structure of the JSON results is as follows:
 
 ```python
 {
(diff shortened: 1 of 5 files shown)


examples/README.md
```diff
@@ -120,6 +120,54 @@ $ PYTHONPATH=../ ../krun.py example.krun
 You should see a log scroll past, and results will be stored in the file:
 `../krun/examples/example_results.json.bz2`.
 
+## Using a Krun results file
+
+Krun generates a bzipped Json file containing results of all executions.
```
vext01: Nitpick: JSON is all caps.

examples/README.md
```diff
@@ -120,6 +120,54 @@ $ PYTHONPATH=../ ../krun.py example.krun
...
+import os
+from krun.vm_defs import (PythonVMDef, JavaVMDef)
+from krun import EntryPoint
+
+# Who to mail
+MAIL_TO = []
+...
+
+$ python krun.py --dump-audit examples/example_results.json.bz2
+{u'cpuinfo': u'processor\t: 0\nvendor_id\t: GenuineIntel\ncpu family\t:
```
vext01: Can we format the output better? For each key, perhaps a heading, and print the value such that the terminal interprets (e.g.) tabs?

snim2: Yes, but I think it would be a mistake to change anything other than whitespace (in case the output stops being loadable by a JSON library).

examples/results2csv.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
```
vext01: `python` doesn't exist on some systems. I tend to use `python2.7` if I know I really want Python 2. It gets more annoying for Python 3 of course, as there are multiple current versions of Python3 (but we don't support them anyway).

Ideally we would package this using distutils/setuptools and the install target would rewrite the shebang according to the python interpreter used... As discussed, not yet.

examples/README.md
vext01: OK, so the output is intended to be machine readable?

snim2: I think it would be a surprise to the user if it wasn't. If you want to make it otherwise, I should add some clarity to the documentation to explain...

snim2:
OK, I think the audits are about as pretty as they are likely to get. I have indented them and separated each key-val pair with newlines, like this:

```python
{
    "cpuinfo":  "processor\t: 0\nvendor_id\t: GenuineIntel\ncpu family\t: 6\nmodel\t\t: 37\nmodel name\t: Intel(R) Core(TM) i5 CPU       M 520  @ 2.40GHz\nstepping\t: 2\nmicrocode\t: 0xe\ncpu MHz\t\t: ... 64\naddress sizes\t: 36 bits physical, 48 bits virtual\npower management:",

    "debian_version":   "jessie/sid",

}
```

vext01:
LGTM

PR #88, opened 2015-11-02
Scaffold OpenBSD Support

vext01:
This is a bare minimum effort to get Krun atleast running on OpenBSD. Support is not complete, but enough to get me developing on my OpenBSD laptop (up until now all development was done on Linux over SSH).

Places where proper implementations are missing have been marked with an XXX, and Krun will also emit a warning at runtime. This should make missing functionality hard to forget.

Also fix a bug in Linux support where the wrong superclass constructor was being called.

OK?

### DIFF
--- a/examples/example.krun
+++ b/examples/example.krun
@@ -1,6 +1,17 @@
 import os
 from krun.vm_defs import (PythonVMDef, JavaVMDef)
 from krun import EntryPoint
+from krun.util import fatal
+from distutils.spawn import find_executable
+
+# For a real experiment you would certainly use absolute paths
+PYTHON27_BIN = find_executable("python2.7")
+if PYTHON27_BIN is None:
+    fatal("Python binary not found in path")
+
+JAVA_BIN = find_executable("java")
+if JAVA_BIN is None:
+    fatal("Java binary not found in path")
 
 # Who to mail
 MAIL_TO = []
@@ -23,12 +34,12 @@ ITERATIONS_ALL_VMS = 5  # Small number for testing.
 
 VMS = {
     'Java': {
-        'vm_def': JavaVMDef('/usr/bin/java'),
+        'vm_def': JavaVMDef(JAVA_BIN),
         'variants': ['default-java'],
         'n_iterations': ITERATIONS_ALL_VMS,
     },
     'CPython': {
-        'vm_def': PythonVMDef('/usr/bin/python2'),
+        'vm_def': PythonVMDef(PYTHON27_BIN),
         'variants': ['default-python'],
         'n_iterations': ITERATIONS_ALL_VMS,
     }
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -4,6 +4,7 @@ import time
 import os
 import difflib
 import random
+import sys
 from collections import OrderedDict
 from krun import ABS_TIME_FORMAT
 from krun.util import fatal, run_shell_cmd, log_and_mail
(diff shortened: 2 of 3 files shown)


examples/example.krun
```diff
@@ -1,6 +1,11 @@
 import os
 from krun.vm_defs import (PythonVMDef, JavaVMDef)
 from krun import EntryPoint
+from distutils.spawn import find_executable
+
+# For a real experiment you would certainly use absolute paths
+PYTHON27_BIN = find_executable("python2.7")
```
ltratt: Having these as top-level variables is obfuscatory, since they're only referenced in one place. Why not say `'vm_def' : JavaVMDef(find_executable("python2.7"))`?

ltratt:
OK, although is it worth merging this if you're going to soon implement the other missing pieces? Should they be added to this PR or to a new PR? [I'm neutral, but I'm not quite sure I know what your plan is.]

vext01:
I was actually planning on working on something else on my travels, but it
quickly became obvious I couldn't on my OpenBSD laptop. As it stands krun
outright refuses to run on OpenBSD.

This PR just gets me to the point where I can develop krun on OpenBSD.

I would like to merge if you don't mind. (After I fix that other comment).

ltratt:
OK, I'll wait for the fix then merge.

vext01:
It turns out `find_executable` could actually return `None`, so i propose the previous commit instead of composing the calls. OK?

ltratt:
I can live with that. Though you could, I think, do `'vm_def' : JavaVMDef(find_executable("python2.7") or fatal("Can't find python2.7")` if you want something shorter. I'm fine either way.

[Out of curiosity, does anyone use the `None` result of `find_executable` to do something other than call `fatal`?]

vext01:
If we hadn't checked for the `None`, it would be passed inside krun, and explode somewhere. The result is used later to invoke the VM, so it is essential it is a valid path to a VM and not None.

Re: Shorter. Don't mind either form. Let's just merge (maybe squash first)?

ltratt:
I guess my point is: does anyone ever make use of the fact that `find_executable` returns `None`? If every user immediately calls `fatal`, perhaps `find_executable` should just call `fatal` itself and save users the hassle?

vext01:
Find_executable is not our code.

I did consider wrapping it, but thought it overkill...

ltratt:
Ah, good point.

ltratt:
OK, do the squashing you want, then I'll merge.

vext01:
Splat

vext01:
I need to merge master, please hold off the merge.

vext01:
OK, go for it.

ltratt:
The merge commit should be squashed away too by rebasing the whole branch against master, right?

vext01:
I clicked the "update branch" button on GH. It does not rebase. Just merges
master.

We've not thought about what to do in these cases. To rebase or not? I
don't mind.

ltratt:
Not true :) We've explicitly been rebasing branches against master, as is good practice in other projects.

ltratt:
[In most cases, this simply disappears the merge commits without any effort. Occasionally you need to handle merge errors. It leads to much cleaner and easier-to-follow history.]

vext01:
OK next time I will merge and rebase manually. Its what I would have done
had i not seen the button.

ltratt:
OK, no prob! Let me know when you've rebased against master and I'll merge this branch.

vext01:
OK

ltratt:
Thanks!

PR #89, opened 2015-11-03
Eliminate duplicate log messages.

vext01:
I notice that some logging messages are printed twice. This is because `setup_logging` adds another log stream handler to the already existing one.

Here is the fix. Tested on both OpenBSD and Linux.

OK?

Example duplicate log messages:
```
$ ../krun.py example.krun --resume
INFO:root:Krun starting...
[2015-11-03 14:55:24 INFO] Krun starting...
ERROR:root:No results file to resume. Expected 'example_results.json.bz2'
[2015-11-03 14:55:24 ERROR] No results file to resume. Expected 'example_results.json.bz2'
```

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -262,7 +262,7 @@ def setup_logging(parser):
     stream = logging.StreamHandler()
     stream.setLevel(level)
     stream.setFormatter(CONSOLE_FORMATTER)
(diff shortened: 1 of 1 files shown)


snim2:
lgtm. Nice to hear that krun is working on BSD too.

PR #100, opened 2015-11-09
Revise ruby runner to call JRuby libkruntime wrapper.

### LINKED ISSUE #95: Ruby runner uses wrong timer if run under JRuby.
This was originally report by @cfbolz here (https://github.com/softdevteam/warmup_experiment/issues/55):

The Ruby runner contains this code:

```
if /linux/ =~ RUBY_PLATFORM
    MONOTONIC_CLOCK = Process::CLOCK_MONOTONIC_RAW
else
    MONOTONIC_CLOCK = Process::CLOCK_MONOTONIC
end
```

however, on JRuby ``RUBY_PLATFORM`` is "java", thus the wrong timer is used. Here is a discussion how to fix this: https://www.ruby-forum.com/topic/150942

vext01:
See also https://github.com/softdevteam/warmup_experiment/pull/56

OK?

Fixes #95

### DIFF
--- a/iterations_runners/iterations_runner.rb
+++ b/iterations_runners/iterations_runner.rb
@@ -1,14 +1,13 @@
-if /linux/ =~ RUBY_PLATFORM
-    MONOTONIC_CLOCK = Process::CLOCK_MONOTONIC_RAW
-else
-    MONOTONIC_CLOCK = Process::CLOCK_MONOTONIC
-end
-
-def clock_gettime_monotonic
-    Process.clock_gettime(MONOTONIC_CLOCK)
-end
-
 class AssertionError < RuntimeError
+end
+
+def clock_gettime_monotonic()
+    # XXX this is for a patched JRuby/truffle only for now! 
+    #
+    # JRuby does not yet provide access to the (raw) monotonic clock, and
+    # adding support is non-trivial (not as simple as adding the C constant).
+    # For now we patch JRuby/Truffle to expose our libkruntime function.
+    Truffle::Primitive.clock_gettime_monotonic()
 end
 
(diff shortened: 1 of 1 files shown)


cfbolz:
Hm, I wonder whether we should still support MRI in the runner (by having a check for ``RUBY_PLATFORM != "JRuby"`` or so).

vext01:
If you know a way to detect JRuby, then by all means :)

eregon:
I am implementing `Process.clock_gettime(Process::CLOCK_MONOTONIC_RAW)` in JRuby+Truffle on Linux, so this workaround should not be needed then, unless you prefer to use your native library.

vext01:
As long as it results in a call to `clock_gettime(CLOCK_MONOTONIC_RAW, ...)` on linux.

eregon:
yes it does as of https://github.com/jruby/jruby/commit/4221e4d147a1818f42d35f5ba8426647f8408ca2,
although it might have some overhead as there is some FFI cost (also happens with the other approach).

cfbolz:
you can detect JRuby like this, no?: ``RUBY_PLATFORM``

vext01:
@eregon We are currently stuck on an older version of JRuby/Truffle as we were having issues building a newer one (although I don't fully remember the details now). We would only be able to use such a fix if we could get a newer JRuby/Truffle working.

eregon:
@vext01 No problem, use what works for you!
(if it was a missing maven snapshot it should be fixed)

vext01:
This should do the trick. OK?

cfbolz:
Ok, great!

PR #102, opened 2015-11-10
Simplify platform audit.

### LINKED ISSUE #101: Krun "same platform" check broken?
Hi,

I don't understand the current implementation of `audits_same_platform`:

```
def audits_same_platform(audit0, audit1):
    """Check whether two platform audits are from identical machines.
    A machine audit is a dictionary with the following keys:

        * cpuinfo (Linux only)
        * packages (Debian-based systems only)
        * debian_version (Debian-based systems only)
        * uname (all platforms)
        * dmesg (all platforms)

    Platform information may be Unicode.
    """
    for key in ["cpuinfo", "dmesg", "uname"]:
        if (not key in audit0) or (not key in audit1):
            return False
    if not audit0["uname"] == audit1["uname"]:
        return False
    return True
```

It just checks the existence of keys, and not the content of the audit. Is that intentional? On OpenBSD this always returns `False` as `cpuinfo` is linux specific.

I think:

 * We should be checking the contents of the audits.
 * This should be hooked into the platform hierarchy. This would allow specific platform checks without too much `if linux() elif OpenBSD...`.

Thanks

snim2:
Makes audit check return a sensible value for BSD platforms.

Fixes #101

### DIFF
--- a/krun/util.py
+++ b/krun/util.py
@@ -149,9 +149,6 @@ def audits_same_platform(audit0, audit1):
 
     Platform information may be Unicode.
     """
-    for key in ["cpuinfo", "dmesg", "uname"]:
-        if (not key in audit0) or (not key in audit1):
-            return False
-    if not audit0["uname"] == audit1["uname"]:
+    if ("uname" not in audit0) or ("uname" not in audit1):
         return False
-    return True
+    return audit0["uname"] == audit1["uname"]


vext01:
Works for me.

I don't know if you care for this, but the static analyser says:
```
../krun/util.py|152 col 9| E713 test for membership should be 'not in'
```

Up to you.

snim2:
OK, hang on...

snim2:
OK, changed and squashed.

PR #104, opened 2015-11-10
Eta estimates to file

vext01:
Ensure the ETA's persist between runs. Also moved the rough timer up into the scheduler. I think it makes more sense there.

Tested in dry run reboot mode on the example (but with krun kludged to actually run code. Seems there is a need to separate out `--dryrun` from the ability to fake reboots?). ETAs look sensible.

Added tests and improved existing ones.

OK?


krun/scheduler.py
```diff
@@ -225,6 +217,19 @@ def build_schedule(self, config, current_result_json):
...
+            self.eta_estimates = current_result_json["eta_estimates"]
+            for key, exec_data in current_result_json["data"].iteritems():
+                got_len = len(current_result_json["eta_estimates"][key])
+                expect_len = len(exec_data)
+                if expect_len != got_len:
+                    msg = "ETA estimates didn't tally with results: "
+                    msg += "key=%s, expect_len=%d, got_len=%d" % \
+                        (key, expect_len, got_len)
+                    util.log_and_mail(self.mailer, error,
+                                      "Fatal Krun Error",
```
snim2: I notice you have used `util.log_and_mail()` to report a fatal error and I have used `fatal()` just below. Should these be consistent?

snim2:
OK, I think I have finished reviewing this now. 
Do you want to add `--dump-etas` as a command-line option?

You say "Seems there is a need to separate out --dryrun from the ability to fake reboots?)" -- I thought `--develop` did that? If not, should it?

snim2:
Actually, one last thing... `krun.tests.util.test_read_results()` needs an extra assert.

krun/scheduler.py
vext01: I think we need to establish a general rule for this kind of stuff. Basically, if a crash happens mid-benchmarking, the user will want to know about it via email, otherwise they might just assume it's taking a while. That said it's tricky, as the mailer is instantiated quite late IIRC, so we may not be able to mail out everything without moving the mailer instantiation...

Any thoughts?

snim2: I don't think it helps to have two ways of doing things. If the user needs to know about a fatal crash via email, that should probably go in the `fatal()` function.

vext01:
We could add a `--dump-etas` if you like. What would we use it for though?

WRT dryrun/develop, we have:
```
        if self.dry_run:                                                        
            info("SIMULATED: reboot (restarting Krun in-place)")
            ...
        else:                                                                   
            subprocess.call(self.platform.get_reboot_cmd())
```

and in vm_defs:
```
        if self.dry_run:                                                        
            info("Dry run. Skipping.")                                          
            return ("", "", 0)              
```

then later:
```
        if not dry_run:                                                         
            try:                                                                
                iterations_results = util.check_and_parse_execution_results(stdout, stderr, rc)
            ...
        else:
            return []
```

So whether experiments actually run, and whether we reboot, are governed by the same flag.

It's a separate issue, and a low-priority one at that. Agree?

snim2:
I agree that dryrun etc. is low priority, but if you want to change the behaviour I think you need to raise a separate issue with a spec, or it'll be long forgotten!

In terms of `--dump-etas`, we have a `--dump` flag to output each part of the results file, so I think for consistency it would be better to have one. Would it be useful though? I think that if a benchmark crashes or is interrupted the user might want to look at the ETAs just to get a rough idea of how long is left to run (i.e. to determine whether it is worth starting again from scratch). wdyt?

vext01:
There's the `--dump-etas`. If this looks OK, I will squash some stuff.

snim2:
OK, `--dump-etas` works for me but you will want to add to `README.md`

vext01:
Ah always forgetting the docs.

vext01:
Done.

PR #108, opened 2015-11-12
Persistent error flag and deal with dmesg changes more gracefully.

### LINKED ISSUE #107: Preserve error indicator across reboots.
There is a flag in krun that inidcates whether an error occurred. This is currently not preserved across runs.

This flag is also used to determine whether we should print dmesg changes at the end. I would argue this is no longer needed. It should be sufficient to mail out and log the dmesg changes as they occur.

vext01:
* Store error flag out to results file and restore in resume mode.
 * Don't buffer up `dmesg` changes -- just print as they occur and warn at end if necessary.
 * Refactor `dump_results` whose argument list was beginning to offend.

OK?

Fixes #107


krun/tests/test_util.py
```diff
@@ -124,30 +124,37 @@ def test_read_results():
 
 
 def test_dump_results():
-    config_file = 'krun/tests/example.krun'
```
snim2: Changes in `test_dump_results` should probably suggest a change in the test that deals with reading results and also the binary results files in `krun/tests/`

snim2:
Have you found a way to create a suitable error to test this? (not necessarily in the unit tests, just from the command line)

vext01:
I tested by:
 * Making a benchmark crash. Check warning printed at time of crash and at end.
 * Kludging the `lines` variable in `platform.check_dmesg_for_changes()`. Check warning printed at time of change and end.

krun/tests/test_util.py
vext01: `krun/tests/quick_results.json.bz2` was already updated. Just amending another test now.

snim2: Thanks, I think I missed some of that further down the diff.

snim2:
OK, I will have a go at testing that. In the mean time, you need to run a `git pull --rebase origin master` on this branch and push the result to bring the branch up to date now that the other PR has been merged.

vext01:
Rebased.

snim2:
OK, so I've clobbered a benchmark and I can see some error reporting from the text on STDOUT, and when I run `--dump-data` the timing information for the clobbered benchmark is empty. It this what you'd expect?

vext01:
Yes, I think so. Please ensure the benchmark still has an ETA estimate in the results file.

What happens if you try to `--resume`? I think we expect it *not* to retry.

snim2:
The etas are as follows (the dummy-cpython benchmark was clobbered):

```json
[2015-11-12 11:41:08: INFO] Krun starting...
{
  "dummy:CPython:default-python": [
    0.1731560230255127, 
    0.12900996208190918
  ], 
  "dummy:Java:default-java": [
    5.096611976623535, 
    5.119085073471069
  ], 
  "nbody:CPython:default-python": [
    0.22838115692138672, 
    0.27140188217163086
  ], 
  "nbody:Java:default-java": [
    166.4416651725769, 
    176.5242908000946
  ]
}
```

but the benchmark ran to completion...

vext01:
Looks good. This indicates that the benchmark took time to crash, and thus will be factored into ETA estimates.

snim2:
OK, do you want to do any squashing before we merge?

vext01:
Good as-is.

snim2:
OK, just to check, did you add another `assert` to `krun.tests.test_util.test_read_results` to check the value of the new entry in the JSON?

krun/tests/test_util.py
```diff
@@ -124,30 +124,37 @@ def test_read_results():
...
             assert dumped_results['config'] == config
-            assert dumped_results['data'] == all_results
-            assert dumped_results['reboots'] == reboots
-            assert dumped_results['eta_estimates'] == dummy_etas
-        os.unlink(out_file)  # Clean-up generated file.
+            assert dumped_results['data'] == dummy_sched.results
+            assert dumped_results['reboots'] == dummy_sched.nreboots
+            assert dumped_results['eta_estimates'] == \
+                dummy_sched.eta_estimates
+            assert dumped_results['error_flag'] == dummy_sched.error_flag
```
vext01: Here.

snim2: Strange, that didn't appear when I pulled this PR. Still, we can trust in GitHub (hopefully)...

PR #110, opened 2015-11-12
Ensure ASLR is off on Linux

### LINKED ISSUE #70: Turn off ASLR if possible
We should turn off address space layout randomisation on platforms that support it.

vext01:
* Check is ASLR is off, if not turn off.
 * More robust checking of CPU governors.

Fixes #70 

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -194,6 +194,7 @@ class BasePlatform(object):
             warn("Not adjusting CPU governor due to developer mode")
             return
         else:
+            debug("Save power")
             self._save_power()
 
     @abstractmethod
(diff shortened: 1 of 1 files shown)


snim2:
This looks OK to me. What does it do in the presence of `--develop`?

krun/platform.py
```diff
@@ -400,6 +402,7 @@ def check_preliminaries(self):
         self._check_cpu_scaler()
         self._check_perf_samplerate()
         self._check_tickless_kernel()
+        self._check_aslr_disabled()
```
vext01: The function `check_preliminaries` is called from `krun.py` and is guarded like this:

```
    if not args.develop:                                                        
        platform.check_preliminaries()                                          
    else:                                                                       
        # Needed to skip the use of certain tools and techniques.               
        # E.g. taskset on Linux, and switching user.                            
        warn("Not checking platform prerequisites due to developer mode")       
        platform.developer_mode = True
```

vext01: So it is skipped if `--develop` is passed.

snim2:
OK, lgtm.

Issue #103, opened 2015-11-10
Improve rudimentary platform check


snim2:
When using resume-model, Krun performs a rudimentary checks to determine whether the benchmark is being resumed on the same machine that it was started on. Currently, this check only checks that the `uname` of the two machines is identical.

This test should differ between platforms, and it would make sense to refactor it into the classes in `platform.py`. This may happen as part of #75 

For Linux, we could use a strong check such as `blkid`, which returns a unique ID of the root file system. I guess someone could have installed more RAM between resumes, so this doesn't catch every case, but anyhow @ltratt felt this was overkill and suggested we use `hostname`, which is available on most (all?) Unix like platforms.

A more detailed fix would check every value in the audit dictionary. However, `platform.audit()` returns a `str`, but the audit in the JSON file is a `unicode`. On my machine I have packages with names that contain acutes and other accents, and a `==` between "this" platform and the previous one always returned `False`. The Internet contains a wide variety of solutions to this problem, and I didn't get any of them to quite work: hence the current hack.

ltratt:
I must admit, I was never unconvinced that this is either necessary or feasible; and your travails have probably reinforced that feeling!

vext01:
You mean you were never *convinced*, right? I think so :)

ltratt:
Oops, yes!

PR #111, opened 2015-11-13
Refactor configuration and results information

### LINKED ISSUE #75: Minor refactorings of commonly used dictionaries
This is from the discussion on PR #74 

Data from krun configuration files (and results files) is loaded when `krun.py` starts, and are used in various parts of the code. Usually parts of the configuration (such as filenames) are passed around as separate variables. It would be neater to factor these out into classes, so that a single object can be passed around (and tested).

snim2:
This PR:

  * Creates three new objects `krun.audit.Audit`, `krun.config.Config`, `krun.results.Results`
  * Simplifies the 'dump' CLI options considerably. 
  * Removes some complex logic from the scheduler, since results from previous executions are loaded directly into `krun.results.Results` objects from disk.
  * Remove `krun.util.audits_same_platform`, since `krun.audit.Audit` objects can now be compared directly with `==`. 
  * `unicode(Audit(...))` is correctly formatted for `--dump-audit`.
  * `ExecutionScheduler` objects now initialise VM definition objects, simplifying the logic in `krun.py` and making it less likely that clients of the library will forget to correctly initialise the objects.
  * `krun.results.Results` objects can serialise themselves on disk, `krun.config.Config` objects can read from config files on disk or from a string.

Fixes #75

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -12,6 +12,7 @@ import locale
 
 import krun.util as util
 from krun.platform import detect_platform
+from krun.results import Results
 from krun.scheduler import ExecutionScheduler
 from krun import ABS_TIME_FORMAT
 from krun.mail import Mailer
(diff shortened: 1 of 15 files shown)


krun.py
```diff
@@ -101,109 +103,90 @@ def sanity_check_user_change(platform):
 def create_arg_parser():
     """Create a parser to process command-line options.
     """
-    parser = argparse.ArgumentParser(description='Benchmark, running many fresh processes.')
```
vext01: All of the changes to this block are quote substitutions? Which rules are you using to decide which to use?

vext01: Ah i see, you store a string of which dump.

krun.py
```diff
@@ -101,109 +103,90 @@ def sanity_check_user_change(platform):
...
-    # In this case print will raise.
-    #
-    # The correct thing to do is to encode() the unicode to the system locale.
-    print(text.encode(locale.getpreferredencoding()))
-
 
 def main(parser):
     args = parser.parse_args()
 
-    if (args.dump_config or
```
vext01: This is much nicer :)

krun.py
snim2: When I wrote the argument parsing code I used single quotes throughout, then I looked a bit further and realised that you had used double-quotes throughout. I thought I'd already changed all these to be consistent with your code, but apparently not.

Were you following some more complicated convention?

krun.py
```diff
@@ -260,8 +242,6 @@ def main(parser):
         platform.developer_mode = True
 
     platform.collect_audit()
-    for vm_name, vm_info in config["VMS"].items():
```
vext01: Note to self, figure out how platform is set in the VM def

krun.py
vext01: I don't currently have any rule in place. I actually had to search to see what the convention is.

There's no difference between `'` and `"` in python, but it seems people use `'` to express the notion of a string which is constant and `"` for format strings, which are *conceptually* expanded.

E.g.:

 * `'hello'`
 * `"hello %s" % name`

We could adopt such a notion if you like(?).

snim2: Well, up to you, but maybe we should change code as we come to it rather than a big bang. Let me know if you want me to change the strings in this PR.

ltratt: Let's please not use arbitrary rules for `""` and `''` -- we'll never remember it, no-one will know why we did it, and so usage will quickly drift. Please just pick one and use it consistently in a file.

krun/config.py
```diff
@@ -0,0 +1,94 @@
...
+        return self.text
+
+    def __eq__(self, other):
+        # Equality should ignore filename.
+        return (isinstance(other, self.__class__) and
+               (self.text == other.text) and
+                (self.MAIL_TO == other.MAIL_TO) and
+                (self.MAX_MAILS == other.MAX_MAILS) and
+                (self.VMS == other.VMS) and
+                # (self.VARIANTS == other.VARIANTS) and
```
vext01: commented code.

krun/results.py
```diff
@@ -0,0 +1,100 @@
...
+
+import bz2  # decent enough compression with Python 2.7 compatibility.
+import json
+
+
+class Results(object):
+    """Results of a Krun benchmarking session.
+    Can be serialised to disk.
+    """
+    def __init__(self, config, platform, results_file=None):
```
vext01: If it were me, I would have added some whitespace in the constructor, but up to you.

snim2: You mean a blank line before line 12?

krun.py
vext01: Not sure what you mean Laurie. Do you mean use only `"` or `'` in any given file?

krun/results.py
vext01: I suppose you could add one there too, but I was referring to the body of the constructor. Just to separate out the concepts.

krun/tests/test_scheduler.py
```diff
@@ -141,22 +130,21 @@ def dummy_shell_cmd(text):
...
-    sched = ExecutionScheduler("krun/tests/example.krun", "example_test.log",
-                               "example_test.json.bz2", mailer,
-                               platform, resume=True,
+    sched = ExecutionScheduler(Config("krun/tests/example.krun"),
+                               mailer,
+                               platform, resume=False,
                                reboot=True, dry_run=True,
                                started_by_init=True)
-    sched.build_schedule(config, None)
+    # for vm_name, vm_info in sched.config.VMS.items():
```
vext01: commented code

krun.py
vext01: Hrm, where does that happen actually?

krun/results.py
```diff
@@ -0,0 +1,106 @@
...
+    def jobs_completed(self, key):
+        """Return number of executions for which we have data for a given
+        benchmark / vm / variant triplet.
+        """
+        return len(self.data[key])
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        return (self.config == other.config and \
```
vext01: Not sure if this interests you: `krun/results.py|100 col 49| E502 the backslash is redundant between brackets`

krun.py
snim2: `krun.scheduler.Scheduler` lines 131--133 in the constructor.

krun/config.py
```diff
@@ -0,0 +1,94 @@
...
+
+        return False
+
+    def __str__(self):
+        return self.text
+
+    def __eq__(self, other):
+        # Equality should ignore filename.
+        return (isinstance(other, self.__class__) and
+               (self.text == other.text) and
```
vext01: Odd indent, found by static checker.

krun.py
vext01: OK

vext01:
OK, so this looks good to me and it seems to work on my laptop.

Do you want to do any squashing?

snim2:
OK, I have squashed.

There are only two small things I am not entirely happy with. One is that comparing two audits ignores the packages. This needs to be fixed as part of #103 

The other thing is that the scheduler now sets the platform of the VM objects. Should anyone use Krun as a library, this is still rather opaque, but I couldn't think of a better place to put that code.

vext01:
I don't think these should block your merge. Overall this PR improves the quality of the code and those nits can be addressed later. Agree?

snim2:
Agree. I'll add the second point as a new issue.

krun.py
ltratt: Yes. [And preferably be consistent in a project; but at least be consistent in a file.]

ltratt:
I really don't think we should bother with package checks and the like. It's too much effort for no real gain. Let's please drop it.

krun.py
vext01: Right. We may find that we can't be this black and white, for example when nesting quotes. But we can try if you really feel strongly.

Not sure I want to impose any rules TBH.

vext01:
OK, let's go.

PR #114, opened 2015-11-16
Sanitise audit objects. 

### LINKED ISSUE #103: Improve rudimentary platform check

snim2:
Store all strings as unicode. 
Remove complex tests for equality.

Fixes #103 
(resolves rather than fixes!)

### DIFF
--- a/krun/audit.py
+++ b/krun/audit.py
@@ -7,6 +7,9 @@ class Audit(object):
     def __init__(self, audit_dict):
         assert isinstance(audit_dict, dict)
         self._audit = audit_dict
+        for key, value in audit_dict.iteritems():
+            if type(value) is str:
+                audit_dict[key] = value.decode("utf-8")
 
     def __contains__(self, key):
         return key in self._audit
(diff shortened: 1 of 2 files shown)


vext01:
LGTM

vext01:
But we need to rebase.

snim2:
Rebased.

snim2:
rebased.

PR #115, opened 2015-11-18
Refactor temperature code.

vext01:
* Store temperatures in a dict instead of in a list with an implicit ordering.
 * Simplify handling of setting starting temperatures and thresholds.
 * Update tests.
  * In doing so make the `MockPlatform` generic so it can be tested on any platform. Previously it was a Linux platform.
 * Add a fixture for making a mock platform and use where appropriate.

This is in preparation for the OpenBSD port.

Thoughts? Suggestions? OK?


snim2:
Looks nice, I like the fixture!

PR #124, opened 2015-12-14
A few fixes

vext01:
Small fixes found from last reboot mode run.

See individual commit messages for more info.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -37,13 +37,6 @@ class BasePlatform(object):
         self.mailer = mailer
         self.audit = OrderedDict()
 
-        # We will be looking for changes in the dmesg output.
-        # In the past we have seen benchmarks trigger performance-related
-        # errors and warnings in the Linux dmesg. If that happens, we
-        # really want to know about it!
-        self.last_dmesg = self._collect_dmesg_lines()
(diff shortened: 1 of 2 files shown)


snim2:
Looks good, but I'm not sure I have understood the reason for not printing out the session time to the log. Presumably we are still collecting the ETAs in the results file, and the ETA that we had in the last run was only two hours out (I suspect because we need to add the 3 minute wait and another minute for rebooting to the ETAs in order to create a more accurate estimate?).

vext01:
You are right that we could compute the overall session time, but as it stands the reported time is wrong:

```
Subject: [krun] Benchmarks Complete                                                                    
                                                                                                       
Message from krun running on bencher3.dcs.kcl.ac.uk:                                                   
                                                                                                       
Completed in (roughly) 533.225929 seconds.                                                             
Log file at: warmup_20151211_170457.log 
```

So that's reporting a session time of 8 minutes, but we know it took >12 hours.

In reboot mode, it is only counting the time for the last execution.

vext01:
If we really want to print the session time, we should raise a low-priority bug and look at it next year.

snim2:
OK

PR #132, opened 2015-12-16
Work around a minor PyPy bug on OpenBSD.

vext01:
See commit message for details.

Also make a util function more robust.

OK?

If, OK, a warmup experiment change will follow (use `PyPyVMDef` for PyPy).


vext01:
OK, try this.

snim2:
OK, this makes sense to me. Do you want to squash the last commit?

vext01:
I think it makes sense to keep separate.

PR #136, opened 2015-12-17
Use sudo on OpenBSD.

vext01:
As discussed in person, use sudo on OpenBSD instead of doas.

Since sudo lives in a different place on different OSs, we have to detect the path dynamically.

Will need rebase after the other PR is in.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -5,6 +5,7 @@ import os
 import difflib
 import random
 import sys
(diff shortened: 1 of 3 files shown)


vext01:
Rebased.

ltratt:
Looks good to me, and definitely the right thing to do.

PR #137, opened 2015-12-18
Tweak clean targets.

vext01:
* Clean in platform_sanity_checks
 * Use `rm -f` instead of `-rm`. Less noise if already cleaned.

OK?


ltratt:
Yes, definitely better this way!

PR #139, opened 2015-12-18
Move all PHONYs after first make target.

vext01:
OK?


ltratt:
My guess is that this worked correctly (because it was `.PHONY`, not `PHONY`) -- do we actually need to merge this?

vext01:
Yep, this is invalid.

PR #140, opened 2015-12-18
Fix a crash on Linux.

vext01:
Trivial mistake.

I don't think we need to restart bencher6 as a result, as this will have no bearing on performance. Agree?

OK?


snim2:
lgtm

PR #144, opened 2016-01-14
Remove CPU pinning support.

vext01:
CPU pinning seemed like a good idea, but in reality it negatively
impacts VMs which use separate threads (or processes)for (e.g.)
compilation.

When the VMs create such threads these children inherit the CPU affinity
(the CPUs on which they can run) from the parent. The result is that we
create artificial CPU contention, thus slowing the VM down.

We found that the initial iterations of some JRubyTruffle benchmarks
were impacted by a >2x slowdown.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,6 @@ If you are using a Linux system, you will need to set some kernel arguments.
 If your Linux bootloader is Grub, you can follow these steps:
 
   * Edit /etc/default/grub (e.g. `sudo gedit /etc/default/grub`)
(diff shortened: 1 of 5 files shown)


snim2:
This all looks fine to me, but you also need to remove the relevant text from the documentation in `README.md`.

vext01:
I believe I did, but will double check. Is there a specific bit I missed?

snim2:
Sorry, you are quite right, I missed the first chunk in the diff.

PR #146, opened 2016-01-14
Pause before taking initial temperature readings.

vext01:
The idea is to allow the system to cool down before the initial temperature readings are taken. Configurable via config file, but defaults to 20 seconds.

OK?

### DIFF
--- a/examples/example.krun
+++ b/examples/example.krun
@@ -59,3 +59,8 @@ SKIP=[
 ]
 
 N_EXECUTIONS = 2  # Number of fresh processes.
+
+# No. of seconds to wait before taking the initial temperature reading.
+# You should set this high enough for the system to cool down a bit.
+# The default (if omitted) is 60 seconds.
+TEMP_READ_PAUSE = 1
--- a/krun.py
+++ b/krun.py
@@ -9,6 +9,7 @@ usage: runner.py <config_file.krun>
 import argparse, json, logging, os, sys
 from logging import debug, info, warn
 import locale
+import time
 
 import krun.util as util
 from krun.config import Config
@@ -245,6 +246,16 @@ def main(parser):
         _, _, rc = util.run_shell_cmd("touch " + args.filename)
         if rc != 0:
             util.fatal("Could not touch config file: " + args.filename)
+
+
+        info(("Wait %s secs to allow system to cool prior to "
+             "collecting initial temperature readings") %
+             config.TEMP_READ_PAUSE)
+
+        if args.develop or args.dry_run:
+            info("SIMULATED: time.sleep(%s)" % config.TEMP_READ_PAUSE)
+        else:
+            time.sleep(config.TEMP_READ_PAUSE)
 
         debug("Taking fresh initial temperature readings")
         platform.starting_temperatures = platform.take_temperature_readings()
--- a/krun/config.py
+++ b/krun/config.py
@@ -26,6 +26,7 @@ class Config(object):
         self.filename = config_file
         self.HEAP_LIMIT = None
         self.STACK_LIMIT = None
+        self.TEMP_READ_PAUSE = 60
 
         if config_file is not None:
             self.read_from_file(config_file)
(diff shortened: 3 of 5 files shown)


examples/example.krun
```diff
@@ -59,3 +59,7 @@ SKIP=[
 ]
 
 N_EXECUTIONS = 2  # Number of fresh processes.
+
+# No. of seconds to wait before taking the initial temperature reading.
+# You should set this hight enough for the system to cool down a bit.
```
ltratt: s/hight/high/

examples/example.krun
```diff
@@ -59,3 +59,7 @@ SKIP=[
...
+TEMP_READ_PAUSE = 1
```
ltratt: Why not set this to a higher number so that people realise it should be high?

ltratt:
Why not make it 30 or 60 seconds? And maybe have Krun do a countdown so that people don't think it's broken?

krun.py
```diff
@@ -246,6 +247,12 @@ def main(parser):
         if rc != 0:
             util.fatal("Could not touch config file: " + args.filename)
 
+        if not args.develop and not args.dry_run:
+            info(("Wait %s secs to allow system to cool prior to "
```
snim2: In the scheduler, `--dry_run` does not wait N seconds after rebooting (or simulating a reboot) but it does print "SIMULATED: time.sleep (would have waited %s seconds)." to STDOUT. I think we should be consistent and print something similar here.

examples/example.krun
snim2: Good idea.

vext01: This particular config file is supposed to be a fast-running example to get user going. See below regarding the default value however.

snim2:
Doing a "countdown" is OK, but we should make sure we switch that off if the Krun has been started with `--started-by-init`.

krun.py
vext01: @ltratt I'm not all that fussed what the default is, we could easily have 60. As for a countdown, I would have thought the log message was enough already.

examples/example.krun
snim2: I think it would be helpful to say what the default is in this comment.

vext01:
Temperature readings (and thus the wait) only happen once at the beginning.

vext01:
So we have now:

```
[2016-01-14 16:00:27 INFO] Krun starting...
[2016-01-14 16:00:27 INFO] Attached log file: example_20160114_160027.log
[2016-01-14 16:00:27 INFO] Checking platform preliminaries
[2016-01-14 16:00:27 INFO] Wait 60 secs to allow system to cool prior to collecting initial temperature readings
[2016-01-14 16:00:27 INFO] SIMLUATED: time.sleep(60)
[2016-01-14 16:00:27 INFO] Running sanity checks
...
```

Normal execution would not show the `SIMULATED` line.

I don't think it's necessary to have a ticking countdown as the user already sees `Wait 60 secs to allow system to cool...`.

OK?

vext01:
Wait, missed a test case.

snim2:
OK, sorry to be a pain, but can we change:

```sh
[2016-01-14 16:00:27 INFO] SIMLUATED: time.sleep(60)
```

To:
```sh
[2016-01-14 16:00:27 INFO] SIMLUATED: time.sleep(60) to give the machine a chance reach a stable temperatue
```

or similar?

vext01:
@snim2 is the prior line not saying something equivalent already?

```
[2016-01-14 16:00:27 INFO] Wait 60 secs to allow system to cool prior to collecting initial temperature readings
```

ltratt:
Can we spell "SIMULATED" properly too? ;)

snim2:
Ah, it wasn't obvious to me that the two lines were related. ISWYM.

vext01:
Typo fixed.

snim2:
OK, it seems we have resolved all the comments. Do you want to squash?

vext01:
Squelch. Go ahead.

PR #147, opened 2016-01-18
Check for CONFIG_NO_HZ_FULL_AUTO tickless kernel without overrides.

vext01:
This properly checks for a tickless kernel (with all CPUs in adaptive-tick mode apart from the boot CPU, since at-least one CPU must not be adaptive-tick).

Our prior tickless check was not correct -- we had enabled tickless functionality, but not enabled it on any CPUs.

Also added a frozen requirements.txt.

OK?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ python:
   - "2.7"
 
 install:
-  - pip install cffi
-  - pip install pytest-cov
-  - pip install pytest-capturelog
+  - pip install -r requirements.txt
 
 script: py.test --cov-report term --cov=krun  krun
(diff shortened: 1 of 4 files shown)


snim2:
Looks fine to me. Presumably the requirements is for travis?

vext01:
Actually, no. Laurie and I realised that we should start freezing versions so our projects have more chance of working in the future. However, you are right that travis should use the requirements file.

Please wait while I fix that...

snim2:
OK, cool. Maybe it would be good to add that clarification to the commit message also?

vext01:
I'll add a note in the commit message when I squash.

OK?

snim2:
OK, want to squash?

vext01:
Here we go.

vext01:
thanks!

PR #148, opened 2016-01-26
Implement --info

vext01:
In the warmup paper draft we want to write something like:

```
Our experiment runs X unique process executions giving a total
of Y in-process iteration readings.
```

Since we will need to recompute X and Y each time we change the SKIP list (we will certainly unskip CPython and maybe some of the explicitly skipped slow benchmark keys at a later date) it makes sense to automate this.

The new parser switch `--info` prints these counts and also lists the keys that were skipped (the latter is just something I personally find useful). After printing this stuff, Krun exits.

E.g.:

```
$ ./krun/krun.py --info warmup.krun 
[2016-01-26 11:14:49 INFO] Krun starting...

Session Info
============

Counts:
  Total process executions:           450
  Total in-process iterations:     900000

Skipped keys:
  fasta:JRubyTruffle:default-ruby
  fannkuch_redux:CPython:default-python
  fasta:CPython:default-python
  richards:HHVM:default-php
  spectralnorm:CPython:default-python
  spectralnorm:JRubyTruffle:default-ruby
  richards:CPython:default-python
  binarytrees:CPython:default-python
  nbody:CPython:default-python
```

Much opportunity for commit squashing and please don't merge until I have updated the documentation.

Otherwise looks good?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ The example here contains two benchmark programs (*nbody* and *dummy*),
 executed on two VMs (*cPython* and a standard *JVM* such as HotSpot).
 Each benchmark is run for 5 iterations on the same VM, then the VM is
 restarted and the benchmark is re-run for another 5 iterations.
-We say that the experiment runs 2 *executions* and 5 *iterations* of
+We say that the experiment runs 2 *process executions* and 5 *in-process iterations* of
 each benchmark.
 
 This configuration can be found in the file `examples/example.krun`.
@@ -123,7 +123,7 @@ You should see a log scroll past, and results will be stored in the file:
 
 ## Using a Krun results file
 
-Krun generates a bzipped JSON file containing results of all executions.
+Krun generates a bzipped JSON file containing results of all process executions.
 The structure of the JSON results is as follows:
 
 ```python
@@ -131,8 +131,8 @@ The structure of the JSON results is as follows:
     'audit': '',  # A dict containing platform information
     'config': '', # A unicode object containing your Krun configuration
     'data': {     # A dict object containing timing results
-        u'bmark:VM:variant': [  # A list of lists of timing results
-            [ ... ], ...        # One list per execution
+        u'bmark:VM:variant': [  # A list of lists of in-process iteration times
+            [ ... ], ...        # One list per process execution
         ]
     },
     'reboots': N, # An int containing the number of reboots that have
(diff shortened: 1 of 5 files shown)


vext01:
Pushed documentation changes.

README.md
```diff
@@ -210,19 +210,23 @@ printed to STDOUT.
 Valid debug levels are: `DEBUG`, `INFO`, `WARN`, `DEBUG`,
 `CRITICAL`, `ERROR`.
 
+The `--info` switch prints the total number of process executions and the total
+number of in-process iterations. It also prints the benchmark keys which will
+be skipped.
+
 ## Running in reboot and resume modes
```
snim2: Having read up to here, it isn't quite clear to me whether `--info` should be used with a config file or a results file (and Krun now takes either as an argument)

krun.py
```diff
@@ -130,6 +130,8 @@ def create_arg_parser():
     parser.add_argument("--develop", action="store_true",
                         dest="develop", required=False,
                         help=("Enable developer mode"))
+    parser.add_argument("--info", action="store_true",
+                        help="Print session info and exit")
```
snim2: Used with a config file?

krun/tests/test_util.py
```diff
@@ -2,12 +2,15 @@
                        log_and_mail, fatal,
                        check_and_parse_execution_results,
                        run_shell_cmd,
-                       ExecutionFailed)
+                       ExecutionFailed, get_session_info)
```
snim2: Is it worth going so far as to rename things like `ExecutionFailed` to be consistent with the new terminology?

snim2:
Nice tests. Some small comments, but this looks good.

krun/tests/test_util.py
vext01: Can do if you like. Probably a separate PR?

snim2: Yes, definitely keep that separate, I think.

vext01:
OK, addressed your comments.

README.md
```diff
@@ -210,19 +210,23 @@ printed to STDOUT.
 Valid debug levels are: `DEBUG`, `INFO`, `WARN`, `DEBUG`,
 `CRITICAL`, `ERROR`.
 
+The `--info` switch reports the total number of process executions and
+in-process iterations that would be run using the specified config file. . It
```
snim2: Is there a stray period here? `file. . It`?

vext01:
fixed.

snim2:
OK, so long as the tests pass then this looks ready to squash...

vext01:
Splat.

PR #156, opened 2016-02-16
Allow the user to pass down extra environment from the config file.

vext01:
Needed for forcing our GCC's libs into the LD_LIBRARY_PATH.

A PR for the warmup experiment config file will follow this PR.

OK?


krun/tests/env.krun
```diff
@@ -0,0 +1,39 @@
+import os
+from krun.vm_defs import PythonVMDef, PyPyVMDef
+from krun import EntryPoint
+
+MAIL_TO = []
+
+HEAP_LIMIT = 2097152  # K == 2Gb
```
ltratt: Is this actually kilobytes or should it just be bytes?

ltratt:
One minor comment; otherwise looks OK (though I was surprised users can't override Krun's environment, and can only append to it).

krun/tests/env.krun
vext01: It's passed to ulimit, which measures in kb.

ltratt: OK, make sense. Can we use "KiB" not "K", please?

vext01:
Until now we have never needed to allow the config file to adjust the environment. Regarding append: krun uses some environment variables internally to make the benchmarks run (e.g. LD_LIBRARY_PATH for PyPy, and CLASSPATH for Java). The user shouldn't be able to overwrite those, otherwise the benchmarks won't run.

ltratt:
OK, then we should probably throw an error if the user specifies those, otherwise they're going to be deeply confused that they're not having an effect?

vext01:
We shouldn't throw an error. It should be possible for the user to append to these variables, but not overwrite them. See `krun/env.py`.

vext01:
To be clear, the interface I just implemented only ever appends the user's environment to whatever Krun needs to do.

krun/vm_defs.py
```diff
@@ -70,6 +70,16 @@ def __init__(self, iterations_runner):
         # environment changes specific to a benchmark.
         self.common_env_changes = []
 
+        # The user can pass in a dict to influence the environment.
+        # These variables are *appended* (they do not override Krun).
```
ltratt: I see the confusion: what this comment means is that if someone specifies (e.g.) `LD_LIBRARY_PATH` then the users values are appended to those set by krun. Let's reword the comment a little bit to make that clearer.

vext01:
I've removed the 'K == 2GB' comments from tests, leaving the revised comment only in the example krun config file. I also revised the confusing comment.

Now let me see why tests failed...

vext01:
OK, try this Laurie.

vext01:
(Please allow time for squashing)

vext01:
I think this is a bit overkill, but here's the change. OK?

ltratt:
Works for me (given that I didn't understand things at first, I really don't think it's overkill!).

Squash and then I'll merge?

vext01:
I suppose it's easy to understand code that you wrote yourself and that you have read hundreds of times ;)

(Splat)

PR #163, opened 2016-02-26
Dont sync disks for sanity checks

vext01:
I noticed that our small run was syncing disks (and waiting 3 mins) multiple times per execution due to sanity checks. This is not necessary. We just need to sync disks before runs we actually hope to take data from.

New tests to confirm this behavior.

OK?


ltratt:
Squash and then I'll merge (in general, if you know you want to squash something, you might as well do it before opening the PR).

vext01:
splat.

PR #166, opened 2016-02-29
Pre-allocate results arrays and print once after all iterations.

### LINKED ISSUE #162: Pre-allocate lists in iterations runners.
The following was migrated from the warmup experiment repo (issue from ltratt):

At the moment, I believe that after every in-process iteration we output a timing to `stdout`. This means the OS does some work and, since IO is often delayed, we might force it to do some work *during* subsequent iterations. However, if we were to have a normal language-level list and add to it after each in-process iteration, we'd be muddling around with the language's memory handling.

My suggestion therefore is that the per-language benchmark handlers should do the equivalent of something like the following:

```python
timings = [None] * NUM_IN_PROCESS_ITERATIONS
for i in range(NUM_IN_PROCESS_ITERATIONS):
    start = time.time()
    run_benchmark()
    duration = time.time() - start
    timings[i] = duration

print(timings)
```

That way we put minimal pressure on the memory subsystem during execution, but also don't run the risk of involving the OS in IO during in-process executions.

vext01:
And some small tidy-ups along the way.

Fixes #162.

OK?

### DIFF
--- a/iterations_runners/iterations_runner.c
+++ b/iterations_runners/iterations_runner.c
@@ -14,6 +14,7 @@
 #include <errno.h>
 #include <dlfcn.h>
 #include <err.h>
+#include <string.h>
 
 #define BENCH_FUNC_NAME "run_iter"
 
@@ -51,6 +52,7 @@ main(int argc, char **argv)
     void     *krun_dl_handle = 0;
     int     (*krun_bench_func)(int); /* func ptr to benchmark entry */
     double    start_time = -1, stop_time = -1;
+    double   *krun_iter_times = NULL;
 
     if (argc != 5) {
         printf("usage: iterations_runner_c "
@@ -76,12 +78,19 @@ main(int argc, char **argv)
         goto clean;
     }
 
-    /* Building a JSON list */
-    fprintf(stdout, "[");
-    fflush(stdout);
+    krun_iter_times = calloc(krun_total_iters, sizeof(double));
+    if (krun_iter_times == NULL) {
+        errx(EXIT_FAILURE, "%s", strerror(errno));
+        goto clean;
+    }
 
-    for (krun_iter_num = 0;
-        krun_iter_num < krun_total_iters; krun_iter_num++) {
+    for (krun_iter_num = 0; krun_iter_num < krun_total_iters;
+        krun_iter_num++) {
+        krun_iter_times[krun_iter_num] = -1.0;
+    }
+
+    for (krun_iter_num = 0; krun_iter_num < krun_total_iters;
+        krun_iter_num++) {
 
         if (krun_debug > 0) {
             fprintf(stderr, "[iterations_runner.c] iteration %d/%d\n",
@@ -93,17 +102,23 @@ main(int argc, char **argv)
         (void) (*krun_bench_func)(krun_param);
         stop_time = clock_gettime_monotonic();
 
+        krun_iter_times[krun_iter_num] = stop_time - start_time;
+    }
+
+    fprintf(stdout, "[");
+    for (krun_iter_num = 0; krun_iter_num < krun_total_iters;
+        krun_iter_num++) {
         fprintf(stdout, "%f", (stop_time - start_time));
+
         if (krun_iter_num < krun_total_iters - 1) {
             fprintf(stdout, ", ");
         }
-
-        fflush(stdout);
     }
-
     fprintf(stdout, "]\n");
 
 clean:
+    free(krun_iter_times);
+
     if (krun_dl_handle != NULL) {
         dlclose(krun_dl_handle);
     }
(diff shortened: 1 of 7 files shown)


iterations_runners/iterations_runner.c
```diff
@@ -76,9 +78,16 @@ main(int argc, char **argv)
         goto clean;
     }
 
-    /* Building a JSON list */
-    fprintf(stdout, "[");
-    fflush(stdout);
+    if ((krun_iter_times =
```
ltratt: This would be easier to read if split into:

```c
krun_iter_items = calloc(...);
if (krun_iter_items == NULL) {
    ...
}
```

vext01:
Rebased.

iterations_runners/iterations_runner.c
```diff
@@ -76,9 +78,16 @@ main(int argc, char **argv)
...
+            calloc(krun_total_iters, sizeof(double))) == NULL) {
+        errx(EXIT_FAILURE, "%s", strerror(errno));
+        goto clean;
+    }
+
+    for (krun_iter_num = 0;
+        krun_iter_num < krun_total_iters; krun_iter_num++) {
```
ltratt: Confusing indentation.

iterations_runners/iterations_runner.c
vext01: It's a standard C idiom, no?

ltratt: That's not the problem: the line is too long to be easily read.

iterations_runners/iterations_runner.c
```diff
@@ -93,17 +102,23 @@ main(int argc, char **argv)
         (void) (*krun_bench_func)(krun_param);
         stop_time = clock_gettime_monotonic();
 
+        krun_iter_times[krun_iter_num] = stop_time - start_time;
+    }
+
+    fprintf(stdout, "[");
+    for (krun_iter_num = 0;
+        krun_iter_num < krun_total_iters; krun_iter_num++) {
```
ltratt: Confusing indentation.

iterations_runners/iterations_runner.lua
```diff
@@ -31,15 +28,14 @@ for BM_i = 1, BM_iters, 1 do -- inclusive upper bound in lua
...
     if BM_i < BM_iters then
         io.stdout:write(", ")
     end
-
-    io.stdout:flush()
 end
-
-io.stdout:write("]")
-io.stdout:flush()
+print("]")
```
ltratt: Should be `io.stdout:write` to match the other lines?

iterations_runners/iterations_runner.rb
```diff
@@ -53,11 +54,15 @@ def assert(cond)
...
+
+    STDOUT.write "["
+    for iter_num in 0..iters - 1 do
+        STDOUT.write String(iter_times[iter_num])
+        if iter_num < iters - 1 then
             STDOUT.write ", "
         end
     end
-    STDOUT.write "]"
+    puts "]"
```
ltratt: Should be `STDOUT.write` to match the others?

ltratt:
I'm not going to pretend that I'm au fait with every language involved in this PR, but it looks good to me, modulo a few minor comments.

cfbolz:
I'd like to take a look at this too

ltratt:
Good idea.

iterations_runners/iterations_runner.lua
vext01: print has a newline suffixed, io.stdout:write does not. Where possible I tried to keep all output on one line. It doesn't really matter since all benchmarking is done by the time we print.

vext01:
I'm not fulent in many of the languages here, so you guys need to check carefully.

iterations_runners/iterations_runner.lua
ltratt: OK, maybe change to something like `io.stdout.write("]\n")` then? It's just if, like me, you flick through, it's not clear what the relationship between `print` and `io.stdout.write` is (do they share the same buffer etc.).

iterations_runners/iterations_runner.c
vext01: It's KNF with tabs expanded to spaces (at your request)?!

ltratt: I don't even understand why it's broken onto 2 lines -- the whole thing is less than 80 chars on 1 line I think? And, if it does need breaking, surely it's better to break the minimum amount onto the second line, not the maximum?

vext01: The lines are 80 long. I've pushed a commit that split it as you suggested.

vext01:
Laurie's comments addressed. @cfbolz any comments?

cfbolz:
looking now

ltratt:
@vext01 Re: line wrapping. The point, I think, is to wrap lines >80 (not >=80) chars. So those lines don't need to be broken at all.

cfbolz:
It looks good to me

ltratt:
OK, if we agree about the 80 char thing, let's fix that, then we can squash and merge.

vext01:
splat.

Issue #160, opened 2016-02-22
Kill daemons


vext01:
Laurie would like to try killing cron and smtpd prior to benchmarking.

ltratt:
This may be the best ever title for an issue!

Yes, Laurie would. And he's also suggesting that there's part of the user config file which allows users to specify which things are killed (e.g. `["smtpd$", "cron$"]`). I suggest that the killing is done *after* krun has done its "wait for 3 minutes after rebooting" thing (which is enough time for e.g. smtpd to send status emails).

vext01:
Sounds good.

ltratt:
This is now implemented.

PR #168, opened 2016-03-01
Implement pre-exec, post-exec and post-session shell commands support.

### LINKED ISSUE #150: Run configurable command every X process executions.
TLDR: It would be useful to have the results and log files offloaded to another host every few executions.

We have been relying heavily on partial results for the warmup draft, and ideally we should only ever copy off results while the machine is *not* benchmarking. If we let krun do the copy itself, we can copy at a time which won't affect benchmarks.

The copy could probably be done by scp -- the user would supply (optionally) in the config file: a host string (e.g. `edd@myhost:/path`), a frequency (e.g. every 10 execs), and an SSH key file (as the login would need to be passwordless).

I'm open to other ways of course.

I recommend implementing #143 first to avoid large network copies.

Thoughts?

CC @snim2 @ltratt @cfbolz

vext01:
Fixes #150 (since we can scp off results with `POST_EXEC_CMDS`and #160 (since we can kill daemons with a `PRE_EXECUTION_CMDS`).

OK?

### DIFF
--- a/examples/example.krun
+++ b/examples/example.krun
@@ -64,3 +64,7 @@ N_EXECUTIONS = 2  # Number of fresh processes.
 # You should set this high enough for the system to cool down a bit.
 # The default (if omitted) is 60 seconds.
 TEMP_READ_PAUSE = 1
+
+# Commands to run before and after each process execution
+#PRE_EXECUTION_CMDS = ["sudo service cron stop"]
+#POST_EXECUTION_CMDS = ["sudo service cron start"]
--- a/krun/util.py
+++ b/krun/util.py
@@ -18,16 +18,22 @@ class ExecutionFailed(Exception):
     pass
 
 
+class FatalKrunError(Exception):
+    pass
+
 def fatal(msg):
     error(msg)
-    sys.exit(1)
+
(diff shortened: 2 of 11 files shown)


examples/example.krun
```diff
@@ -64,3 +64,7 @@ N_EXECUTIONS = 2  # Number of fresh processes.
 # You should set this high enough for the system to cool down a bit.
 # The default (if omitted) is 60 seconds.
 TEMP_READ_PAUSE = 1
+
+# Commands to run before and after each process execution
+PRE_EXECUTION_CMDS = ["date > /tmp/pre_exec"]
```
ltratt: Let's put in examples of pre/post commands people might want to execute (e.g. `/etc/rc.d/cron stop` or whatever).

krun/util.py
```diff
@@ -18,16 +18,22 @@ class ExecutionFailed(Exception):
...
 
 
+class FatalKrunError(Exception):
+    pass
+
 def fatal(msg):
     error(msg)
-    sys.exit(1)
+
+    # We raise, then later this is trapped in attmpt to run the user's
```
ltratt: Typos and grammar.

PR #178, opened 2016-03-03
Move pre/post execution hooks.

vext01:
This moves the pre/post execution hooks up from the Job into the Scheduler so that we can do the post-exec hook after we dump results out.

Also make tests run faster.

OK?


ltratt:
Looks good.

PR #179, opened 2016-03-03
Implement a closed temperature interval when waiting for sensors.

### LINKED ISSUE #176: Improve "System didn't cool down" error message.
```
[2016-03-02 18:35:11: ERROR] System didn't cool down
```

What temperature where we aiming for, and what temperature did we get?

vext01:
This implements a temperature interval 3 degrees either side of the starting temperatures. The sensor must fall within the interval before we continue.

Also wait longer (up to an hour now).

Also improves timeout message: Fix #176.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -1,4 +1,4 @@
-# Various platform related abstrations
+# Various platform related abstractions
 
 import time
 import os
@@ -26,13 +26,8 @@ LIBKRUNTIME_DIR = os.path.join(DIR, "..", "libkruntime")
 class BasePlatform(object):
     __metaclass__ = ABCMeta
 
-    # Each temperature sensor must cool to its starting temperature plus
-    # TEMP_THRESHOLD_DEGREES. An absolute value in degrees centigrade.
-    TEMP_THRESHOLD_DEGREES = 1
-
-    TEMP_MANDATORY_WAIT = 1
-    TEMP_POLL_FREQ = 10                 # seconds between polls
-    TEMP_POLLS_BEFORE_MELTDOWN = 60     # times 10 = ten mins
+    TEMP_THRESHOLD_DEGREES = 3
+    TEMP_WAIT_SECS_BEFORE_GIVEUP = 60 * 60
 
     def __init__(self, mailer):
         self.developer_mode = False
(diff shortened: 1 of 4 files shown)


krun/platform.py
```diff
@@ -26,13 +26,11 @@
...
 
-    # Each temperature sensor must cool to its starting temperature plus
-    # TEMP_THRESHOLD_DEGREES. An absolute value in degrees centigrade.
-    TEMP_THRESHOLD_DEGREES = 1
+    TEMP_THRESHOLD_DEGREES = 3
 
     TEMP_MANDATORY_WAIT = 1
     TEMP_POLL_FREQ = 10                 # seconds between polls
-    TEMP_POLLS_BEFORE_MELTDOWN = 60     # times 10 = ten mins
+    TEMP_POLLS_BEFORE_MELTDOWN = 360    # times 10 = 1 hour
```
ltratt: Can we make this time in seconds, rather than forcing the user to (slightly weirdly) have to divide it by `TEMP_POLL_FREQ`?

krun/platform.py
```diff
@@ -136,31 +131,27 @@ def check_dmesg_for_changes(self):
...
         return False
 
-    def wait_until_cool(self):
+    def wait_for_temperature_sensors(self):
+        """A polling loop waiting for temperature sensors to return (close) to
+        their starting values."""
+
         if self.developer_mode:
-            warn("Not waiting for cooling due to developer mode")
+            warn("Not waiting temperature sensors due to developer mode")
```
ltratt: "Not waiting *for* ..."?

krun/platform.py
vext01: Do we check every second?

ltratt:
A few minor comments. I also wonder if the `MELTDOWN` variable makes sense any more...

krun/platform.py
ltratt: I don't think that matters too much either way. That said, we're probably more likely to hit our threshold if we poll more often, so perhaps 1s is better...

vext01: I ask because, we can't do `time.sleep(ONE_HOUR)`. We need to check every X seconds up to an hour. If we check every second, we are grand as long as the check itself being so frequent won't prevent us from cooling down.

vext01: Shall we try it, and fix later if this becomes an issue?

vext01:
Depends if you mind waiting indefinitely.

krun/platform.py
ltratt: It doesn't matter either way -- you would just sleep for `TEMP_POLLS_BEFORE_MELTDOWN / TEMP_POLL_FREQ` seconds.

ltratt:
What I mean is: "meltdown" implies "too hot", but we now realise we need to care about "too cold" too.

krun/platform.py
ltratt: Yes, let's try polling every second.

vext01: I'm not sure what you mean. Can you show me. Suppose we want to wait an hour.

vext01:
Ah i see. Agreed.

krun/platform.py
vext01: Ah, just got your previous message, ok.

vext01:
Last commit implements our ideas.

ltratt:
OK, let's squash, then I can merge.

vext01:
*crumple*

Issue #126, opened 2015-12-14
Make more parameters configurable (was "Startup sleep duration should be configurable.")


vext01:
```
# Wait this many seconds for the network to come up.                            
STARTUP_WAIT_SECONDS = 3 * 60
```

This should probably be configurable via config file.

vext01:
There actually are many hard-coded parameters that could go into the config file with sensible defaults. Renamed issue to reflect this.

ltratt:
I think that we will do these on an as-needed basis (i.e. if users squeal), but we don't need to do so proactively. Closing.

PR #180, opened 2016-03-04
Shorter wait for system to come up.

vext01:
Fixes # 170.

Went for a two minute wait in the end. What do you think?

In the interest of getting some results quicker, I've not made the wait configurable yet. I revised #126 to include such a task.

OK?


ltratt:
Seems sensible.

Issue #181, opened 2016-03-04
Number of emails sent so far does not persist between reboots.


vext01:
We have a email limiter system, which attempts to not flood us with emails. It holds the number of emails sent so far, and stops sending further mails once this number exceeds a configurable limit.

Sadly the number of emails sent so far does not persist between reboots, so the limit is in-fact per execution for reboot mode.

The number needs to be written into the `json.bz2` file so that it can be reinstated after a reboot.

PR #182, opened 2016-03-04
Try to send an email when a FatalKrunError is recieved.

### LINKED ISSUE #177: Always try to send an email when a FatalKrunError is raised.
This would let us know that krun has crashed almost immediately. Otherwise we need to wait a suitable amount of time in which we saw no log copy, before we decide it (possibly) crashed.

vext01:
This adds code which tries to send an email when something has gone wrong. This should make it easier to know when something has gone wrong (instead of waiting until the results file seems to have taken too long to appear on bencher2).

Additional tweaks:

 * Fix broken exception handler which referenced the post-session hook we since removed.
 * Exit with non-zero return code if we fail.
 * Include hostname in subject line of emails (useful when we have several hosts running)
 * sendmail(8) on OpenBSD returns exit code 75 if smtpd is not running. The mail is still submitted, and when the post-execution hooks bring smptd back up, the mail is delivered OK. At a loss for a better plan of action, I have downgraded the error message to a warning message. This situation *can* happen; if a benchmark crashes, we will get mailed stdout and stderr, (all *after* the execution, so the I/O is OK).

There is still a section of code in `main()` which can raise a `FatalKrunException` outside of our handler. We would not recieve an email if these exceptions fire. Do you think it's worth trying to refactor `main` to cover these code paths too?

In reading our mailer code I see a bug. See #181.

Fixes #177.

OK? CC @snim2

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -7,7 +7,7 @@ usage: runner.py <config_file.krun>
 """
 
 import argparse, json, logging, os, sys
-from logging import debug, info, warn
+from logging import debug, info, warn, error
 import locale
 import time
 
@@ -288,7 +288,9 @@ def main(parser):
     try:
         sched.run()
     except util.FatalKrunError as e:
-        util.run_shell_cmd_list(config.POST_SESSION_CMDS)
+        subject = "Fatal Krun Exception"
+        mailer.send(subject, e.args[0], bypass_limiter=True)
+        util.run_shell_cmd_list(config.POST_EXECUTION_CMDS)
         raise e
 
 
@@ -330,5 +332,5 @@ if __name__ == "__main__":
 
     try:
         main(parser)
-    except util.FatalKrunError:
-        pass
+    except util.FatalKrunError as e:
+        sys.exit(1)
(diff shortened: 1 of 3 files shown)


ltratt:
`sendmail` depends on which mailer we're using. Are you referring to OpenSMTPd or Postfix?

vext01:
Default opensmtpd configuration.

ltratt:
I am not sure this holds true for postfix. Let me check...

vext01:
I don't think it matters, if sendmail fails, we always want to continue, and warning is fine at this point too.

ltratt:
Ah, yes, of course, we have OpenSMTPd installed on bencher6. Point taken.

Ok, I'll merge.

vext01:
Any comment on the "should we refactor"?

ltratt:
Erp, it's impossible for me to tell that from the PR (which I have now merged anyway...). Upto you. Sorry for jumping the gun!

vext01:
I'll have a quick look and if it isn't too much hastle, will raise another PR.

ltratt:
OK. Mea culpa!

vext01:
No worries.

PR #184, opened 2016-03-04
Catch more FatalKrunErrors in the mailer handler.

### LINKED ISSUE #183: Try to include bulk of main() in exception handler
As mentioned in #182, there is a section of code in `main` which raises `FatalKrunError` but is outside our emailer exception handler. I wonder if we can re-factor to catch those too.

vext01:
We can't catch all of them (if we can't parse the config file, we can't get an email address to mail, or post-execution commands to run), but this certainly helps. Good enough (tm).

Fixes #183 

OK?

CC @snim2

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -187,6 +187,22 @@ def main(parser):
 
     attach_log_file(config, args.resume)
 
+    mail_recipients = config.MAIL_TO
+    if type(mail_recipients) is not list:
+        util.fatal("MAIL_TO config should be a list")
+
+    mailer = Mailer(mail_recipients, max_mails=config.MAX_MAILS)
+
+    try:
+        inner_main(mailer, config, args)
+    except util.FatalKrunError as e:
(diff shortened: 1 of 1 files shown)


krun.py
```diff
@@ -187,6 +187,22 @@ def main(parser):
 
     attach_log_file(config, args.resume)
 
+    mail_recipients = config.MAIL_TO
+    if type(mail_recipients) is not list:
```
snim2: Does it have to be a list or can it be any `collections.Sequence` type except `str`?

vext01: Well I guess it could be another collection, but since I have not tested it, and since the config file is ours to define, I think enforcing a list is OK.

vext01: Note this is old code which was moved.

ltratt:
I'm fine with this. Shall I merge?

vext01:
In the nick of time! Yes please!

PR #185, opened 2016-03-07
Try to spin up CPU to make heat if we are too cold.

vext01:
As discussed on email, if we are too cold, we can try running a busy loop to make some heat.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -29,6 +29,10 @@ class BasePlatform(object):
     TEMP_THRESHOLD_DEGREES = 3
     TEMP_WAIT_SECS_BEFORE_GIVEUP = 60 * 60
 
+    TEMP_OK = 0
+    TEMP_TOO_HOT = 1
+    TEMP_TOO_COLD = 2
+
     def __init__(self, mailer):
         self.developer_mode = False
         self.mailer = mailer
@@ -78,11 +82,13 @@ class BasePlatform(object):
 
         Returns tuple: (bool_ok, str_reason_if_false)
 
-        'bool_ok' is True if all temperature sensors are within range.
+        'flag' is BasePlatform.TEMP_OK if all temperature sensors are within
+        range, or BasePlatform.TEMP_TOO_HOT or BasePlatform.TEMP_TOO_COLD
+        otherwise.
 
-        'str_reason_if_false' is None if 'bool_ok' is True, otherwise it is a
-        string indicating the reason the system is not within the desired
-        temperature range.
+        'str_reason' is None if 'flag' is BasePlatform.TEMP_OK, otherwise it is
+        a string message indicating the reason the system is not within the
+        desired temperature range.
         """
 
         readings = self.take_temperature_readings()
@@ -98,8 +104,15 @@ class BasePlatform(object):
                 # This reading is too hot/cold
                 reason = ("Temperature reading '%s' not within interval: "
                           "(%d <= %d <= %d)" % (name, low, now_val, high))
-                return (False, reason)  # one or more sensor too hot
-        return (True, None)
+
+                if now_val < low:
+                    flag = BasePlatform.TEMP_TOO_COLD
+                else:
+                    assert now_val > high
+                    flag = BasePlatform.TEMP_TOO_HOT
+
+                return (flag, reason)  # one or more sensor too hot or cold
+        return (self.TEMP_OK, None)
 
     def _collect_dmesg_lines(self):
         return run_shell_cmd("dmesg")[0].split("\n")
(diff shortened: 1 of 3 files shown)


krun/platform.py
```diff
@@ -98,8 +104,16 @@ def temp_sensors_within_interval(self):
...
                           "(%d <= %d <= %d)" % (name, low, now_val, high))
-                return (False, reason)  # one or more sensor too hot
-        return (True, None)
+
+                if now_val < low:
+                    flag = BasePlatform.TEMP_TOO_COLD
+                elif now_val > high:
+                    flag = BasePlatform.TEMP_TOO_HOT
+                else:
+                    assert False  # unreachable
```
ltratt: This would probably be clearer as:
```python
if now_val < low:
  ...
else:
  assert now_val > high
  ...
```

krun/util.py
```diff
@@ -180,3 +181,15 @@ def print_session_info(config):
...
+
+def make_heat():
+    """A loop which spins, attempting to make heat.
+
+    Used when the CPU is too cool to run a benchmark."""
+
+    until = time.clock() + 1  # in roughly one second
+    while True:
+        # Waste cycles
+        if time.clock() > until:
```
ltratt: I'm not sure that making a syscall so frequently is a great idea (time.clock(), for example, will interfere with `ntp`). Maybe stick a dummy for loop in there as well? Even if we just make the syscall 100 times a second, that'll probably be better.

vext01: I can't guarantee how many times a second the syscall will be called, it depends upon how fast the system can execute python code.

So I think you are suggesting that we should add a loop like `for i in xrange(1000000): pass` or some such? Am I right?

ltratt: Yep, I agree that any no-op will execute for a variable time. Fortunately, in this case, we really don't care about that too much.

Does krun run under cpython or pypy? If the latter, I think that `for` loop will be entirely optimised away. [You'd need to do something like have a variable `j = 0` outside it, to which you add 1 on each iteration of the loop.]

vext01: I run it on CPython. And yeah, it did occur to me that a JIT will eat these loops for breakfast, lunch and dinner (maybe even with a distinct loop variable).

These loops certainly cause a ton of CPU on CPython. I'm half tempted to make krun quit if it's run on PyPy. What do you think?

ltratt: Well, in this case, we can make it still do *something* sensible on PyPy. I'm not super keen on forcing one implementation over the other unless there's a compelling reason.

vext01: How about shelling out to a dash shell loop to make heat?

ltratt: We definitely mustn't do that! It'll cause the OS to do waaaaaaaay too much work. All we need is something like:
```python
j = 0
for i in range(10000000):
   j += 1
assert j == 10000000
```
and we should be OK on PyPy too (at least for the time being... in theory, if the trace optimiser becomes clever enough, it could even turn that into a constant, but let's not worry about that).

vext01: Or I could write a little C program that we call.

ltratt: That's still a shell call. The `for` loop I outlined is OK for our purposes. Let's just do that and not worry too much about it.

vext01: Eh, sorry github PRs dont seem to update properly without refreshing the page. I will check to see if the above loop causes the CPU to spin up under pypy.

vext01: re: I/O doesn't matter since we sync, but I prefer your loop anyway if it works.

vext01:
Ah, we need to think again. If our `make_heat` function doesn't take exactly one second to run, we will end up waiting the wrong amount of time in the temperature timeout.

Any ideas?

ltratt:
If that's the case, we have other problems. Our "wait" functionality should be checking that it is waiting N seconds, not that it is calling M things which each wait N/M seconds (if that makes sense). So I suggest we do the "proper" fix that solves this.

vext01:
I fear that's the only thing we can do yeah.

vext01:
I think we want something like this then?

vext01:
Here we go.

krun/util.py
```diff
@@ -180,3 +180,16 @@ def print_session_info(config):
...
+    """A loop which spins, attempting to make heat.
+
+    Used when the CPU is too cool to run a benchmark."""
+
+    # Waste cycles with this loop. At the time of writing PyPy is unable to
+    # optimise this away. Heat is a consequence.
+    j = 0
+    for i in xrange(10000000):
+        j += 1
+        assert j == 1000000
```
ltratt: This assert is in the wrong level of indentation. I guess that mean the loop isn't currently tested?

vext01: Hrm, investigating.

vext01: I see, it was mocked away earlier. Since this `make_heat()`  runs pretty fast on a modern machine, I'm happy to run it for real. Latest commit fixes and un-mocks it. OK?

ltratt:
Works for me. Squash and then I'll merge?

vext01:
splat.

PR #186, opened 2016-03-15
Check that pre/post-execution commands can contain shell syntax.

vext01:
I used this quick test to check whether our `|| true` hack would work (that pre/post commands use a shell rather than just a raw exec(3)).

Seems it will work just fine and the test might as well go into the suite.

OK?


ltratt:
I'm not totally sure that this test quite models what we want, but there again, I can't think of anything that would (and this is test is better than nothing). So merge it is!

vext01:
All I was looking for is that shell constructs are interpreted. That's all we need :)

PR #189, opened 2016-03-22
Implement a simple result stripper.

vext01:
Last week we had the need to re-run the C benchmarks. I ran them as a separate experiment, but realised that the starting temperatures will differ, so this is wrong.

Using this change we could have run `krun.py --strip-results '*:C:*'` and resumed the experiment to re-run only the C experiments with the same starting temperature.

Let's not merge this hastily, as it has the potential to waste benchmarking time (e.g. run a long experiment and halfway through the reboot count is wrong).

Tested with unit tests, and with the example krun scenario. Looks good, but quite a fiddly change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,31 @@ been prepared for reliable benchmarking.
 Note that you should not collect results intended for publication with
 `--develop`.
 
+## Re-running Part of your Experiment
+
+Sometimes it is necessary to re-run a subset of your experiment. For example,
+if after completing an experiment, you find that one of your VMs was
+miscompiled, you may want to re-run all benchmarks for the troublesome VM only.
+Of course, under ideal circumstances, you would collect all results in one go.
+
+You can use the `--strip-results` mode to strip results from your result file.
+Once stripped, you can run with `--resume` to re-collect the results you
+stripped.
+
+The switch accepts a "key-spec". All process executions matching the key-spec
+are removed. A key-spec is of the form 'benchmark:vm:variant". Any of the three
+fields can also be a star '*' to indicate a wildcard. E.g. `--strip-results
+'*:CPython:*'` will remove all results for the 'CPython' VM. If ou use a
+wildcard, be sure to quote the argument so the shell does not expand the stars.
+
+Note also that the stars are not true wildcards. E.g. the key-spec '*:CPy*:*'
+will not match 'mybenchmark:CPython:myvariant'. Fields are either a value or
+literally '*'.
+
+Once you have stripped results, if you ran an experiment in reboot mode you
+will need to reboot manually to re-run the stripped results (with rc.local set
+up correctly).
+
 ## Licenses
 
 The *nbody* benchmark comes from the Computer Language Benchmarks Game, which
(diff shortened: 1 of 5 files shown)


krun/results.py
```diff
@@ -107,3 +108,47 @@ def __eq__(self, other):
...
+                removed_keys += 1
+                removed_execs += len(new_data[key])
+                new_data[key] = []
+                self.eta_estimates[key] = []
+                info("Removed results for: %s" % key)
+            else:
+                completed_execs += len(new_data[key])
+
+        self.data = new_data
+        info("reb: %s" % self.reboots)
```
vext01: Debug info needs to be removed.

snim2:
Just to be clear, you want to be able to restart a stopped benchmark, and run a specified subset `X` of the config file, having removed any results from `X` from the results file? 

I'm not sure I understand the issue with the starting temperatures. When you resume running the benchmarks in `X`, if the current temperature is higher than the starting temperature in the results file, presumably krun will wait to cool. Is it possible for the current temperature to be lower than the value in the results file?

README.md
```diff
@@ -341,6 +341,22 @@ been prepared for reliable benchmarking.
 Note that you should not collect results intended for publication with
 `--develop`.
 
+## Re-running Part of your Experiment
+
+You can use the `--strip-results` mode to strip results from your result file.
+Once stripped, you can run with `--resume` to re-collect the results you
+stripped.
```
snim2: Maybe add a note to say why / when it is reasonable to do something like this. Generally, running all the benchmarks in one run would be preferable...

README.md
```diff
@@ -341,6 +341,22 @@ been prepared for reliable benchmarking.
...
+
+The switch accepts a "key-spec". All process executions matching the key-spec
+are removed. A key spec is of the form 'benchmark:vm:variant". Any of the three
+fields can also be a star '*' to indicate a wildcard. E.g. `--strip-results
+'*:CPython:*'` will remove all results for the 'CPython' VM. If ou use a
+wildcard, be sure to quote the argument so the shell does not expand the stars.
+
+Note also that the stars are not true wildcards. E.g. the keyspec '*:CPy*:*'
```
snim2: `keyspec` on this line, `key-spec` above.

krun.py
```diff
@@ -132,7 +132,11 @@ def create_arg_parser():
                         help=("Enable developer mode"))
     parser.add_argument("--info", action="store_true",
                         help=("Print session info for specified "
-                              "config file and exit")),
+                              "config file and exit"))
+    parser.add_argument("--strip-results", action="store",
+                        metavar="keyspec",
```
snim2: `metavar="KEYSPEC",` is the usual convention (used below in `metavar="FILENAME",`).

vext01:
The scenario I was faced with last week was:

The benchmarks had completed but a bug in the C iterations runner rendered a subset of the results unusable. I could run a new session, with all but C skipped, but this would mean we would collect a new starting temperature.

Hope this helps.

vext01:
Try this.

snim2:
That looks nice. I'll try out a few test cases before I comment any further.

vext01:
Brought up to date with master.

PR #190, opened 2016-04-12
Fix process priority

vext01:
We were running priority 19 by accident.

This fixes this. Part of the problem was that only the root user can adjust the process priority. To work around this, I use sudo to become root, then adjust the process priority, then drop to the krun user. The krun user inherits the new priority.

This snippet from the krun debug log explains the new model best:

```
[2016-04-12 14:18:20 DEBUG] Wrapper script:
#!/bin/dash
ulimit -d 2097152 || exit $?
ulimit -s 8192 || exit $?
env LD_LIBRARY_PATH=/home/vext01/research/warmup_experiment/krun/krun/../libkrun /usr/bin/python2.7 /home/vext01/research/warmup_experimen
t/krun/iterations_runners/iterations_runner.py /home/vext01/research/warmup_experiment/krun/examples/benchmarks/dummy/python/bench.py 5 10
00 1
exit $?
[2016-04-12 14:18:20 DEBUG] Execute wrapper: /usr/bin/sudo -u root /usr/bin/nice -n -20 /usr/bin/sudo -u krun /bin/dash /tmp/krun_wrapper.
dash
```

Notice that `nice` is no longer in the wrapper script itself, but in the invocation of the wrapper script.

Here's a screenshot of top showing the benchmark now running under priority -20:

prio

Also implemented a Python sanity check which checks that benchmarks are running priority -20. This involved calling out to C to ask for the current process priority, which in turn required a little shared object. I put this in the same dir as` libkruntime.so`, and renamed the directory to reflect that fact that the dir was no longer just for libkruntime.

In hindsight I would have written the sanity check in pure C, which would have removed the need for the shared object altogether. Is it worth doing that now, instead of using Python?


snim2:
I'll read through the code in detail in a moment, but if you would prefer a C-only solution I think it would be better to do this in the current PR, rather than raising an issue.

Makefile
```diff
@@ -6,15 +6,15 @@ PASS_DOWN_ARGS =	ENABLE_JAVA=${ENABLE_JAVA} JAVAC=${JAVAC} \
 			CC=${CC} CFLAGS=${CFLAGS} CPPFLAGS=${CPPFLAGS} \
 			LDFLAGS=${LDFLAGS}
 
-.PHONY: libkruntime vm-sanity-checks clean
+.PHONY: libkrun vm-sanity-checks clean
```
snim2: I think `all` should also be in `PHONY`

vext01:
OK, will make the sanity check C only.

snim2:
Other than that minor comment, code looks good.

vext01:
I decided to keep the directory name change. It's not inconceivable that we might put other shares objects in there in the future.

Please don't merge just yet. Want to test on a clean checkout.

Comments welcome.

snim2:
The code looks good to me. Do you want me to do any testing before we squash and merge or are you happy that the unit tests check the new code?

vext01:
Just tested on a fresh clone. One small thing fixed.

I think the sanity check suffices as a runtime test in this instance.

Can I squash?

snim2:
What I understand from that comment is that we have tested that the sanity-check can correctly give a "true" response, have we also checked that it can fail?

vext01:
Informally, yes.

If I change the expected priority to, e.g. -2:

```
#define EXPECT_PRIORITY -2
```

```
[2016-04-12 15:40:56 DEBUG] running 'Process priority sanity check' sanity check
[2016-04-12 15:40:56 DEBUG] Writing out wrapper script to /tmp/krun_wrapper.dash
[2016-04-12 15:40:56 DEBUG] Wrapper script:
#!/usr/local/bin/dash
ulimit -d 1048576 || exit $?
ulimit -s 8192 || exit $?
env LD_LIBRARY_PATH=/home/krun/krun/krun/../libkrun MALLOC_OPTIONS=sfghjpru /home/krun/krun/iterations_runners/iterations_runner_c /home/krun/krun/krun/../platform_sanity_checks/check_nice_priority.so 1 666 1
exit $?
[2016-04-12 15:40:56 DEBUG] Execute wrapper: /usr/local/bin/sudo -u root /usr/bin/nice -n -20 /usr/local/bin/sudo -u krun /usr/local/bin/dash /tmp/krun_wrapper.dash
[2016-04-12 15:40:56 INFO] stderr: [iterations_runner.c] iteration 1/1
[2016-04-12 15:40:56 INFO] stderr: process priority: expect -2 got -20
[2016-04-12 15:40:56 ERROR] Process priority sanity check sanity check failed: Benchmark returned non-zero or didn't emit JSON list. Exception string: No JSON object could be decoded
return code: 1
stdout:

stderr:
[iterations_runner.c] iteration 1/1
process priority: expect -2 got -20

```

Pushed a fix for duplication in log message.

snim2:
Cool! So, do you want to automate that check or just squash and merge?

vext01:
Do you mean add a unit test? I was hoping to move on as I need new results for next week!

snim2:
OK, have you squashed?

vext01:
One moment

vext01:
squashed.

Will you be on mattermost today Sarah?

snim2:
Yup

PR #192, opened 2016-04-13
Use CPU pinning on a set of CPUs

vext01:
Talks with Intel suggest that we should pin our benchmarks to a *set* of isolated cores. For us it makes a great deal of sense to isolate and pin to our adaptive ticks CPUs (all but the boot CPU due to `NO_HZ_FULL_ALL`).

This set of commits:

 * Ensures the adaptive ticks CPUs are isolated and guides the user if not.
 * Uses taskset to pin benchmarks with the correct affinity mask.
 * Tests all of this.
 * Sanity checks the affinity of benchmarks with a C program.
 * Updates related documentation.

The code was loosely based upon the previously reverted pinning code, where we pinned to one CPU only.

After all of this I ran a jruby benchmark and used top to check how the threads are being scheduled. Here is how it looks:

affin2

The benchmark threads have the correct affinity, and are on an adaptive ticks CPU, but all threads run on the same CPU. I think we were expecting the threads to be spread out over the CPUs in the affinity mask (all CPUs apart from CPU 0). The left hand pane in the screen shot is showing that the affinity mask is correctly set.

Having seen this, I wondered if taskset has caused all threads to all go on one core. I commented out the use of taskset and the accompanying sanity check and re-ran. This time all threads are scheduled on the boot processor.

What this means is that JRuby does not and has never utilised >1 CPU despite having loads of threads. Further, I can't say why removing pinning with a single CPU in the affinity mask (as we did a little while back) gave a speed-up.

Raising the PR for code review and discussion. Please don't merge yet.

Next port of call: write a small C program which spawns a load of threads. I want to know if threads simply default to being scheduled on the same CPU as their parent, or if the JVM is forcing all the threads on one core somehow.

Thanks

cc @ltratt for discussion

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,11 +24,12 @@ using, e.g. on Linux you should be able to run `sudo`.
 
 You need to have the following installed:
 
-  * Python2.7 (other versions of Python are not supported)
-  * a Java SDK (version 7)
-  * GNU make, a C compiler and libc (e.g. `sudo apt-get install build-essential`)
-  * cpufrequtils (e.g. `sudo apt-get install cpufrequtils`)
-  * cffi (e.g. `sudo apt-get install python-cffi`)
+  * Python2.7 (pre-installed in Debian)
+  * A Java SDK 7 (`openjdk-7-jdk` package in Debian)
+  * GNU make, a C compiler and libc (`build-essential` package in Debian)
+  * cpufrequtils (Linux only. `cpufrequtils` package in Debian)
+  * cffi (`python-cffi` package in Debian)
+  * taskset (Linux only. `util-linux` package in Debian)
 
 ### Kernel arguments
 
@@ -37,14 +38,24 @@ If your Linux bootloader is Grub, you can follow these steps:
 
   * Edit /etc/default/grub (e.g. `sudo gedit /etc/default/grub`)
   * Add `intel_pstate=disable` to `GRUB_CMDLINE_LINUX_DEFAULT`
+  * Add `isolcpus=x,y,z` to `GRUB_CMDLINE_LINUX_DEFAULT`, where 'x,y,z' is a
+    comma separated list of all logical CPUs apart from the boot processor
+    (i.e. CPU 0). This ensures that the adaptive tick cores are used soley for
+    benchmarks (see 'Tickless Mode Linux Kernel').
   * Run `sudo update-grub`
 
 ### Tickless Mode Linux Kernel
 
+The Linux kernel can run in ``tickless'' configurations, where under certain
+conditions regular tick interrupts can be avoided for a subset of logical CPUs.
+More info here:
+https://www.kernel.org/doc/Documentation/timers/NO_HZ.txt
+
 On a Linux system, Krun will insist that the kernel is running in
-"full tickless" mode. Tickless mode is a compile time kernel parameter, so if it is
-not enabled, you will need to build a custom kernel.  You can verify the
-tickless mode of the current kernel with:
+"full tickless" mode with the NO_HZ_FULL_ALL compile time flag. This will place
+all logical CPUs apart from the boot processor (i.e. CPU 0) into adaptive ticks
+mode. If this is not enabled, you will need to build a custom kernel. You can
+verify the tickless mode of the current kernel with:
 
 ```
 cat /boot/config-`uname -r` | grep HZ
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -852,6 +852,7 @@ class LinuxPlatform(UnixLikePlatform):
     def sanity_checks(self):
         UnixLikePlatform.sanity_checks(self)
         self._check_taskset_installed()
+        self._check_isolcpus()
         self._sanity_check_cpu_affinity()
 
     def _sanity_check_cpu_affinity(self):
@@ -862,6 +863,41 @@ class LinuxPlatform(UnixLikePlatform):
         vd = NativeCodeVMDef()
         util.spawn_sanity_check(self, ep, vd, "CPU affinity",
                                 force_dir=PLATFORM_SANITY_CHECK_DIR)
+
+    def _advise_isolcpus_arg(self, got_cpus, expect_cpus):
+        """Error out and guide user in isolating CPUs"""
+
+        arg = "isolcpus=%s" % ",".join(expect_cpus)
+        self._fatal_kernel_arg(
+            arg, "CPUs incorrectly isolated. Got: %s, expect: %s" %
+            (got_cpus, expect_cpus)
+        )
+
+    def _check_isolcpus(self):
+        """Checks the correct CPUs have been isolated.
+        (All but the boot processor)"""
+
+        expect_cpus = [str(x) for x in xrange(1, self.num_cpus)]
+        all_args = self._get_kernel_cmdline()
+
+        args = all_args.split(" ")
+        for arg in args:
+            if "=" not in arg:
+                continue
+
+            k, v = arg.split("=", 1)
+
+            if k != "isolcpus":
+                continue
+
+            break  # found the isolcpus arg
+        else:
+            self._advise_isolcpus_arg([], expect_cpus)  # exits
+
+        got_cpus = list(sorted(v.split(",")))
+
+        if expect_cpus != got_cpus:
+            self._advise_isolcpus_arg(got_cpus, expect_cpus)  # exits
 
 
 class DebianLinuxPlatform(LinuxPlatform):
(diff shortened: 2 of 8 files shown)


vext01:
Sorry the screenshot was wrong. Updated.

I've manually checked the htop side-pane shows the affinity mask contains CPUs 1,2,3 for all threads when run on this branch.

README.md
```diff
@@ -37,14 +38,19 @@ If your Linux bootloader is Grub, you can follow these steps:
...
+    benchmarks (see 'Tickless Mode Linux Kernel').
   * Run `sudo update-grub`
 
 ### Tickless Mode Linux Kernel
 
 On a Linux system, Krun will insist that the kernel is running in
-"full tickless" mode. Tickless mode is a compile time kernel parameter, so if it is
-not enabled, you will need to build a custom kernel.  You can verify the
-tickless mode of the current kernel with:
+"full tickless" mode with the NO_HZ_FULL_ALL compile time flag. This will place
```
snim2: This is great, but can we add a simple sentence that explains what `tickless mode` means?

krun/platform.py
```diff
@@ -221,7 +221,8 @@ def bench_cmdline_adjust(self, args, env_dct):
         changes = self.bench_env_changes()
         EnvChange.apply_all(changes, combine_env)
 
-        return self.adjust_env_cmd(combine_env) + args
+        return self.adjust_env_cmd(combine_env) + \
```
snim2: The docs for this method (which GitHub won't let me comment on directly) still say `libkruntime` rather than `libkrun`.

krun/platform.py
```diff
@@ -815,6 +833,72 @@ def get_reboot_cmd(self):
...
+        if self.num_cpus == 1:
+            fatal("not enough CPUs to pin")
+
+        cpus = ",".join([str(x) for x in xrange(1, self.num_cpus)])
+        return ["taskset", "-c", cpus]
+
+    def _check_taskset_installed(self):
+        from distutils.spawn import find_executable
+        if not find_executable("taskset"):
+            fatal("taskset not installed (needed for pinning).")
```
snim2: `taskset` is provided by package `util-linux` in Debian

vext01:
OK, I can explain why we saw the speedup.

If I switch back to master *and* remove `isolcpus` then the threads distribute themselves across the cores properly. I had forgotten to remove `isolcpus` so of course the only usable core was 0.

screenshot:

affin3

So what this means now is that taskset is causing all threads so be scheduled on one core. This is not what we want.

I'll probably need to make a minimal C example and raise a stack overflow question. Or ask one of the guys at Intel.

krun/platform.py
vext01: The shared object is still called libkruntime.so. Will make more clear.

vext01:
Alright, here's the next iteration of this.

I fixed your comments and switched benchmarks running on Linux to use the realtime fifo scheduler:

 * It correctly schedules threads on sets of isolated cores.
 * It is actually better suited to our needs.

Effectively when our benchmark is the only process running on a realtime scheduler (the remaining processes stick to the default scheduler), then it means it's pretty much impossible for another process to kick it off. Further the fifo scheduler has no notion of a timeslice.  For a general purpose system, this would be a pretty unfair, but for us it should work nicely.

We also set the thread prority to the max.

I notice that OpenBSD exposes the POSIX realtime scheduler options, but ignores them. I also note that OpenBSD does implement thread priorities, but you cannot set the priorities for the new thread and it's children ahead of time using a shell command (e.g. `chrt`) like in Linux. I think we would have to patch every thread spawn site in the VMs. For these reasons the thread priority changes here only affect Linux.

Please check my work carefully.

Reading material:
 * `man sched`
 * `man sched_get_priority_max`
 * `man sched_getscheduler`
 * `man chrt`

vext01:
Here's a screenshot of jruby/truffle threads now:

works

 * Threads distributed over isolated cores.
 * Thread priorities now display `RT` for real-time.

This is how the benchmark command line and wrapper looks now:

```
[2016-04-14 14:38:03 DEBUG] Writing out wrapper script to /tmp/krun_wrapper.dash
[2016-04-14 14:38:03 DEBUG] Wrapper script:
#!/bin/dash
ulimit -d 2097152 || exit $?
ulimit -s 8192 || exit $?
env LD_LIBRARY_PATH=/home/vext01/research/warmup_experiment/work/gcc-inst/lib64:/home/vext01/research/warmup_experimen
t/krun/krun/../libkrun JAVACMD=work/jvmci/jdk1.8.0_72-internal/product/bin/java taskset -c 1,2,3 work/jruby/bin/jruby 
-X+T -J-server /home/vext01/research/warmup_experiment/krun/iterations_runners/iterations_runner.rb /home/vext01/resea
rch/warmup_experiment/benchmarks/binarytrees/ruby/bench.rb 2000 25 1
exit $?
[2016-04-14 14:38:03 DEBUG] Execute wrapper: /usr/bin/sudo -u root chrt --fifo 99 /usr/bin/nice -n -20 /usr/bin/sudo -
u krun /bin/dash /tmp/krun_wrapper.dash
```

README.md
vext01: fixed

krun/platform.py
vext01: fixed

README.md
snim2: The new text is very nice!

krun/platform.py
snim2: Fixed.

krun/platform.py
```diff
@@ -498,6 +513,11 @@ class LinuxPlatform(UnixLikePlatform):
     KERNEL_ARGS_FILE = "/proc/cmdline"
     ASLR_FILE = "/proc/sys/kernel/randomize_va_space"
 
+    # The best realtime scheduler for out purposes. Also correctly schedules
```
snim2: s/out/our/g

krun/platform.py
```diff
@@ -815,6 +845,93 @@ def get_reboot_cmd(self):
...
+        if not find_executable("taskset"):
+            fatal("util-linix is not installed "
+                  "(needed for pinning and scheduler tweaking).")
+
+    def sanity_checks(self):
+        UnixLikePlatform.sanity_checks(self)
+        self._check_util_linux_installed()
+        self._check_isolcpus()
+        self._sanity_check_cpu_affinity()
+        self._sanity_check_scheduler()
```
snim2: What's the difference between `check_` and `sanity_check`?

vext01: sanity_check runs a fake benchmark.

krun/platform.py
```diff
@@ -815,6 +845,93 @@ def get_reboot_cmd(self):
...
+        """Checks the correct CPUs have been isolated.
+        (All but the boot processor)"""
+
+        expect_cpus = [str(x) for x in xrange(1, self.num_cpus)]
+        all_args = self._get_kernel_cmdline()
+
+        args = all_args.split(" ")
+        for arg in args:
+            if '=' not in arg:
+                continue
```
snim2: Mixing `'` and `"` arbitrarily here

platform_sanity_checks/Makefile
```diff
@@ -1,6 +1,7 @@
 OS != uname
 
-all: check_openbsd_malloc_options.so check_nice_priority.so
+all: check_openbsd_malloc_options.so check_nice_priority.so \
+	check_linux_cpu_affinity.so check_linux_scheduler.so
 
 .PHONY: clean
```
snim2: `all` should be in `.PHONY`

snim2:
I've finished reviewing, I haven't found any significant issues, just a few small things.

Am I right in thinking that the new tests exercise the two new C files?

krun/platform.py
snim2: That sounds like something we should document, but I'm not sure where. If we can't think of a good place I will raise an issue.

krun/platform.py
vext01: I never know what to use tbh. It's mixed in every python project i ever worked on.

vext01:
Tests do not run the platform sanity checks. These happen at runtime before each krun session.

vext01:
OK to start squashing?

krun/platform.py
snim2: I think it would be good to be consistent within the file, so can we change the `if '='` to `if "="`?

snim2:
Just that one pedantic thing about the quote marks, feel free to squash at the same time

krun/platform.py
vext01: fixed

snim2:
Squash?

snim2:
OK, happy for me to merge?

vext01:
go ahead

vext01:
Many thanks!

PR #195, opened 2016-04-15
Move a couple of preliminaries to the right place.

vext01:
These checks actually belong in `check_preliminaries()`, since they are not "sanity checks" (fake benchmarks).

Also add debug statements to be consistent.

OK?


snim2:
This looks like an improvement to me.

Issue #188, opened 2016-03-18
Make ETA available to pre/post execution commands.


vext01:
It would be nice to set an variable which contains the ETA in the pre/post execution environment. This would allow us to monitor the ETA over time (it becomes more accurate over time)

Related: #125

vext01:
I'm wondering why we decided this needs to be exposed to the pre/post-exec hooks. Would it suffice to simply store the ETA in the results file? (Under our scenario this is copied to bencher2 after each execution).

@ltratt, thoughts?

ltratt:
My suggestion is two env variables, something like `ETA_GENERATED_AT` and `ETA_IS` (better names are probably not hard).

vext01:
This is implemented.

Issue #125, opened 2015-12-14
ETA doesn't account for startup sleeps in reboot mode


vext01:
ETAs in reboot mode are currently a little optimistic due to the fact that they don't account for the sleeps that allow the system to come up.

ltratt:
If we're being clever, krun can notice that it's done N executions and count how long they all took. It can then subtract that amount from the time between "now" and "when we started the experiments", and then divide it by the number of process executions. This will give a good average of how long it took krun to reboot and start running a benchmark.

vext01:
Assigned "camera ready" milestone. For the real experiment, we want to minimise the SSH activity to the machine during benchmarking.

vext01:
Duplicate of #127

PR #196, opened 2016-04-18
Add environment vars for ETA, and add sleeps to ETA.

vext01:
Opening for discussion.

This adds two environment variables to the pre/post hook environment:

 * `KRUN_ETA_DATUM`
 * `KRUN_ETA_VALUE`

It also attempts to take sleeps for sync() and waiting for the system to come up.

I hacked the sync wait to 5 seconds, and used the following hooks:

```
PRE_EXECUTION_CMDS = [                                                          
    'echo "At ${KRUN_ETA_DATUM} ETA is ${KRUN_ETA_VALUE}" >> eta_pre.log'       
]                                                                               
POST_EXECUTION_CMDS = [                                                         
    'echo "At ${KRUN_ETA_DATUM} ETA is ${KRUN_ETA_VALUE}" >> eta_post.log'      
]                                                                               
```

Then I ran without reboots.

Seems to work at first glance:

```
wilfred:examples] cat eta_post.log 
At 2016-04-18 16:35:13 ETA is Unknown. Known in 1 process executions.
At 2016-04-18 16:35:31 ETA is 2016-04-18 16:36:47
At 2016-04-18 16:35:41 ETA is 2016-04-18 16:36:42
At 2016-04-18 16:35:46 ETA is 2016-04-18 16:36:37
At 2016-04-18 16:35:56 ETA is 2016-04-18 16:36:32
At 2016-04-18 16:36:02 ETA is 2016-04-18 16:36:27
At 2016-04-18 16:36:12 ETA is 2016-04-18 16:36:22
At 2016-04-18 16:36:30 ETA is 2016-04-18 16:36:30
```
But what confuses me is that the estimates are frequently too pessimistic (estimates are frequently longer than in reality), whereas (unless the example benchmark had a variance in the order of seconds) I would expect the estimates to be too optimistic, as there are other delays (e.g. temperature waits, time for the hooks to execute), which are not (and cannot be easily) accounted for in the ETA.

I must be missing something. Will continue to work on this, but shout if you have an idea.

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -1,6 +1,7 @@
-from krun.time_estimate import TimeEstimateFormatter
+from krun.time_estimate import TimeEstimateFormatter, now_str
 from krun.results import Results
 from krun import util
+from krun.platform import SYNC_SLEEP_SECS
 
 from collections import deque
 from logging import warn, info, error, debug
@@ -123,12 +124,6 @@ class ExecutionScheduler(object):
 
         self.log_path = self.config.log_filename(self.resume)
 
-        # Used for pre/post cmd hooks
-        self.pre_post_cmds_extra_env = {
-            "KRUN_RESULTS_FILE": self.config.results_filename(),
-            "KRUN_LOG_FILE": self.config.log_filename(resume=True),
-        }
-
     def set_eta_avail(self):
         """call after adding job before eta should become available"""
         self.eta_avail = len(self)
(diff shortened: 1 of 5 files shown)


vext01:
Fixed :)

With fake reboots, 10 second reboot wait, 5 second sync wait, no temperature waits:
```
wilfred:examples] cat eta_post.log
At 2016-04-18 17:54:13 ETA is Unknown. Known in 1 process executions.
At 2016-04-18 17:54:29 ETA is 2016-04-18 17:56:15
At 2016-04-18 17:54:49 ETA is 2016-04-18 17:56:15
At 2016-04-18 17:55:05 ETA is 2016-04-18 17:56:16
At 2016-04-18 17:55:25 ETA is 2016-04-18 17:56:16
At 2016-04-18 17:55:41 ETA is 2016-04-18 17:56:16
At 2016-04-18 17:56:01 ETA is 2016-04-18 17:56:17
At 2016-04-18 17:56:17 ETA is 2016-04-18 17:56:17
```

I would document the new environment variables, but I think there is no existing documentation for the config file at all(?). For now I have added a comment describing the variables to the example config file.

snim2:
You are right, the example config file contains the only "documentation", so we should make sure that everything that *can* go in a config file goes in there. Do you think we should add some documentation for that in the README or add a template configuration for users to copy?

vext01:
Requires some thought, but I think this should be a separate PR.

Mind if we proceed to squash an merge?

vext01:
Fixes #188, fixes #125.

vext01:
Rebase and squash. Also add a commit that was supposed to be in the last PR. Oops.

krun/scheduler.py
```diff
@@ -163,6 +158,7 @@ def __len__(self):
 
     def get_estimated_exec_duration(self, key):
         previous_exec_times = self.results.eta_estimates.get(key)
+
```
vext01: blank line!

vext01:
Fixed blank line, good to go.

PR #197, opened 2016-04-19
More info in results

### LINKED ISSUE #141: Improving the "benchmark started" email
Since users will start a benchmark run and not want to ssh into the b/marking box again, it would be nice for the benchmark-started email to contain a few more details. This would make it easier to catch config mistakes after a run has started.

An easy fix for this issue would just be to email out the contents of the config file.

vext01:
This change:

 * Adds debug output for the CLI args (does not appear in production mode)
 * Adds CLI args to the audit, which appears in the results file.
 * Inserts Krun git hash into the audit.

The latter two are re-written upon each reboot. I think that is good enough for now.

OK?

Fixes #141


snim2:
lgtm

Issue #157, opened 2016-02-18
Deprecate --develop and --dryrun?


vext01:
It seems that whenever I am testing krun and reach for `--develop` or `--dryrun`, I find that I want to disable a different subset of production features than these options provide.

I wonder if we should instead split these options out into many individual flags, such as: `--x-fake-process-executions`, `--x-no-temperature-checks`, `-x-fake-reboots`, ...

ltratt:
Shouldn't these just be options in the config file rather than impossible to remember CLI switches?

vext01:
I don't mind where they live, but what makes config file options easier to remember?

ltratt:
IMHO, the problem with commands that have lots of CLI switches is a) they start breeding like rats b) they interact like spaghetti c) documenting this becomes impossible. Of course, this isn't black and white, and there are many acceptable shades of grey.

vext01:
I'm neutral about where the options live, just as long as we have fine-grained control over the various options that skip production-mode checks. Since you prefer config variables, we could roll with this.

I'm also going to downgrade the priority as I think I overestimated this.

snim2:
Whether the options are in the config file or the CLI switches, there is still an overhead on the user to remember all the options. To my mind, the user will probably want to use the same config file in different ways, and we probably want to minimise the number of times the user has to edit the configuration, when they aren't changing their VM / language / benchmark options. 

The problem we have with the current CLI switches is not so much that there are too many of them, but in the case of `--develop` and `--dry-run` we named the switches in a very confusing way,

vext01:
Having worked more on krun, and having watched @snim2 work on travis, I think that the debugging options we are proposing to add here are best as switches.

 * When developing krun, it's easier to tack on switches quickly, rather than open an editor, locate the right line...
 * If you forget the name of a switch,  `-h` will list them all.
 * You can't forget to document a switch, as the argument parser requires at-leas t a short description of each switch.
 * Tacking on command line switches to make krun on travis is much easier than having a separate config file for travis, or worse patching existing config files.

Regarding spaghetti interactions, I think this is a separate issue. The logic will be the same regardless of whether the options are a switch or a config file option. To avoid spaghetti, we have to try hard to keep the switches orthogonal. E.g., never `if not fake_reboots and no_sync:` like conditions.

Off the top of my head, these are the switches I would like implemented:
 * `--fake-reboots`
 * `--no-sync`
 * `--no-temp-wait`
 * `--no-pinning`
[...]

PR #198, opened 2016-05-10
Add a pinning switch to the config file.

vext01:
This change:

 * Turns off pinning by default
 * Adds a switch in the config file to re-enable pinning (in its current (broken) state).

Some of the plumbing here will be of help to #157, which I would like to push on with also.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -34,10 +34,11 @@ class BasePlatform(object):
     TEMP_TOO_HOT = 1
     TEMP_TOO_COLD = 2
 
-    def __init__(self, mailer):
+    def __init__(self, mailer, config):
         self.developer_mode = False
         self.mailer = mailer
         self.audit = OrderedDict()
+        self.config = config
 
         # Temperatures should always be a dict mapping a descriptive name of
         # the sensor to a platform dependent linear temperature measurement.
@@ -224,8 +225,12 @@ class BasePlatform(object):
         changes = self.bench_env_changes()
         EnvChange.apply_all(changes, combine_env)
 
-        return self.adjust_env_cmd(combine_env) + \
-            self.pin_process_args() + args
+        new_args = self.adjust_env_cmd(combine_env)
+        if self.config.ENABLE_PINNING:
+            new_args += self.pin_process_args()
+        new_args += args
+
+        return new_args
 
     @abstractmethod
     def _change_user_args(self, user="root"):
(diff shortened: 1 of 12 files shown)


krun/platform.py
```diff
@@ -963,9 +977,12 @@ def _check_isolcpus(self):
 
             break  # found the isolcpus arg
         else:
-            self._advise_isolcpus_arg([], expect_cpus)  # exits
+            v = ""  # didn't find a isolcpus arg
```
snim2: Can we rename `k` and `v` so they aren't one letter variables?

krun/platform.py
```diff
@@ -963,9 +977,12 @@ def _check_isolcpus(self):
...
 
-        got_cpus = list(sorted(v.split(",")))
+        if v == "":
+            got_cpus = []
+        else:
+            got_cpus = list(sorted(v.split(",")))
```
snim2: I would be inclined to move `got_cpus = ` into the previous blocks, not just to avoid the extra if/else but by the time I got to here, I have forgotten exactly what `v` should look like. YMMV.

krun/tests/test_vmdef.py
```diff
@@ -56,7 +57,8 @@ def test_env_ctor0002(self):
     def test_sync_disks0001(self, monkeypatch):
         """Check disk sync method is called"""
 
-        platform = MockPlatform(None)
+        config = Config()
+        platform = MockPlatform(None, config)
         ep = EntryPoint("test")
         vm_def = PythonVMDef('/dummy/bin//python')
```
snim2: Why two slashes?

krun/tests/test_vmdef.py
```diff
@@ -77,7 +79,8 @@ def test_sync_disks0002(self, monkeypatch):
         """We throw away the results from sanity checks, so there's no need to
         sync disks (and wait)."""
 
-        platform = MockPlatform(None)
+        config = Config()
+        platform = MockPlatform(None, config)
         ep = EntryPoint("test")
         vm_def = PythonVMDef('/dummy/bin//python')
```
snim2: Why two slashes?

snim2:
Generally this looks OK to me.

Since we have this problem with pinning, even though the problem lies upstream, I think we should have an Issue for it in this repo.

It is a shame to have `Config` and `Platform` so closely coupled, maybe if/when the pinning thing is resolved, we can divorce the two.

krun/platform.py
vext01: I was trying to protect against "isolcpus=" in the kernel cmdline, but actually I don't think that is valid.

vext01:
Regarding the coupling, I perhaps wouldn't bother trying to decouple. If we are going to have switches in the config file to turn on and off certain features (#157), then you will need the config instance around. The alternative is to pass the config around a lot, which is a pain.

vext01:
Try this.

Note that this code is likely to change soon anyway, if cgroups work out for us.

snim2:
Much neater! Do you want to squsah?

vext01:
splat

PR #202, opened 2016-05-12
New Linux pinning strategy using a cset shield. Pin by default.

vext01:
Here's a start on the new pinning code. Please do not merge yet. I would like to test it some more and look for non-linux breakage.

Work in progress...

I recommend skimming the cset tutorial to see roughly what this is all about:
https://rt.wiki.kernel.org/index.php/Cpuset_Management_Utility/tutorial

In short, if the system has N cores, we create a N-1 core shield. All user tasks are migrated off the shield (a.k.a. the "user cset") onto the remaining core (the "system cset"). We move many kernel threads as possible too (you can't move them all). The N-1 cores are the same cores that are in adaptive tick mode. The core not in the shield is the boot processor.

Also:
 * We can now use the default scheduling policy `SCHED_OTHER`. A platform sanity check is checking for this.
 * This scheduler uses only nice priorities, so I've removed the static priority code.
 * Turn pinning back on by default.
 * Check `isolcpus` is not used, as this will interfere with the shield.

It's a big change and there's bound to be mistakes. Please check carefully.

### DIFF
--- a/examples/example.krun
+++ b/examples/example.krun
@@ -75,5 +75,5 @@ TEMP_READ_PAUSE = 1
 #PRE_EXECUTION_CMDS = ["sudo service cron stop"]
 #POST_EXECUTION_CMDS = ["sudo service cron start"]
 
-# CPU pinning (off by default)
-#ENABLE_PINNING = False
+# CPU pinning (on by default)
+#ENABLE_PINNING = True
--- a/README.md
+++ b/README.md
@@ -29,24 +29,22 @@ You need to have the following installed:
   * GNU make, a C compiler and libc (`build-essential` package in Debian)
   * cpufrequtils (Linux only. `cpufrequtils` package in Debian)
   * cffi (`python-cffi` package in Debian)
-  * taskset (Linux only. `util-linux` package in Debian)
+  * cset (for pinning on Linux only. `cpuset` package in Debian)
+
+Note that to use pinning on Linux, `cset shield` must be in a working state.
+Some Linux distributions have been known to ship with this functionality
+broken. See the `cset` tutorial for information on how to test `cset shield`:
+https://rt.wiki.kernel.org/index.php/Cpuset_Management_Utility/tutorial
 
 ### Kernel arguments
 
-If you are using a Linux system, you will need to set some kernel arguments.
-If your Linux bootloader is Grub, you can follow these steps:
+If you are using a Linux system, you will need to set a kernel arguments to
+disable Intel P-states. If your Linux bootloader is Grub, you can follow these
+steps:
 
   * Edit /etc/default/grub (e.g. `sudo gedit /etc/default/grub`)
   * Add `intel_pstate=disable` to `GRUB_CMDLINE_LINUX_DEFAULT`
-  * If you want to use CPU pinning, add `isolcpus=x,y,z` to
-    `GRUB_CMDLINE_LINUX_DEFAULT`, where 'x,y,z' is a comma separated list of
-    all logical CPUs apart from the boot processor.
-    (i.e. CPU 0). This ensures that the adaptive tick cores are used soley for
-    benchmarks (see 'Tickless Mode Linux Kernel').
   * Run `sudo update-grub`
-
-Note that CPU pinning is off by default due to a bug in the Linux kernel:
-https://bugzilla.kernel.org/show_bug.cgi?id=116701
 
 ### Tickless Mode Linux Kernel
 
--- a/krun/config.py
+++ b/krun/config.py
@@ -27,7 +27,7 @@ class Config(object):
         self.HEAP_LIMIT = None
         self.STACK_LIMIT = None
         self.TEMP_READ_PAUSE = 60
-        self.ENABLE_PINNING = False
+        self.ENABLE_PINNING = True
 
         self.PRE_EXECUTION_CMDS = []
         self.POST_EXECUTION_CMDS = []
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -225,12 +225,7 @@ class BasePlatform(object):
         changes = self.bench_env_changes()
         EnvChange.apply_all(changes, combine_env)
 
-        new_args = self.adjust_env_cmd(combine_env)
-        if self.config.ENABLE_PINNING:
-            new_args += self.pin_process_args()
-        new_args += args
-
-        return new_args
+        return self.adjust_env_cmd(combine_env) + args
 
     @abstractmethod
     def _change_user_args(self, user="root"):
@@ -278,10 +273,6 @@ class BasePlatform(object):
 
     @abstractmethod
     def pin_process_args(self):
-        pass
-
-    @abstractmethod
-    def change_scheduler_args(self):
         pass
 
 
@@ -506,9 +497,6 @@ class OpenBSDPlatform(UnixLikePlatform):
     def pin_process_args(self):
         return []  # not supported on OpenBSD
 
-    def change_scheduler_args(self):
-        return []  # no control over scheduler
-
 
 class LinuxPlatform(UnixLikePlatform):
     """Deals with aspects generic to all Linux distributions. """
@@ -520,12 +508,8 @@ class LinuxPlatform(UnixLikePlatform):
     CPU_SCALER_FMT = "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_driver"
     KERNEL_ARGS_FILE = "/proc/cmdline"
     ASLR_FILE = "/proc/sys/kernel/randomize_va_space"
-
-    # The best realtime scheduler for our purposes. Also correctly schedules
-    # threads when a process is forced onto isolated cores:
-    # http://stackoverflow.com/questions/36604360/why-does-using-taskset-to-run-a-multi-threaded-linux-program-on-a-set-of-isolate
-    SCHED_ALGO = "fifo"
-    SCHED_RT_RUNTIME_US = "/proc/sys/kernel/sched_rt_runtime_us"
+    CSET_CMD = "/usr/bin/cset"
+    USER_CSET_DIR = "/cpusets/user"
 
     # Expected tickless kernel config
     #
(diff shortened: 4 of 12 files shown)


vext01:
I added some tests to OpenBSD for added confidence.

Here's a log snippet from running JRubyTruffle from the warmup experiment on a Linux host:
```
[2016-05-12 12:37:30 DEBUG] Attached log file: jruby_20160512_123730.log
[2016-05-12 12:37:30 DEBUG] Krun invoked with arguments: ['./krun/krun.py', '--debug', 'debug', 'jruby.krun']
...
[2016-05-12 12:37:30 DEBUG] Check cset is installed
[2016-05-12 12:37:30 DEBUG] Check isolcpus not in use
[2016-05-12 12:37:30 DEBUG] create/check/remove cset shield
[2016-05-12 12:37:30 DEBUG] execute shell cmd: /usr/bin/sudo -u root /usr/bin/cset shield -c 1-3
[2016-05-12 12:37:30 DEBUG] cset: --> shielding modified with:
cset: "system" cpuset of CPUSPEC(0) with 102 tasks running
cset: "user" cpuset of CPUSPEC(1-3) with 0 tasks running
[2016-05-12 12:37:30 DEBUG] execute shell cmd: /usr/bin/cset shield -k on
[2016-05-12 12:37:30 DEBUG] cset: --> activating kthread shielding
cset: kthread shield activated, moving 56 tasks into system cpuset...
[==================================================]%
cset: **> 56 tasks are not movable, impossible to move
cset: done
...
[2016-05-12 12:38:35 DEBUG] Writing out wrapper script to /tmp/krun_wrapper.dash
[2016-05-12 12:38:35 DEBUG] Wrapper script:
#!/bin/dash
ulimit -d 2097152 || exit $?
ulimit -s 8192 || exit $?
env LD_LIBRARY_PATH=/home/vext01/research/warmup_experiment/work/gcc-inst/lib64:/home/vext01/research/warmup_experiment/krun/krun/../libkrun JAVA
CMD=work/jvmci/jdk1.8.0_72-internal/product/bin/java work/jruby/bin/jruby -X+T -J-server /home/vext01/research/warmup_experiment/krun/iterations_
runners/iterations_runner.rb /home/vext01/research/warmup_experiment/benchmarks/binarytrees/ruby/bench.rb 2000 25 1
exit $?
[2016-05-12 12:38:35 DEBUG] Execute wrapper: /usr/bin/sudo -u root /usr/bin/nice -n -20 /usr/bin/sudo -u root /usr/bin/cset shield -e -- /usr/bin
/sudo -u krun /bin/dash /tmp/krun_wrapper.dash
[2016-05-12 12:38:35 DEBUG] sync disks...
[2016-05-12 12:39:08 INFO] stderr: [iterations_runner.rb] iteration 1/2000
...
```

You can see the shield being created and tasks being migrated, then you see the wrapper script being executed into the shield.

And here's what it looks like in htop:

jruby-pinned

Here, the shield is CPUs 1-3.

Notice the JRuby threads are being spread over the shielded cores, and that the cores all appear to be quite busy (look at the {0,1,2,3} bars at the top of the display). In `SCHED_OTHER` the static priority (`PRI` column) is derived from the nice value, which for our benchmarks in -20 (highest). This is why the benchmarks have a different static priority.

You might notice some other non-benchmark tasks shown as running on the sheild. This is stale information. Those tasks have not woken up since the shield was created, so their `CPU` field is showing the last CPU they ran on. When they wake (and if)  up, they will be on core 0 outside of the shield. Discussion of this here: http://serverfault.com/questions/775906/how-does-cpu-affinity-interact-with-cgroups-in-linux

All looks good?

examples/example.krun
```diff
@@ -75,5 +75,5 @@ TEMP_READ_PAUSE = 1
 #PRE_EXECUTION_CMDS = ["sudo service cron stop"]
 #POST_EXECUTION_CMDS = ["sudo service cron start"]
 
-# CPU pinning (off by default)
-#ENABLE_PINNING = False
+# CPU pinning (on by default)
+#ENABLE_PINNING = True
```
snim2: Does this need to be mentioned in README.md?

vext01: We are not mentioning config options in README.md at all. You are right that I do need to update the README a little though.

vext01: Actually, maybe it's time to start making some more structured documentation in general. Not sure what format would be best... man, tex, ...?

snim2:
Can I test this with a normal Linux kernel?

vext01:
Maybe with --develop, otherwise comment out code which stops it running. Alternatively try on becnher3 or 5.

examples/example.krun
snim2: This entirely depends on how you think Krun should be distributed. Man pages work well if you want to distribute it as a UNIX utility, readthedocs if you expect Krun to go in the Cheese Shop.

snim2:
OK, I don't get any errors from the unit tests, and I get this from running the example config:

```
/usr/bin/nice: /usr/bin/cset: No such file or directory
```

snim2:
OK, I can't see anything obviously silly in the code, but it seems that a platform check isn't working on my machine. I'll install cset when the platform check works.

vext01:
weird, I'll take a look.

examples/example.krun
ltratt: At the moment I wouldn't worry too much about this at the moment. In the long run, I tend to prefer man pages, but I don't feel strongly about it.

vext01: I like manual pages too, at-least for UNIX utilities. The thing is, Krun is not installable per-se, so we won't be placing manual pages in the man path. This would mean to read the manual you would have to do `mandoc /path/to/krun.1 | less` for example.
Kinda awkward, but I guess I could live with it.

Definitely split the docs down into many pages though. One for switches, one for config file, one discussing how to set it up.

ltratt: Point taken about a man page. Until the thing gets too unwieldy (which will be a while), I'd prefer a single doc page though.

vext01:
Can you do a `which cset` on your machine? My theory is that it has been installed in another path...

snim2:
It isn't installed at all....

vext01:
In a REPL, what does this code return:

```
from distutils.spawn import find_executable
find_executable("cset")
```

vext01:
Should be `None`

snim2:
```python
>>> find_executable("cset")
>>> find_executable("which")
'/usr/bin/which'
>>> type(find_executable("cset"))
<type 'NoneType'>
>>> 
```

But, now I get:

```bash
$ PYTHONPATH=.. ../krun.py example.krun 
[2016-05-12 15:16:08: ERROR] cset is not installed (needed for pinning).
```

from Krun, so maybe I did something silly like run Krun from a virtualenv. Ugh, sorry.

vext01:
Oh, OK :)

vext01:
Pushed a commit which updates the README.

snim2:
OK, so I installed `cpuset` and now I get:

```bash
$ PYTHONPATH=.. ../krun.py example.krun 
[2016-05-12 15:24:25: ERROR] Command failed: '/usr/bin/sudo -u root /usr/bin/cset shield -c 1-3'
stdout:
cset: --> failed to create shield, hint: do other cpusets exist?
cset: **> [Errno 22] Invalid argument

stderr:


```

Do I need to reboot?

vext01:
Theory: newer versions of cpuset don't allow you to create a shield if one is already up.

Try this in a shell:
```
$ cset shield -r
$ cset shield -c n
$ cset shield -c n
```

Where `n` is a cpu number in your system. Does it complain the second time?

snim2:
Baffling:

```bash
$ sudo cset shield -r
cset: --> deactivating/reseting shielding
cset: **> shielding not active on system

$ sudo cset shield -c 1
cset: --> failed to create shield, hint: do other cpusets exist?
cset: **> [Errno 22] Invalid argument

$ sudo cset shield -c n
cset: **> CPUSPEC "n" contains invalid charaters: n
```

vext01:
So it fails on the first call. That's weird. `cset` looks broken. It should work like this:

```
vext01@bencher3:~$ sudo cset shield -r
cset: --> deactivating/reseting shielding
cset: **> shielding not active on system
vext01@bencher3:~$ sudo cset shield -c 1
cset: --> activating shielding:
cset: moving 106 tasks from root into system cpuset...
[==================================================]%
cset: "system" cpuset of CPUSPEC(0,2-3) with 106 tasks running
cset: "user" cpuset of CPUSPEC(1) with 0 tasks running
vext01@bencher3:~$ sudo cset shield -c 1
cset: --> shielding modified with:
cset: "system" cpuset of CPUSPEC(0,2-3) with 106 tasks running
cset: "user" cpuset of CPUSPEC(1) with 0 tasks running
```

Can you have a play around and see if you can figure out why it doesn't work like this on your system?

snim2:
Rebooting didn't help. I think we have different versions of cset with different syntaxes. I installed version 1.5.6 and this works:

```bash
$ sudo cset set --cpu=1 --set=new_set
cset: --> created cpuset "new_set"

$ sudo cset set --destroy new_set
cset: --> processing cpuset "new_set", moving 0 tasks to parent "/"...
cset: --> deleting cpuset "/new_set"
cset: done
```

vext01:
Ah. So the version in the ubuntu package manager is old? Or there's more than one?

vext01:
wait, `cpuset set` is not what we want. We want to make a shield.

snim2:
OK, so the new news is that cset seems to work on my desktop (with Ubuntu 16.04) but not on my laptop (Ubuntu 15.10).

vext01:
So what do you recommend? Since this looks like an Ubuntu 15 bug, I'm inclined to do nothing to the Krun code. There's not much we can do if cset is broken on a specific ubuntu version.

snim2:
I recommend adding something to the documentation so that Ubuntu 15 users know what to expect. I think that's the best we can do.

ltratt:
I don't think it's worth documenting that it doesn't work on an old version of Ubuntu. Just say "requires Ubuntu 16 or later" (or maybe simply "this does require an upto date Linux distro").

vext01:
I think we should say something like "On linux requires a working `cset`. Some Linux distributions have been known to ship with a broken cset".

I think this covers all cases. We have seen both broken debian and ubuntu installs, and it is possible that new releases may be broken too.

ltratt:
That wording is fine with me.

snim2:
+1

vext01:
So if this looks good I will squash.

snim2:
Sure, once you have done that and the PR is up to date with master I will merge.

vext01:
Ah yes, I didn't notice i was OOD.

vext01:
splat

PR #205, opened 2016-05-16
Improve examples

### LINKED ISSUE #191: Remove Java from the example.krun config and benchmarks
Having Java as a dependency on the examples is problematic for users who just want to quickly see if they can get Krun working. To compile Krun with Java enabled users need to do this:

```bash
 make JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/linux"' \
    JAVA_LDFLAGS=-L${JAVA_HOME}/lib ENABLE_JAVA=1
```

rather than just running `make`.

It would be nice to have a variety of language-variants and benchmarks in the examples, to demonstrate how to use the configuration file. It would be preferable to avoid C. It would be preferable if users on most mainstream UNIX platforms did not have to install anything new just to run the examples.

snim2:
* Default example uses Native and Python VMs
  * Separate example for those who want to use Java
  * New Travis config which tests C, Java and Python examples, but only uses one process execution and iteration
  * Brings documentation up to date

Fixes #191

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,16 @@ The `examples/` directory contains a simple experiment using Krun.
 This is a good starting point for setting up your own Krun configuration.
 
 The example here contains two benchmark programs (*nbody* and *dummy*),
-executed on two VMs (*cPython* and a standard *JVM* such as HotSpot).
+executed on two VMs (*cPython* and native code with no VM). There is a separate
+example with Java (using *JVM* such as HotSpot), which requires you to set up
+some environment variables before you compile Krun.
 Each benchmark is run for 5 iterations on the same VM, then the VM is
 restarted and the benchmark is re-run for another 5 iterations.
 We say that the experiment runs 2 *process executions* and 5 *in-process iterations* of
 each benchmark.
 
-This configuration can be found in the file `examples/example.krun`.
+This configuration can be found in the file `examples/example.krun`. The
+example with Java can be found in `examples/java.krun`.
 
 ## Step 1: prepare the benchmarking machine
 
@@ -25,11 +28,13 @@ using, e.g. on Linux you should be able to run `sudo`.
 You need to have the following installed:
 
   * Python2.7 (pre-installed in Debian)
-  * A Java SDK 7 (`openjdk-7-jdk` package in Debian)
   * GNU make, a C compiler and libc (`build-essential` package in Debian)
   * cpufrequtils (Linux only. `cpufrequtils` package in Debian)
   * cffi (`python-cffi` package in Debian)
   * cset (for pinning on Linux only. `cpuset` package in Debian)
+
+If you want to benchmark Java, you will also need:
+  * A Java SDK 7 (`openjdk-7-jdk` package in Debian)
 
 Note that to use pinning on Linux, `cset shield` must be in a working state.
 Some Linux distributions have been known to ship with this functionality
@@ -120,25 +125,33 @@ $ git clone https://github.com/softdevteam/krun.git
 $ cd krun
 ```
 
-## Step 3: set `JAVA_HOME`
-
-```bash
-$ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64/
-```
-## Step 4: Build Krun
+## Step 3: Build Krun
 
 The Krun Makefile honours the standard variables: `CC`, `CPPFLAGS`, `CFLAGS`
 and `LDFLAGS`. For example, if you wish to use `clang` rather than `gcc` you
-can append `CC=/bin/clang` to the options here:
+can append `CC=clang` to the `make` command below. You should build Krun
+by invoking GNU make:
 
 ```bash
 $ pwd
 .../Krun
-$ make JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/linux"' \
+$ make  # gmake on non-Linux platforms.
+```
+
+If you want to benchmark Java programs, you will also need to set the
+`JAVA_HOME` environment variable, and build Krun with some extra flags. The
+invocation below comes from a Ubuntu Linux machine, you may need to replace
+some paths and invoke `gmake` on other platforms:
+
+```bash
+$ pwd
+.../Krun
+$ env JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/ make  \
+    JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/linux"' \
     JAVA_LDFLAGS=-L${JAVA_HOME}/lib ENABLE_JAVA=1
 ```
 
-## Step 5: Build the benchmarks
+## Step 4: Build the benchmarks
 
 ```bash
 $ cd examples/benchmarks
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -120,25 +125,30 @@ $ git clone https://github.com/softdevteam/krun.git
 $ cd krun
 ```
 
-## Step 3: set `JAVA_HOME`
+## Step 3: Build Krun
+
+The Krun Makefile honours the standard variables: `CC`, `CPPFLAGS`, `CFLAGS`
+and `LDFLAGS`. For example, if you wish to use `clang` rather than `gcc` you
+can append `CC=/bin/clang` to the `make` command here:
```
vext01: Minor nit-pick. A clang binary is unlikely to appear under `/bin`, on linux it's in `/usr/bin` and on BSD it's under `/usr/local/bin`. To avoid putting any path, you could put something like `CC=/your/path/to/clang` and that would probably suffice.

README.md
```diff
@@ -120,25 +125,30 @@ $ git clone https://github.com/softdevteam/krun.git
...
-## Step 4: Build Krun
 
-The Krun Makefile honours the standard variables: `CC`, `CPPFLAGS`, `CFLAGS`
-and `LDFLAGS`. For example, if you wish to use `clang` rather than `gcc` you
-can append `CC=/bin/clang` to the options here:
+If you want to benchmark Java programs, you will also need to set the
+`JAVA_HOME` environment variable, and build Krun with some extra flags:
 
 ```bash
+$ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64/
```
vext01: Similarly for the path here, but I'm torn on this one. The path is obscure and having that concrete example will help users find the Java cruft. Any thoughts?

README.md
```diff
@@ -159,6 +178,18 @@ $ PYTHONPATH=../ ../krun.py example.krun
 You should see a log scroll past, and results will be stored in the file:
 `../krun/examples/example_results.json.bz2`.
 
+If you wan to try the example Java benchmarks, there is a separate
```
vext01: Typo "wan"

README.md
snim2: I prefer to keep this one. It isn't always clear with Java env vars whether they should point to the top level directory or `../jre/` or somewhere else.

README.md
```diff
@@ -159,6 +178,18 @@ $ PYTHONPATH=../ ../krun.py example.krun
...
+configuration file called `java.krun`, which contains configuration for the Java
+and Python examples:
+
+```bash
+$ pwd
+.../krun/examples
+$ PYTHONPATH=../ ../krun.py java.krun
```
vext01: Is the PYTHONPATH needed? I never set that.

README.md
ltratt: Do we really want to encourage people to use `export`? It's rather dangerous...

examples/benchmarks/Makefile
```diff
@@ -2,19 +2,44 @@ JAVAC ?=	javac
...
 		cd ${i}/java && \
 		rm *.class; \
-		cd ../../ \
+		cd ../../; \
+	)
+
+clean_c:
+	$(foreach i, ${BENCHMARKS}, \
+		cd  ${i}/c && \
+		rm *.so; \
```
vext01: If you add a `-f` to `rm` the target will not fail if the shared object does not exist.

README.md
ltratt: In general won't `CC=clang` work (and automatically find it in the user's path)?

examples/example.krun
```diff
@@ -77,3 +73,4 @@ TEMP_READ_PAUSE = 1
 
 # CPU pinning (on by default)
 #ENABLE_PINNING = True
+
```
vext01: Intentional blank line?

README.md
vext01: Yes, but I don't think we should encourage this, as it's very easy to invoke the wrong compiler.

ltratt: I use this all the time. How could it invoke the wrong compiler?

README.md
vext01: I prefer to pass them as make variables, but you could also do `env JAVA_HOME=... make`

vext01:
Asides from nitpicking this looks good. I'm going to test it now.

README.md
vext01: For example, on the openbsd machine, there are two system gccs and a manually built gcc. Being the paranoid I am, I would always list the absolute path to the gcc I want, and I would recommend our users do the same. This could save a week of benchmarking time.

ltratt: We can't protect people from everything, and trying to do so can sometimes make things worse.

That said, I don't care that much in this case (personally, I think we can assume that people know what to do with the `CC` variable. I don't even think we need to give an example).

vext01: Yeah, users will break anything and everything.

I'm actually OK with just mentioning the variable too tbh.

README.md
```diff
@@ -120,25 +125,30 @@ $ git clone https://github.com/softdevteam/krun.git
...
+
+The Krun Makefile honours the standard variables: `CC`, `CPPFLAGS`, `CFLAGS`
+and `LDFLAGS`. For example, if you wish to use `clang` rather than `gcc` you
+can append `CC=/bin/clang` to the `make` command here:
 
 ```bash
-$ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64/
+$ pwd
+.../Krun
+$ make
```
vext01: All `make` would need to be `gmake` on most non-linuxes. Maybe a quick note near the top. The user needs to invoke GNU make.

README.md
```diff
@@ -120,25 +125,30 @@ $ git clone https://github.com/softdevteam/krun.git
...
-and `LDFLAGS`. For example, if you wish to use `clang` rather than `gcc` you
-can append `CC=/bin/clang` to the options here:
+If you want to benchmark Java programs, you will also need to set the
+`JAVA_HOME` environment variable, and build Krun with some extra flags:
 
 ```bash
+$ export JAVA_HOME=/usr/lib/jvm/java-7-openjdk-amd64/
 $ pwd
 .../Krun
 $ make JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/linux"' \
```
vext01: I had to replace `linux`with `openbsd` here to build on my laptop. Maybe a note about this somewhere.

vext01: FWIW, here's the full invokation I used (once `javac` was in the path):

```
wilfred:krun] env JAVA_HOME=/usr/local/jdk-1.7.0/ gmake JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/openbsd"' \
    JAVA_LDFLAGS=-L${JAVA_HOME}/lib ENABLE_JAVA=1
```

README.md
```diff
@@ -147,7 +157,16 @@ $ pwd
...
 ```
 
-## Step 6: Run the example experiment
+If you also want to try the example Java benchmarks, you must build them
+as a separate step:
+
+```bash
+$ pwd
+.../krun/examples/benchmarks
+$ make java_bench
```
vext01: Got caught out by this. Typically, non-filename targets in makefiles use hyphens instead of underscores.

vext01:
The examples worked for me. So just these few items to fix.

README.md
vext01: I just noticed that the warmup experiment uses java 8, so in theory 7 or 8 could be used.

snim2:
OK, I *think* this addresses everything. When you are happy I will need to squash.

vext01:
Good stuff. Squash when ready.

snim2:
Splat.

PR #206, opened 2016-05-17
Debug switches

### LINKED ISSUE #181: Number of emails sent so far does not persist between reboots.

vext01:
Hi,

Don't merge this yet. Opening for discussion.

Here's an attempt at sorting out our debug flags. I've managed to untangle `--develop` and `--dryrun` nicely (i hope).

Removed flags:
 * `--dryrun`
 * `--develop`

New flags:
 * Any feature which causes a delay can be skipped with `--quick`.
 * Platform specific flags:
   * `--no-pstate-check`
   * `--no-tickless-check`
 * Other flags:
   * `--no-user-change` -- Means you don't need to set up a krun user and make sure it has permissions to the benchmarks etc.
   * `--fake-reboots` -- Useful for testing stuff which needs persist across reboots, without actually rebooting. E.g. When we come to fix #181 this will be useful.
   * `--dry-run` (not `--dryrun`) -- Fakes the benchmarks, making the finish instantaneously. Useful when hacking a Krun feature which does not touch benchmarks themselves.

Using any of these switches will cause krun to emit warnings that will also appear in the log file. The warning details which switch caused the warning, e.g.:

```
[2016-05-17 14:20:28 WARNING] Ignoring enabled P-states (--no-pstate-check)
```

I know we were aiming to have only `--quick` and platform specific switches. The "other flags" bits are things which I have untangled from the old `--dryrun` and `--develop`. I thought hard about trying to group these, but I now feel strongly that we shouldn't. If we were to group them, I would only end up having to comment out parts of code all the time, which is what we are trying to avoid.

What do others think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ You need to have the following installed:
   * cpufrequtils (Linux only. `cpufrequtils` package in Debian)
   * cffi (`python-cffi` package in Debian)
   * cset (for pinning on Linux only. `cpuset` package in Debian)
+  * virt-what (Linux only. `virt-what` package in Debian)
 
 If you want to benchmark Java, you will also need:
   * A Java SDK 7 (`openjdk-7-jdk` package in Debian)
@@ -270,19 +271,16 @@ how to consume results in Krun format:
 
 It is often useful to test a configuration file, without actually
 running a full benchmark (especially if the benchmark program is
-long).
-Krun supports this with the `--dryrun` command line switch:
+long). The best to test a config is to do something like:
 
 ```bash
-$ ../krun.py --dryrun --debug=INFO example.krun
+$ ../krun.py --dry-run --quick --debug=INFO example.krun
 ```
 
-By passing in `--debug=INFO` you will see a full log of krun actions
-printed to STDOUT.
-Valid debug levels are: `DEBUG`, `INFO`, `WARN`, `DEBUG`,
-`CRITICAL`, `ERROR`.
+See the "Development and Debug Switches" section for a description of these
+switches.
 
-The `--info` switch reports various statistics about the setup described in the
+Another switch, `--info`, reports various statistics about the setup described in the
 specified config file, such as the total number of process executions and which
 benchmark keys will be skipped etc.
 
@@ -320,7 +318,7 @@ This file is compatible with some Linux machines.
 ### Testing a benchmark run with `--reboot`
 
 If you need to test a benchmark configuration with `--reboot`, you can
-still use the `--dryrun` flag.
+still use the `--dry-run` flag.
 In a dry run, Krun will not reboot your machine (it will simulate
 rebooting by restarting Krun automatically) and will not pause to
 wait for your network interface to come up.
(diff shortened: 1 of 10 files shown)


vext01:
Travis is barfing:
```
IOError: [Errno 2] No such file or directory: '/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor'
```

Will investigate.

snim2:
Do you have a unit test for:

```python
if self.config.ENABLE_PINNING:
 +            self._check_cset_installed()
```

?

vext01:
No, but I can add one.

snim2:
Generally this looks good to me. I definitely need something to turn off tickless checks and anything else that needs a custom kernel.

You could add `--dry-run` to the travis-ci run, as we want travis to test krun not the benchmarks themselves.

vext01:
Do you have any thoughts on the "other flags"?

snim2:
I guess I'm not sure how useful `--no-user-change` would be. I can't think of a reason why creating a new user would be tricky; maybe if you don't have root access?

vext01:
This is a somewhat long story.

The krun user needs access to the benchmark code. My home dir is (by principal) mode `700`, meaning that no other user (including krun) can see in my home.

Up until now I have been logging in as krun and developing as the krun user itself. Since the benchmarks are then owned by krun, when krun switches to itself, there are no problems with permissions. In itself this is awkward (duplicated shell config, vim config etc.) , and I am unable to use gvim as the krun user, since the user `edd` owns the X server.

`--no-user-switch` fixes this.

vext01:
This is annoying. Travis does not expose information about the scaling governor. We have two options:

 * Emit a warning and continue if we can't find the governor info.
 * Add a flag to skip the governor check.

Any thoughts?

snim2:
For some odd reason I thought that error had to do with pinning. Anyhow, personally I would add a new switch to turn the check off. On my Ubuntu machine I can't replicate what you are seeing on Travis, FWIW.

vext01:
I think it has to do with ACPI capabilities.

snim2:
Ah yes, that makes sense

ltratt:
An alternative -- which is much easier than a flag, and also doesn't burden the user with options they don't care about -- is just to say `if <env> != <travis>: do_cpu_stuff()`. [Assuming there is a way of determining if one is building under Travis / qemu / whatever.]

vext01:
We could work around travis in this fashion by setting an environment variable, but I'm not sure how to detect we are under qemu. I have to admit I prefer adding a flag.

ltratt:
Have a gander at http://unix.stackexchange.com/questions/89714/easy-way-to-determine-virtualization-technology

I'd prefer the simple option, which also scales much better (imagine if every C `#ifdef` was a command-line switch!). Switches take us a lot of effort to implement and test, and then have the annoyance that they're user visible.

vext01:
I wouldn't have done it this way myself, but if you wont accept a flag, then OK.

Putting this on hold for now, as I'm out of time this week (on hols tomorrow and need to start warmup runs)

ltratt:
As a general rule, we should not be exposing knobs to users for things we can do automatically.

vext01:
Alright, that's virtualisation detection working. Pretty annoying to get right.

 * When I tried `dmidecode -s system-product-name` in a qemu, I didn't get a string which indicated virtualisation, so I couldn't use this.
 * `systemd-detect-virt` worked on Debian, but failed on Travis, since they don't use a Linux system that uses systemd. Arguably, Krun should not depend upon a specific init system anyway.

So I've settled on using `virt-what` (http://people.redhat.com/~rjones/virt-what/) , which works nicely and looks to be fairly complete in terms of what virtualisation it can detect. It's also widely packaged across linux distros.

Now I have a query for you guys/gals. We currently have a config option `ENABLE_PINNING`, which arguably should always be on for reliable production results. It seems odd for this to be a config file option, no? Should this be a "platform-specific flag"?

vext01:
P.S. Much squashing to do here.

ltratt:
My gut feeling is that pinning (and the like) should default to on, and throw an error if Krun can't make them work automatically. But there could then be a command-line to switch that need off. But if we don't need it for now, let's not bother implementing it.

README.md
```diff
@@ -372,21 +370,35 @@ The following platforms are currently supported:
...
+
+  * `--dry-run`: Fakes actual benchmark processes, making them finish
+    instantaneously.
+
+  * `--no-tickless-check`: Do not crash out if the Linux kernel is not
+    tickless.
+
+  * `--no-pstate-check`: Do not crash out if Intel P-states are not disabled.
+
+  * `--fake-reboots`: Restart Krun in-place instead of rebooting.
```
ltratt: What does "restart in-place" mean?

ltratt:
Modulo that one comment, looks good.

vext01:
How about that?

ltratt:
Fine with me. Squash away!

vext01:
Splat

vext01:
Thanks!

PR #209, opened 2016-05-27
Instrument PyPy

vext01:
This is not 100% ready yet, as you will see from some XXX. I'm also going to test the life out of this, as the code isn't overly straightforward.

I'm raising this to get feedback on the approach (described in comments in `PyPyVMDef` changes). Does this look sane @cfbolz?

### DIFF
--- a/krun/results.py
+++ b/krun/results.py
@@ -20,6 +20,11 @@ class Results(object):
         self.data = dict()
         self.reboots = 0
 
+        # Instrumentation counters
+        # "bmark:vm:variant" ->
+        #     (instrumentation name -> [[e0i0, e0i1, ...], [e1i0, e1i1, ...], ...])
+        self.instrumentation_data = {}
+
         # Record how long execs are taking so we can give the user a rough ETA.
         # Maps "bmark:vm:variant" -> [t_0, t_1, ...]
         self.eta_estimates = dict()
@@ -61,6 +66,7 @@ class Results(object):
                 for variant in vm_info["variants"]:
                     key = ":".join((bmark, vm_name, variant))
                     self.data[key] = []
+                    self.instrumentation_data[key] = {}
                     self.eta_estimates[key] = []
 
     def read_from_file(self, results_file):
@@ -82,6 +88,7 @@ class Results(object):
         to_write = {
             "config": self.config.text,
             "data": self.data,
+            "instrumentation_data": self.instrumentation_data,
             "audit": self.audit.audit,
             "reboots": self.reboots,
             "starting_temperatures": self.starting_temperatures,
@@ -91,6 +98,15 @@ class Results(object):
         with bz2.BZ2File(self.filename, "w") as f:
             f.write(json.dumps(to_write,
                                indent=1, sort_keys=True, encoding='utf-8'))
+
+    def add_instrumentation_data(self, bench_key, inst_dct):
+        """Record instrumentation data into results object."""
+
+        for inst_key, v in inst_dct.iteritems():
+            # keys in dict are arbitrary, so we may need to scaffold a new list
+            if inst_key not in self.instrumentation_data[bench_key]:
+                self.instrumentation_data[bench_key][inst_key] = []
+            self.instrumentation_data[bench_key][inst_key].append(v)
 
     def jobs_completed(self, key):
         """Return number of executions for which we have data for a given
--- a/krun/results.py
+++ b/krun/results.py
@@ -23,7 +23,7 @@ class Results(object):
         # Instrumentation counters
         # "bmark:vm:variant" ->
         #     (instrumentation name -> [[e0i0, e0i1, ...], [e1i0, e1i1, ...], ...])
-        self.instrumentation_data = {}
+        self.inst_data = {}
 
         # Record how long execs are taking so we can give the user a rough ETA.
         # Maps "bmark:vm:variant" -> [t_0, t_1, ...]
@@ -66,7 +66,7 @@ class Results(object):
                 for variant in vm_info["variants"]:
                     key = ":".join((bmark, vm_name, variant))
                     self.data[key] = []
-                    self.instrumentation_data[key] = {}
+                    self.inst_data[key] = {}
                     self.eta_estimates[key] = []
 
     def read_from_file(self, results_file):
@@ -88,7 +88,7 @@ class Results(object):
         to_write = {
             "config": self.config.text,
             "data": self.data,
-            "instrumentation_data": self.instrumentation_data,
+            "inst_data": self.inst_data,
             "audit": self.audit.audit,
             "reboots": self.reboots,
             "starting_temperatures": self.starting_temperatures,
@@ -104,9 +104,9 @@ class Results(object):
 
         for inst_key, v in inst_dct.iteritems():
             # keys in dict are arbitrary, so we may need to scaffold a new list
-            if inst_key not in self.instrumentation_data[bench_key]:
-                self.instrumentation_data[bench_key][inst_key] = []
-            self.instrumentation_data[bench_key][inst_key].append(v)
+            if inst_key not in self.inst_data[bench_key]:
+                self.inst_data[bench_key][inst_key] = []
+            self.inst_data[bench_key][inst_key].append(v)
 
     def jobs_completed(self, key):
         """Return number of executions for which we have data for a given
--- a/krun/results.py
+++ b/krun/results.py
@@ -5,6 +5,7 @@ from krun.util import fatal
 
 import bz2  # decent enough compression with Python 2.7 compatibility.
 import json
+from collections import defaultdict
 
 
 class Results(object):
@@ -66,7 +67,7 @@ class Results(object):
                 for variant in vm_info["variants"]:
                     key = ":".join((bmark, vm_name, variant))
                     self.data[key] = []
-                    self.inst_data[key] = {}
+                    self.inst_data[key] = defaultdict(list)
                     self.eta_estimates[key] = []
 
     def read_from_file(self, results_file):
@@ -103,8 +104,6 @@ class Results(object):
         """Record instrumentation data into results object."""
 
         for inst_key, v in inst_dct.iteritems():
-            if inst_key not in self.inst_data[bench_key]:
-                self.inst_data[bench_key][inst_key] = []
             self.inst_data[bench_key][inst_key].append(v)
 
     def jobs_completed(self, key):
--- a/krun/results.py
+++ b/krun/results.py
@@ -103,9 +103,6 @@ class Results(object):
         """Record instrumentation data into results object."""
 
         for inst_key, v in inst_dct.iteritems():
-            # Keys in dict are not known ahead of time (VMDefs are free to put
-            # whatever keys they like here). We therefore scaffold the lists on
-            # an ad-hoc basis when we see a key for the first time.
             if inst_key not in self.inst_data[bench_key]:
                 self.inst_data[bench_key][inst_key] = []
             self.inst_data[bench_key][inst_key].append(v)
--- a/krun/results.py
+++ b/krun/results.py
@@ -103,7 +103,9 @@ class Results(object):
         """Record instrumentation data into results object."""
 
         for inst_key, v in inst_dct.iteritems():
-            # keys in dict are arbitrary, so we may need to scaffold a new list
+            # Keys in dict are not known ahead of time (VMDefs are free to put
+            # whatever keys they like here). We therefore scaffold the lists on
+            # an ad-hoc basis when we see a key for the first time.
             if inst_key not in self.inst_data[bench_key]:
                 self.inst_data[bench_key][inst_key] = []
             self.inst_data[bench_key][inst_key].append(v)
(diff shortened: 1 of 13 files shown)


krun/results.py
```diff
@@ -20,6 +20,11 @@ def __init__(self, config, platform, results_file=None):
         self.data = dict()
         self.reboots = 0
 
+        # Instrumentation counters
+        # "bmark:vm:variant" ->
+        #     (instrumentation name -> [[e0i0, e0i1, ...], [e1i0, e1i1, ...], ...])
+        self.instrumentation_data = {}
```
ltratt: Can we find a shorter name for this attribute?

ltratt:
I'm not totally sure I've understood the logic here, but are we reading from stdin? If so, could we end up in a situation where if we don't read fast enough, PyPy's writes to stdout could block, and thus we might appear to slow it down? If so, it might be safer to have it write to a separate file and then read it in and process it after execution has finished.

vext01:
Alright, here's something that we can now review. I need to test the changes to the iterations runners on a linux system will a full set of VMs built (all machines with VMs built in use, so I will have to build VMs on bencher2).

@ltratt This incarnation uses the Popen constructor to redirect stderr direct to file, bypassing krun altogether. I've checked the CPython code to check that it really is a file descriptor-level redirect. So we are grand.

@cfbolz There's a large comment (which I will point out in a moment) which describes exactly what I am collecting and how. Does it look good?

Comments, thoughts, musings, and so forth?

krun/vm_defs.py
```diff
@@ -480,12 +579,105 @@ def __init__(self, vm_path, env=None):
         lib_dir = os.path.dirname(vm_path)
         self.add_env_change(EnvChangeAppend("LD_LIBRARY_PATH", lib_dir))
 
+    def parse_instrumentation_stderr_file(self, file_handle):
+        """PyPy instrumentation data collected from the PYPYLOG.
```
vext01: @cfbolz here

krun/vm_defs.py
```diff
@@ -33,6 +33,9 @@
 if DASH is None:
     fatal("dash shell not found")
 
+INSTRUMENTATION_END_PROC_ITER_PREFIX = "@@@ END_IN_PROC_ITER:"
```
ltratt: For consistency with elsewhere, I think this should be `INSTR_END...`.

krun/vm_defs.py
```diff
@@ -480,12 +579,105 @@ def __init__(self, vm_path, env=None):
...
 
+    def parse_instrumentation_stderr_file(self, file_handle):
+        """PyPy instrumentation data collected from the PYPYLOG.
+
+        We collect:
+          * The time spent in compilation-related tasks.
+          * The time spent in GC-related tasks.
+
+        Note that the times are not in wall-clock time. Consider the units
+        arbitrary.
```
ltratt: Obvious question: if the units are arbitrary, what can we actually do with them? If (as I suspect) we're just using them to determine parent/child events, let's make that clear.

krun/vm_defs.py
```diff
@@ -480,12 +579,105 @@ def __init__(self, vm_path, env=None):
...
+
+        # This stores the counters for the current in-process iteration
+        data_this_iter = {
+            "gc_time": 0,
+            "compilation_time": 0
+        }
+
+        iter_num = 0
+        current_event = None
+
```
ltratt: Stray blank line.

krun/vm_defs.py
```diff
@@ -480,12 +579,105 @@ def __init__(self, vm_path, env=None):
...
+                    counter_name = None
+
+                if counter_name is not None:
+                    data_this_iter[counter_name] += current_event.get_duration()
+
+                current_event = current_event.parent
+                continue
+
+        # All events should be done and dusted
+        assert current_event is None
```
ltratt: I might be missing out on something obvious, but if, in the `for` loop `current_event` is set to something that isn't `None`, I don't see that it can ever be set back to `None`. At which point this `assert` will fail?

krun/results.py
```diff
@@ -92,6 +99,15 @@ def write_to_file(self):
             f.write(json.dumps(to_write,
                                indent=1, sort_keys=True, encoding='utf-8'))
 
+    def add_instrumentation_data(self, bench_key, inst_dct):
+        """Record instrumentation data into results object."""
+
+        for inst_key, v in inst_dct.iteritems():
+            # keys in dict are arbitrary, so we may need to scaffold a new list
```
ltratt: I don't understand this comment (is it suggesting the keys are in an arbitrary order? or ...?).

ltratt:
OK, the large amount of scaffolding means that I'm not entirely sure I can see the wood for the trees. I've tried to offer some comments anyway, but this is a hard PR to review!

krun/vm_defs.py
vext01: In the latest batch of changes, I've been shortening to "inst". Are you ok with `INST_END_PROC_ITER_PREFIX` for this one?

ltratt: Since that's what I was suggesting too, yes ;)

krun/vm_defs.py
vext01: The times can be used relatively even if we do not know the units. E.g.

```
In iteration X the VM spent twice as long compiling than in iteration Y. This is reflected by a larger spike in the iteration time for X.
```

Regardless, I'm going to read the PyPy code to try to find out what the time is in terms of. At that point, we can revise this comment.

ltratt: Yes, that's my point: if the times are only useful for determining the relative order of things, let's be clear about that. [At the moment, it's implicit in the comment, but very few readers would realise the actual implications IMHO.]

krun/vm_defs.py
vext01: Do you mean butt the variable up against the loop header? I would prefer to keep this white-space for readability.

krun/vm_defs.py
vext01: When we enter an event, a reference to the parent event is stored inside the new event:

```
new_event = PyPyVMEvent(event_type, start_time, current_event)
```

At nesting level 0, current_event is None, so that is what is stored.

When this event is complete, we restore the old current_event value:

```
current_event = current_event.parent
```

The assertion has the effect of ensuring that we are at nesting level 0 when we finish.

krun/vm_defs.py
vext01: You suggested `INSTR`, I suggested `INST`.

krun/vm_defs.py
ltratt: Ah, OK, that makes sense. A (brief) comment pointing this out might be helpful, for the next dummy like me who's confused.

krun/vm_defs.py
ltratt: OK :)

krun/vm_defs.py
vext01: Bear with me while I investigate the unit.

ltratt: I don't think the unit is important (my comment is really about the comment itself, not about the units involved).

krun/results.py
vext01: The keys that appear in the instrumentation data are not known ahead of time. This means different VMs can provide different kinds of data if we so wish. Upshot: we have to instantiate the keys in the results file in an ad-hoc "when we see them" fashion. I will update the comment.

krun/vm_defs.py
cfbolz: FWIW, the time is in terms of CPU high performance counter. It might be clock cycles? Anyway, I doubt that it's possible to turn it into something useful non-relative.

ltratt: That's fine: the units aren't important. But we need to make the reader aware of why the fact that we don't know what the units are isn't important.

krun/vm_defs.py
```diff
@@ -480,12 +579,105 @@ def __init__(self, vm_path, env=None):
...
+
+        We treat the event stream like a tree. We walk this tree, keeping track
+        of the nesting of events. When computing the time for each event, we
+        exclude the time spent in nested events. The nested children have their
+        times computed separately.
+
+        The INSTRUMENTATION mapping decides which kind of events contribute
+        to which time counter (gc or compilation).
+        """
+
```
cfbolz: Just wanted to say that the above description of PyPy's nesting and also the algorithm to un-nest things look good to me.

vext01:
I think this covers all the comments. Please don't merge until I have had a chance to test on bencher2.

krun/vm_defs.py
vext01: Thanks Carl.

vext01:
Good idea!

vext01:
Fixed.

BTW, here's a sample of what we get out at the end (in the results file):

```
...
 "inst_data": {
  "nbody:PyPy:default-python": {
   "compilation_time": [
    [
     112419550, 
     0, 
     0, 
     0, 
     0
    ], 
    [
     113684652, 
     0, 
     0, 
     0, 
     0
    ]
   ], 
   "gc_time": [
    [
     71805477, 
     141415, 
     57904, 
     73053, 
     55717
    ], 
    [
     73175387, 
     152490, 
     62061, 
     74339, 
     57901
    ]
   ]
  }
...
```

ltratt:
I'm not quite sure I understand the output. Are those the start times of events or their duration or ...?

vext01:
It's the amount of time spent per in-process iteration in the corresponding event. E.g. we see that for those 5 in-process iterations, we see that compilation occurred only in the first.

vext01:
@cfbolz can do.

ltratt:
Can we tell if compilation/GC happened more than once in an iteration?

cfbolz:
For GC yes, for tracing that's a lot less well-defined.

vext01:
We can't from this data alone, but we are free to add whatever we wish on a per-vm basis (hence the instrumentatin keys being ad-hoc).

That sais, in some sense I don't really care how many GCs occured, as (e.g.) 2 GCs could mean 2 * 0.0000001s or 2 * 40s. This is somewhat hindered by the timing units of course, but at-least we have a relative bearing.

ltratt:
I would definitely like to know if more than one compilation/GC event happened in an iteration.

ltratt:
I guess it just means that what's currently an integer needs to be a list of integers?

vext01:
@ltratt have a quick look at a PyPy log and look at the myriad of different gc and tracing events. I'm not saying you are wrong, but it may change your opinion.

If you are still sure you want counts, I can add them.

ltratt:
I agree that tracing tends to have lots of little compilation events going on, but that may not be true of other VMs. In general, it's safest if we record all the events separately. We can then merge them when drawing graphs if we want, but if we merge the information too soon, we can't recover it.

vext01:
> I guess it just means that what's currently an integer needs to be a list of integers?

I'm not sure this would play well with the nesting of events. Remember, some events include others...

ltratt:
> I'm not sure this would play well with the nesting of events

Unless I'm missing out on something, it can't be any worse than the current situation. [I'm also not entirely sure that I care too much about the nesting of events, though having that information lurking around doesn't hurt.]

vext01:
I'm not interested in the nesting at-all. But we need to think how to flatten it out. Let me stare at a PYPYLOG and see what I can conclude.

ltratt:
Indeed, if we record the start and end time of events (not just their duration), we can always recover nesting.

Now that I think about it, maybe Krun isn't the best place to be doing this processing. Maybe we just want to record events linearly, something like:

```
  ("gc", <start time>, <end time>)
  ("compilation", <start time>, <end time>)
```

and so on. Recorded that way, it's easy to work out what nesting occurred when we come to graph things (etc.).

vext01:
@cfbolz is a `gc-step` guaranteed to include no other gc events? Is a `gc-minor` guaranteed to include only `gc-minor-walkroots` events?

vext01:
@ltratt so I think you mean, just dump the tree into the results file for post-processing?

ltratt:
More-or-less, yes. A good rule of thumb is that we don't to squash away information too soon.

ltratt:
Also, keeping Krun itself simple is probably a good thing! That way, it's less likely that anything it can do will interfere with timings.

vext01:
I'm OK with that.

Note that the Java instrumentation does not create such a tree. There we will get just times, compilation counts, and (IIRC) a blurry notion of major/minor GC counts (because memory pools overlap).

ltratt:
That's fine -- different VMs will record very different information.

Maybe we should explicitly record which GCs are major and minor (as far as possible) so that we can pull them apart more easily later?

vext01:
@ltratt AFAICS, that issue (and the others we have discussed) now become post-processing issues. Unless @cfbolz has any input, let's get the raw data in the results file for now.

ltratt:
OK, sounds like a plan.

vext01:
Will wait a while in-case @cfbolz has any input.

cfbolz:
Storing raw data makes sense to me. However, we'll have to do some of the work of teasing things apart in post processing then.

vext01:
It just moves from Krun to post :p

I'm still not clear what the post-processing will actually do, or if exactly what we want is possible, but we can cross (or burn) that bridge when we get to it. For now I am going to make a new branch and keep this branch so we can get the parser back for post-processing later.

ltratt:
> I'm still not clear what the post-processing will actually do

It gives us flexibility. At the moment, we're inadvertently destroying some data which we may later want to have available to us.

vext01:
Agreed.

PR #210, opened 2016-06-07
Do not pin by default.

vext01:
As discussed face to face with @ltratt:

We suspect that the VMs are asking the system how many CPUs the system
has (not the shield) and spawning a number of worker threads
accordingly. By limiting the number of cores, we may be making
artificial thread contention.

Another PR will follow shortly, as I notice krun thinks bencher3 is virtualised(!)


vext01:
Need to fix tests. Moment.

ltratt:
Looks good. Squash and then I'll merge?

vext01:
splat

PR #211, opened 2016-06-07
Fix virtualisation warning. Missing ().

vext01:
Well, that explains why Krun thinks bencher3 is virtualised!

OK?


ltratt:
A classic bug -- we've all done it!

PR #212, opened 2016-06-09
Allow some dmesg changes, based upon regex patterns.

vext01:
In response to PyPy making noise in OpenBSD-current. E.g.
pypy(87239): mmap W^X violation

Giving this to @ltratt as he proposed such a feature a while ago.

I guess we expect that a JIT might use W+X pages, but ideally it would re-map them as just one as required. Will ask Armin.

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -7,6 +7,7 @@ import random
 import sys
 import glob
 import subprocess
+import re
 from distutils.spawn import find_executable
 from collections import OrderedDict
 from krun import ABS_TIME_FORMAT
@@ -53,7 +54,6 @@ class BasePlatform(object):
         self.find_temperature_sensors()
 
         self.last_dmesg = None
-        self.last_dmesg_time = None
 
     def sleep(self, secs):
         if self.quick_mode:
@@ -67,7 +67,6 @@ class BasePlatform(object):
         # errors and warnings in the Linux dmesg. If that happens, we
         # really want to know about it!
         self.last_dmesg = self._collect_dmesg_lines()
-        self.last_dmesg_time = localtime()
 
     @property
     def starting_temperatures(self):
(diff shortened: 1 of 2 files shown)


krun/platform.py
```diff
@@ -132,26 +133,64 @@ def _collect_dmesg_lines(self):
...
+                if not self.filter_new_dmesg_line(line, patterns):
+                    # this is a new line which we did not expect
+                    problem_lines.append(rec)
+            else:
+                assert False  # The other codes should not arise
+
+        rv = False
+        if problem_lines:
+            # dmesg changed! Emit something which looks a bit like a diff.
+
```
vext01: extra line, sorry

krun/platform.py
```diff
@@ -132,26 +133,63 @@ def _collect_dmesg_lines(self):
...
+        patterns = self.get_allowed_dmesg_patterns()
 
-        old_fn = self._timestamp_to_str(self.last_dmesg_time)
-        new_fn = self._timestamp_to_str(new_dmesg_time)
-        lines = [x for x in difflib.unified_diff(
-            self.last_dmesg, new_dmesg, old_fn, new_fn, lineterm="")]
+        rv = self._check_dmesg_for_changes(patterns,
+                                           self.last_dmesg,
+                                           self.last_dmesg_time, new_dmesg,
+                                           new_dmesg_time)
```
ltratt: I'm not sure I understand why some lines have 1 param and others 2.

krun/platform.py
```diff
@@ -132,26 +133,63 @@ def _collect_dmesg_lines(self):
...
+        for rec in delta:
+            code = rec[0:2]
+            line = rec[2:]
+
+            if code == "  ":  # in both
+                pass
+            elif code == "+ ":  #  new line in dmesg
+                if not self.filter_new_dmesg_line(line, patterns):
+                    # this is a new line which we did not expect
+                    problem_lines.append(rec)
```
ltratt: This will append the line with a `+ ` prefix, I think?

krun/platform.py
```diff
@@ -132,26 +133,63 @@ def _collect_dmesg_lines(self):
...
+            else:
+                assert False  # The other codes should not arise
+
+        rv = False
+        if problem_lines:
+            # dmesg changed! Emit something which looks a bit like a diff.
+            old_fn = "old_dmesg   %s" % self._timestamp_to_str(last_dmesg_time)
+            new_fn = "new_dmesg   %s" % self._timestamp_to_str(new_dmesg_time)
+
+            warn_s = ("dmesg seems to have (unexpectedly) changed.\n"
```
ltratt: It doesn't seem to have changed -- it has changed, right?

krun/platform.py
```diff
@@ -513,6 +551,12 @@ def is_virtual(self):
 
         return False
 
+    def get_allowed_dmesg_patterns(self):
+        # PyPy uses write+executable pages, which causes this dmesg entry
+        # in OpenBSD-current (as of around 6.0-beta)
+        return UnixLikePlatform.get_allowed_dmesg_patterns(self) + \
+            [re.compile("pypy\([0-9]+\): mmap W\^X violation$")]
```
ltratt: Should we be hard-coding this into Krun, or allowing the user to specify it in a config file?

krun/platform.py
vext01: So as to not have a long line. The standard way is to line up the next line with the open parenthesis.

krun/platform.py
vext01: Yep, that's intentional.

krun/platform.py
ltratt: Yes, but it currently makes no sense as e.g. the first line has enough room for 2 params.

krun/platform.py
ltratt: Why would the user want to do a regex on a diff line? Isn't that an internal detail that should be hidden from them?

krun/platform.py
vext01: I'm not sure what you mean. If there are problem lines, then the dmesg has changed and at-least one of the lines that changed was to matched by the filter.

krun/platform.py
vext01: Good question. But for now let's hard-code it and write a bug report.

krun/platform.py
ltratt: Or, even better, `dmesg changed` (as a user, I don't enjoy reading excess verbiage, as I tend to see the same errors repeatedly -- less is more :)).

krun/platform.py
vext01: Ah, my bad, fixed.

krun/platform.py
vext01: Read again. The pattern is matched against the line, but we report a line beginning "+" so as to give a diff-like error message.

krun/platform.py
vext01: The intuition behind this is that the dmesg can change in an expected manner, which we don't bother to report. If we get to this code the dmesg changed, but in an unexpected manner.

krun/platform.py
vext01: The more I think of it, the more I think it should be configurable. But I'm keen to push on with instrumentation, so let's address that later.

vext01:
I've raised the W^X issues on the pypy-dev list here:
https://mail.python.org/pipermail/pypy-dev/2016-June/014489.html

vext01:
oops, wrong button

krun/platform.py
ltratt: Yes, but the user will know that, if reported, they need to look at things. No need to go overboard in the explanation IMHO.

vext01: fixed

krun/platform.py
vext01: https://github.com/softdevteam/krun/issues/213

vext01:
Any further comments, or shall I squash?

krun/platform.py
ltratt: I mostly mean will the regex be applied to the `+ ` or not?

[I'm also unsure that the `+ ` is useful, as dmesgs are, by definition, concatenative thingies, so it's obvious that it's a new line and that Krun will have suppressed the line if it's not of interest.]

ltratt:
Just the one new(ish) comment.

krun/platform.py
vext01: The regex is not applied to the `+`.

`rec` is a string starting with a two character code, followed by the line itself. We match on the `line` which has the code stripped.

I take your point about dmesg being concatenative (I'm currently assuming that lines dont fall off the top of the dmesg buffer -- cross that bridge if we encounter it). Diff just seemed like a natural output format, but we could just as easily make the error message say:

```
New lines in dmesg:
line1
line2
...
```

ltratt: Things will fall off the top, but I think all we need to do to cope with that is ignore lines starting with `- `.

I think the "new lines" idea works well. [I'd probably indent the lines found from dmesg by a couple of chars, just to make it clear when they start and end, but that's a minor thing.]

vext01: All sounds good to me. Implementing...

vext01:
Try this.

ltratt:
Modulo some possible simplification of the code, I think this is heading in the right direction.

vext01:
Here's the equivalent code using two loops. I'm not happy with it.

The generator tends to get in the way. I'm starting to wonder if I should just suck the generator into a list and use more memory.

Diff coming...

ltratt:
Yes, reading it in would surely be much easier.

vext01:
Here we go then.

ltratt:
Looks good! Squash and then I'll merge?

vext01:
Fixed a potential bug (which shouldn't happen in practice anyway). Also removed a bogus comment. Still OK?

ltratt:
Good catch. Agreed.

vext01:
Got there in the end.

Splat.

ltratt:
Thanks!

PR #214, opened 2016-06-13
Fix a temperature test.

vext01:
Found this whilst doing the dmesg filter. My laptop gets hotter than 100 when translating PyPy:

```
>       assert all([10 <= v <= 100 for v in temps.itervalues()])
E       assert all([False, True])
```

Use an upper bound of 120 degrees instead.

OK?


vext01:
It's always been a hot laptop :P The rubber melted off the CPU after a month or two of purchase!

ltratt:
I'm fine with this in terms of the test, so will merge (but worried about the laptop!).

vext01:
I'll do some research and see if this is a common complaint.

PR #215, opened 2016-06-14
Instrument pypy raw

vext01:
Write out raw PyPy events into results file, e.g.:

Large comment in the code explains the format.

```
 "inst_data": {
  "nbody:PyPy:default-python": {
   "raw_vm_events": [
    [
     [
      "root", 
      null, 
      null, 
      [
       [
        "gc-set-nursery-size", 
        32583494957343, 
        32583495128488, 
        []
       ], 
       [
        "gc-minor", 
        32583495215578, 
        32583495272469, 
        [
         [
          "gc-minor-walkroots", 
          32583495222035, 
          32583495227446, 
          []
         ]
        ]
       ], 
       [
        "gc-set-nursery-size", 
        32583495284669, 
        32583495294252, 
        []
       ], 
```

By all means review, but don't merge yet please. I want to add some tests and try it on a reduced version of the warmup experiment.

### DIFF
--- a/iterations_runners/iterations_runner.py
+++ b/iterations_runners/iterations_runner.py
@@ -17,13 +17,15 @@ clock_gettime_monotonic = libkruntime.clock_gettime_monotonic
 
 # main
 if __name__ == "__main__":
-    if len(sys.argv) != 5:
+    if len(sys.argv) != 6:
         print("usage: iterations_runner.py "
-        "<benchmark> <# of iterations> <benchmark param> <debug flag>\n")
+              "<benchmark> <# of iterations> <benchmark param> <debug flag> "
+              "<instrument flag>\n")
         sys.exit(1)
 
-    benchmark, iters, param, debug = sys.argv[1:]
-    iters, param, debug = int(iters), int(param), int(debug)
+    benchmark, iters, param, debug, instrument = sys.argv[1:]
+    iters, param, debug, instrument = \
+        int(iters), int(param), int(debug) == 1, int(instrument) == 1
 
     assert benchmark.endswith(".py")
     bench_mod_name = os.path.basename(benchmark[:-3])
@@ -44,6 +46,11 @@ if __name__ == "__main__":
         start_time = clock_gettime_monotonic()
         bench_func(param)
         stop_time = clock_gettime_monotonic()
+
+        # In instrumentation mode, write an iteration separator to stderr.
+        if instrument:
+            sys.stderr.write("@@@ END_IN_PROC_ITER: %d\n" % i)
+            sys.stderr.flush()
 
         iter_times[i] = stop_time - start_time
 
--- a/krun/results.py
+++ b/krun/results.py
@@ -5,6 +5,7 @@ from krun.util import fatal
 
 import bz2  # decent enough compression with Python 2.7 compatibility.
 import json
+from collections import defaultdict
 
 
 class Results(object):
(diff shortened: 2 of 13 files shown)


vext01:
Er, github is confused. Some of the old PR comments are here...

ltratt:
Looks good to me. Let me know when you've tested, and I'll merge.

vext01:
Added some tests. VMs still building on bencher2...

vext01:
I've just brought this up to date with master, so I can test the changes to the iterations runners...

vext01:
The iterations runners changes appear to work OK. There were some unrelated crashes to do with Oracle VM updates which I will handle in separate PRs.

OK for me to squash @ltratt?

iterations_runners/iterations_runner.py
```diff
@@ -45,6 +47,11 @@
         bench_func(param)
         stop_time = clock_gettime_monotonic()
 
+        # In instrumentation mode, write a iteration separator to stderr.
```
ltratt: "an iteration separator"

krun/results.py
```diff
@@ -92,6 +100,12 @@ def write_to_file(self):
             f.write(json.dumps(to_write,
                                indent=1, sort_keys=True, encoding='utf-8'))
 
+    def add_instrumentation_data(self, bench_key, inst_dct):
```
ltratt: This method name is a real mouthful... Can we do `add_instr_data` or similar?

ltratt:
There are large numbers of:

```python
    def parse_instrumentation_stderr_file(self, file_handle):
        pass
```

which suggests to me the superclass should have this method as the default?

ltratt:
OK, a few minor comments on the new commits.

vext01:
The `pass` methods are implementing abstract methods. This is a design decision. I am forcing the implementer to think about what needs to be done to support a new VM.

ltratt:
OK, but if they're all `pass`...?

vext01:
Java and PyPy will have implementations of `parse_instrumentation_stderr_file`.

ltratt:
I dunno, I'm not entirely convinced in this case, because if someone doesn't provide a `parse_instr_stderr` (which is a snappier name too) method, then things still work OK? In other words, it'll be pretty obvious to them if they haven't dealt with things.

Maybe a more fundamental question just raised itself to me: is this parsing happening while the iterations runner is happening? Or is it post-processing? Or some combination?

vext01:
If an implementation isn't provided, things will work. Maybe it isn't a huge deal in this case...

Parsing is all post.

ltratt:
OK, so let's make it a `pass` method in the superclass then? [Can we shorten the name at the same time too? Long method names make code harder to read IMHO, as it causes lots of line spills.]

vext01:
Way ahead of you ;)

ltratt:
I expected no less :)

vext01:
Fixed. Caught a bug and some more items that should be `instr_...`. OK to squash?

vext01:
Not sure why GH put my comment in the middle of the changeset :P

ltratt:
Looks good... squash and I'll merge?

vext01:
done

ltratt:
Something just changed? I think there was a second force push.

vext01:
yeah, fixed the test

vext01:
I should have pushed the new commit, so you could see what changed.

There was a test exercising the bogus assertion that I fixed in the last round. I removed this.

ltratt:
OK, mistakes happen, no probs.

PR #220, opened 2016-06-22
Fix a syntax error in Javascript iterations runner.

vext01:
Missed a syntax error introduced into the Javascript iterations runner by the instrumentation flag work (earlier this week).

Simple fix. OK?


snim2:
Revolting syntax, but otherwise looks fine. Will merge when tests are done.

PR #231, opened 2016-08-05
More robust pypy instr

vext01:
I've been reviewing where we got up to with instrumentation. There was some confusion due to a stale branch, but i'm now clear. The in-tree PyPy instrumentation is in-fact complete. This change:

 *  Makes parser a little more robust -- it checks that the only event left at the end of a process execution is a `jit-summary`.
 * Add's a continue. As it happens, no functional change, but for consistency (see other cases in the parser loop).
 * Fix a comment.

OK?

### DIFF
--- a/krun/tests/test_vmdef.py
+++ b/krun/tests/test_vmdef.py
@@ -11,6 +11,9 @@ from krun.tests.mocks import MockPlatform
 from krun import EntryPoint
 from krun import util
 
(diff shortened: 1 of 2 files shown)


ltratt:
In retrospect, it makes much more sense for @cfbolz to review this one, so I've assigned him.

cfbolz:
Yes, this makes sense to me, it's a good extra safety check. do you want to squash?

vext01:
Nah, it's good to go as-is.

PR #232, opened 2016-08-08
Review jdk instr

vext01:
The JDK instrumentation is all but finished in master, but I noticed a possible issue. It's not clear of GC collector names are unique (there may be two of the same name). So here I am switching from using a mapping (collector names -> info) to a sequence, thus allowing duplicate collector names.

Also improved comments.

OK?

### DIFF
--- a/iterations_runners/iterations_runner.java
+++ b/iterations_runners/iterations_runner.java
@@ -24,76 +24,100 @@ class KrunJDKInstrumentation {
     }
 
     // Emit cumulative compilation time in milliseconds to stderr
-    private String getCompJSON() {
-        return "" + this.compBean.getTotalCompilationTime();
+    private void getCompJSON(StringBuilder sb) {
+        sb.append(this.compBean.getTotalCompilationTime());
     }
 
 
-    // Get GC info for the specified collector
-    private String getGCJSON(GarbageCollectorMXBean gcBean) {
-        String json = "[";
+    /**
+     * Write GC info for the specified collector bean to a StringBuilder.
+     *
+     * @param sb The StringBuilder to write to.
+     * @param gcBean Collector in question
+     *
+     * Note, we resisted the temptation to use a mapping from name to info,
+     * since it's not clear if names are always unique.
+     */
+    private void getGCJSON(StringBuilder sb, GarbageCollectorMXBean gcBean) {
+        sb.append("[\"");
+        sb.append(gcBean.getName());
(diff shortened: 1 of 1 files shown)


iterations_runners/iterations_runner.java
```diff
@@ -29,9 +29,18 @@ private String getCompJSON() {
...
+    /**
+     * Get GC info for the specified collector bean.
+     *
+     * @param gcBean Collector in question
+     * @return string of the form [name, [pools, time, count]]
+     */
     private String getGCJSON(GarbageCollectorMXBean gcBean) {
-        String json = "[";
+        /*
+         * Note, resisted the temptation to use a mapping from name
```
snim2: Could also be a Javadoc comment

snim2:
This looks fine to me, I'll trust that you've tested it!

iterations_runners/iterations_runner.java
ltratt: Since this is inside a function, surely this would be neater with the `//` syntax?

ltratt:
How big does the JSON string get? Won't `+=` have non-linear performance? I think Java has something like StringBuilder which would be close to linear?

vext01:
It doesn't get big, but perhaps we should use a stringbuilder anyway.

ltratt:
The stringbuider thingy looks good to me. Thanks. Will let Sarah complete the PR.

vext01:
@snim2 I have inspected the outcome and there are a couple of simple unit tests that just ensure the JSON is passed through into the result file.

OK to squash?

iterations_runners/iterations_runner.java
```diff
@@ -24,63 +24,69 @@ public KrunJDKInstrumentation() {
...
-    private String getCompJSON() {
-        return "" + this.compBean.getTotalCompilationTime();
+    private void getCompJSON(StringBuilder sb) {
+        sb.append(this.compBean.getTotalCompilationTime());
     }
 
 
     /**
-     * Get GC info for the specified collector bean.
+     * Write GC info for the specified collector bean to a StringBuilderc.
```
vext01: typo

snim2:
Cool, squash away

vext01:
Just wait for travis...

snim2:
Travis is good...

vext01:
It's squashed Jim, but not as we know it.

PR #234, opened 2016-08-11
Tsr support in Krun

vext01:
This accompanies https://github.com/softdevteam/warmup_experiment/pull/142.

Don't merge until discussion there is complete.

### DIFF
--- a/iterations_runners/iterations_runner.c
+++ b/iterations_runners/iterations_runner.c
@@ -15,11 +15,13 @@
 #include <dlfcn.h>
 #include <err.h>
 #include <string.h>
+#include <inttypes.h>
 
 #define BENCH_FUNC_NAME "run_iter"
 
 /* from libkruntime */
 double clock_gettime_monotonic();
+u_int64_t read_ts_reg();
 
 int
 convert_str_to_int(char *s)
@@ -51,8 +53,10 @@ main(int argc, char **argv)
     int       krun_debug = 0;
     void     *krun_dl_handle = 0;
     int     (*krun_bench_func)(int); /* func ptr to benchmark entry */
-    double    start_time = -1, stop_time = -1;
+    double    krun_wallclock_start = -1, krun_wallclock_stop = -1;
     double   *krun_iter_times = NULL;
+    u_int64_t krun_tsr_start = 0, krun_tsr_stop = 0;
+    u_int64_t *krun_tsr_iter_times = NULL;
 
     if (argc != 6) {
         printf("usage: iterations_runner_c "
@@ -85,9 +89,16 @@ main(int argc, char **argv)
         goto clean;
     }
 
+    krun_tsr_iter_times = calloc(krun_total_iters, sizeof(u_int64_t));
+    if (krun_tsr_iter_times == NULL) {
+        errx(EXIT_FAILURE, "%s", strerror(errno));
+        goto clean;
+    }
+
     for (krun_iter_num = 0; krun_iter_num < krun_total_iters;
         krun_iter_num++) {
-        krun_iter_times[krun_iter_num] = -1.0;
+        krun_iter_times[krun_iter_num] = 0;
+        krun_tsr_iter_times[krun_iter_num] = 0;
     }
 
     for (krun_iter_num = 0; krun_iter_num < krun_total_iters;
@@ -99,14 +110,20 @@ main(int argc, char **argv)
         }
 
         /* timed section */
-        start_time = clock_gettime_monotonic();
+        krun_wallclock_start = clock_gettime_monotonic();
+        krun_tsr_start = read_ts_reg();
         (void) (*krun_bench_func)(krun_param);
-        stop_time = clock_gettime_monotonic();
+        krun_tsr_stop = read_ts_reg();
+        krun_wallclock_stop = clock_gettime_monotonic();
 
-        krun_iter_times[krun_iter_num] = stop_time - start_time;
+        krun_iter_times[krun_iter_num] =
+            krun_wallclock_stop - krun_wallclock_start;
+        krun_tsr_iter_times[krun_iter_num] =
+            krun_tsr_stop - krun_tsr_start;
     }
 
-    fprintf(stdout, "[");
+    /* Emit results */
+    fprintf(stdout, "[[");
     for (krun_iter_num = 0; krun_iter_num < krun_total_iters;
         krun_iter_num++) {
         fprintf(stdout, "%f", krun_iter_times[krun_iter_num]);
@@ -115,7 +132,16 @@ main(int argc, char **argv)
             fprintf(stdout, ", ");
         }
     }
-    fprintf(stdout, "]\n");
+    fprintf(stdout, "], \n[");
+    for (krun_iter_num = 0; krun_iter_num < krun_total_iters;
+        krun_iter_num++) {
+        fprintf(stdout, "%" PRIu64, krun_tsr_iter_times[krun_iter_num]);
+
+        if (krun_iter_num < krun_total_iters - 1) {
+            fprintf(stdout, ", ");
+        }
+    }
+    fprintf(stdout, "]]\n");
 
 clean:
     free(krun_iter_times);
--- a/iterations_runners/iterations_runner.java
+++ b/iterations_runners/iterations_runner.java
@@ -132,6 +132,7 @@ class IterationsRunner {
     }
 
     public static native double JNI_clock_gettime_monotonic();
+    public static native long JNI_read_ts_reg();
 
     public static void main(String args[]) throws
         ClassNotFoundException, NoSuchMethodException, InstantiationException, IllegalAccessException, java.lang.reflect.InvocationTargetException {
@@ -167,16 +168,31 @@ class IterationsRunner {
         BaseKrunEntry ke = (BaseKrunEntry) instance; // evil
 
         double[] iter_times = new double[iterations];
-        Arrays.fill(iter_times, -1.0);
+        Arrays.fill(iter_times, 0);
 
-        // Please, no refelction inside the timed code!
+        /*
+         * TSR values are unsigned 64-bit, whereas Java long is *signed*
+         * 64-bit. We are safe since:
+         *
+         *   - The only arithmetic we use is subtract, which is the same
+         *     operation regardless of sign (due to two's compliment).
+         *
+         *   - We print the result using toUnsignedString, thus interpreting
+         *     the number as unsigned.
+         */
+        long[] tsr_iter_times = new long[iterations];
+        Arrays.fill(tsr_iter_times, 0);
+
+        // Please, no reflection inside the timed code!
         for (int i = 0; i < iterations; i++) {
             if (debug) {
                 System.err.println("[iterations_runner.java] iteration: " + (i + 1) + "/" + iterations);
             }
 
             double startTime = IterationsRunner.JNI_clock_gettime_monotonic();
+            long tsrStartTime = IterationsRunner.JNI_read_ts_reg();
             ke.run_iter(param);
+            long tsrStopTime = IterationsRunner.JNI_read_ts_reg();
             double stopTime = IterationsRunner.JNI_clock_gettime_monotonic();
 
             // Instrumentation mode emits a JSON dict onto a marker line.
@@ -189,9 +205,11 @@ class IterationsRunner {
             }
 
             iter_times[i] = stopTime - startTime;
+            tsr_iter_times[i] = tsrStopTime - tsrStartTime;
         }
 
-        System.out.print("[");
+        // wall clock
+        System.out.print("[[");
         for (int i = 0; i < iterations; i++) {
             System.out.print(iter_times[i]);
 
@@ -199,6 +217,15 @@ class IterationsRunner {
                 System.out.print(", ");
             }
         }
-        System.out.print("]\n");
+        System.out.print("], [");
+        // TSR
+        for (int i = 0; i < iterations; i++) {
+            System.out.print(Long.toUnsignedString(tsr_iter_times[i]));
+
+            if (i < iterations - 1) {
+                System.out.print(", ");
+            }
+        }
+        System.out.print("]]");
     }
 }
--- a/iterations_runners/iterations_runner.js
+++ b/iterations_runners/iterations_runner.js
@@ -1,4 +1,5 @@
-// NOTE: you need to provide clock_gettime_monotonic.
+// NOTE: you need to provide clock_gettime_monotonic()
+// NOTE: you need to provide read_ts_reg()
 
 if (this.arguments.length != 5) {
     throw "usage: iterations_runner.js <benchmark> <# of iterations> " +
@@ -13,7 +14,10 @@ var BM_debug = parseInt(this.arguments[3]) > 0;
 load(BM_entry_point);
 
 var BM_iter_times = new Array(BM_n_iters);
-BM_iter_times.fill(-1.0);
+BM_iter_times.fill(0);
+
+var BM_tsr_iter_times = new Array(BM_n_iters);
+BM_tsr_iter_times.fill(0);
 
 for (BM_i = 0; BM_i < BM_n_iters; BM_i++) {
     if (BM_debug) {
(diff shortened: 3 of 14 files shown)


krun.py
```diff
@@ -39,6 +39,7 @@ def usage(parser):
 
 
 def sanity_checks(config, platform):
+    return # XXX
```
vext01: Debug code.

iterations_runners/iterations_runner.java
```diff
@@ -169,14 +172,29 @@ public static void main(String args[]) throws
...
+         *   - The only arithmetic we use is subtract, which is the same
+         *     operation regardless of sign (due to two's compliment).
+         *
+         *   - We print the result using toUnsignedString, thus interpreting
+         *     the number as unsigned.
+         */
+        long[] tsr_iter_times = new long[iterations];
+        Arrays.fill(tsr_iter_times, IterationsRunner.UNINITIALISED_TSR);
+
         // Please, no refelction inside the timed code!
```
snim2: Typo: s/refelction/reflection/g

iterations_runners/iterations_runner.php
```diff
@@ -39,26 +39,39 @@
 /* OK, all is well, let's run. */
```
snim2: This file has mixed spaces and tabs

iterations_runners/iterations_runner.java
vext01: Yeah, I noticed that, but forgot to fix. Will fix.

iterations_runners/iterations_runner.php
vext01: Yeah, I also noted this. Will fix this last to minimise diff lines to review.

vext01:
And travis has beef with the Java build, yay.

Lunch...

libkrun/libkruntime.c
```diff
@@ -40,4 +41,37 @@ JNIEXPORT jdouble JNICALL
 Java_IterationsRunner_JNI_1clock_1gettime_1monotonic(JNIEnv *e, jclass c) {
     return (jdouble) clock_gettime_monotonic();
 }
+
+JNIEXPORT jlong JNICALL
+Java_IterationsRunner_JNI_1read_1ts_1reg(JNIEnv *e, jclass c)
+{
```
snim2: Should this open-curly be on the line above?

libkrun/libkruntime.c
```diff
@@ -40,4 +41,37 @@ JNIEXPORT jdouble JNICALL
...
+}
+
+/*
+ * Return TSR as a double.
+ *
+ * For languages like Lua, where there is suitible integer type
+ */
+double
+read_ts_reg_double()
+{
```
snim2: Open curly on line above?

libkrun/libkruntime.c
vext01: I'm using "almost KNF" code style for C. It's KNF with tabs replaced by four spaces. Function braces are on a newline, all others are not.

snim2: `JNICALL` (lines 40/41) is not a function?

vext01: Not sure what you mean.

For functions, we have:

```
return-type
func_name(args, ...)
{
    ...
}
```

snim2: Perhaps I'm just confused, but it looked to me like we have a couple of functions with a different layout to the rest. e.g. lines 60--66:

```c
u_int64_t
read_ts_reg() {
    u_int32_t hi, lo;

    __asm volatile("rdtsc" : "=d" (hi), "=a" (lo));
    return ((u_int64_t) hi << 32) | (u_int64_t) lo;
}
```

vext01:
This is weird, travis is using Java 7, which shouldn't have a Long.toUnsignedString, yet the error message is saying that the method exists but is being called with the wrong arguments.

http://docs.oracle.com/javase/7/docs/api/java/lang/Long.html

Could it be that an early undocumented version of this method is present in JDK7, which had a different signature to that which hit java 8?

Anyway, we have no intention to run the experiment on Java < 8, so we need to see if we can get java 8 in travis...

snim2:
https://blog.travis-ci.com/2013-11-26-test-your-java-libraries-on-java-8/

vext01:
Hah, there is indeed an undocumented method in Java 7 Long:
```
private static String toUnsignedString(long i, int shift)
```

vs. Java 8:

```
private static String toUnsignedString(long i)
```

snim2:
How annoying!

vext01:
What am I doing wrong with travis here sarah?

snim2:
Not sure, mind if I restart that build?

vext01:
go ahead

iterations_runners/iterations_runner.c
```diff
@@ -15,11 +15,14 @@
 #include <dlfcn.h>
 #include <err.h>
 #include <string.h>
+#include <inttypes.h>
 
 #define BENCH_FUNC_NAME "run_iter"
+#define TSR_DEFAULT 0xffffffffffffffff
```
ltratt: `0` would be more sensible here, surely? [Especially if we sometimes use doubles.]

iterations_runners/iterations_runner.c
```diff
@@ -99,14 +111,20 @@ main(int argc, char **argv)
         }
 
         /* timed section */
-        start_time = clock_gettime_monotonic();
+        krun_tsr_start = read_ts_reg();
```
ltratt: It would be better to do:

```c
read_clock();
read_tsr();
do_bench();
read_tsr();
read_clock();
```

so that the TSR doesn't include the clock. [The clock resolution is waaaaaaay below the TSR resolution, which is why I'm suggesting this way around.]

iterations_runners/iterations_runner.java
```diff
@@ -131,7 +131,10 @@ public void getInstJSON(StringBuilder sb, int iterNum) {
         System.loadLibrary("kruntime");
     }
 
+    static long UNINITIALISED_TSR = -1;  // All ones
```
ltratt: Again, if we're emulating unsigned numbers (as a comment below suggests), `-1` doesn't seem like the best choice.

iterations_runners/iterations_runner.java
```diff
@@ -169,14 +172,29 @@ public static void main(String args[]) throws
...
+        long[] tsr_iter_times = new long[iterations];
+        Arrays.fill(tsr_iter_times, IterationsRunner.UNINITIALISED_TSR);
+
         // Please, no refelction inside the timed code!
         for (int i = 0; i < iterations; i++) {
             if (debug) {
                 System.err.println("[iterations_runner.java] iteration: " + (i + 1) + "/" + iterations);
             }
 
+            long tsrStartTime = IterationsRunner.JNI_read_ts_reg();
```
ltratt: Neds to be clock, tsr, bench, tsr, clock (as before).

iterations_runners/iterations_runner.c
vext01: in the case of doubles we can express negative numbers. I don't see how 0 is better than 0xff.... They are one apart modulo 2^64.  It's kinda moot, as all of the default values are overwritten at the end of the execution.

iterations_runners/iterations_runner.c
vext01: Works for me.

iterations_runners/iterations_runner.js
```diff
@@ -15,24 +16,38 @@ load(BM_entry_point);
...
 
+var BM_tsr_iter_times = new Array(BM_n_iters);
+BM_tsr_iter_times.fill(-1.0);
+
 for (BM_i = 0; BM_i < BM_n_iters; BM_i++) {
     if (BM_debug) {
         print_err("[iterations_runner.js] iteration " + (BM_i + 1) + "/" + BM_n_iters);
     }
 
+    var BM_tsr_start_time = read_ts_reg();
```
ltratt: clock, tsr, bench, tsr, clock

snim2:
I think maybe you need to run pip with `sudo` or run it 'locally'

iterations_runners/iterations_runner.c
ltratt: When you do 0xffffffffffffffffff, it makes people think it means something important (at first, I assumed it was a bit mask). So let's make it something that's "obviously" neutral.

iterations_runners/iterations_runner.lua
```diff
@@ -20,23 +23,38 @@ for BM_i = 1, BM_iters, 1 do
...
+for BM_i = 1, BM_iters, 1 do
+    BM_tsr_iter_times[BM_i] = -1.0
+end
+
 for BM_i = 1, BM_iters, 1 do
     if BM_debug then
         io.stderr:write(string.format("[iterations_runner.lua] iteration %d/%d\n", BM_i, BM_iters))
     end
 
+    local BM_tsr_start_time = kruntime.read_ts_reg_double();
```
ltratt: clock, tsr, bench, tsr, clock

iterations_runners/iterations_runner.java
vext01: Not sure what you mean. long is signed, so -1 when interpreted unsigned is 2^64-1. This default value is always overwritten.

iterations_runners/iterations_runner.java
vext01: Yep, they are all in this order, will fix them all.

vext01:
I didn't change any Python stuff, and it used to work, hmm.

iterations_runners/iterations_runner.java
ltratt: Please see my other comment (in essence, -1 is an odd default, and suggests something clever is going on -- when it isn't).

iterations_runners/iterations_runner.java
ltratt: OK, will avoid commenting on all of them :)

snim2:
You changed the language key in the config though, I think that did it.

libkrun/libkruntime.c
```diff
@@ -40,4 +41,37 @@ JNIEXPORT jdouble JNICALL
...
+ * This assumes and x86 CPU.
+ *
+ * Derived from:
+ * $OpenBSD: pctr.h,v 1.5 2014/03/29 18:09:28 guenther Exp $
+ */
+u_int64_t
+read_ts_reg() {
+    u_int32_t hi, lo;
+
+    __asm volatile("rdtsc" : "=d" (hi), "=a" (lo));
```
ltratt: Think we agreed on rtscp. [Not that it makes *much* difference, but still...]

libkrun/libkruntime.c
vext01: Thats a bug.

iterations_runners/iterations_runner.c
vext01: So, let's do 0 across the board for both arrays.

ltratt: Works for me.

vext01:
Can we have two languages in the config?

libkrun/libkruntime.c
vext01: Yep, going to flip that last. See the original PR description.

snim2:
This is discussed here: https://github.com/travis-ci/travis-ci/issues/4090

It isn't clear to me which bits of the default environment you would get with two language specified. It might be cleaner to choose one then install the necessary packages for others with `apt-get`.

vext01:
OK, let me clean up the other bits and I will try. Not that we did have and apt-get line for java 8 (which I removed when fiddling with jdk section).

vext01:
My filthy hack worked.

snim2:
+1

cfbolz:
another thought about the double: you could cast the uint to a double and back, to check that you didn't lose precision (in the C code). if you did, you could throw an error. Then you at least know if something's wrong.

ltratt:
Since we don't have a choice, a double is fine.

CF: I'm not sure that the two-way coercion will work, because C's coercion rules can sometime do unintuitive things, particularly with signed/unsigned. However, I have long since forgotten what little I knew of the rules, so if you know this does what we expect, I'm fine with it.

cfbolz:
Simply ``value == (unsigned long) (double) value`` compiles without warnings and gave the expected results for a few critical values that I tried.

ltratt:
I *think* that's not surprising since that won't change any of the underlying bits (indeed, I assume the optimiser rewrites that statement statically to simply `true`). Warning: I might be wrong.

cfbolz:
no, you are wrong ;-). this cast turns the value (which is an unsigned long) into the closest double. The bits aren't the same. so turning it back might lose precision. Which is what I tried. The above returns true for small values, and false for very large values, which is what we want.

ltratt:
OK, happy to be corrected :)

vext01:
That's actually very clever. The only small issue I can see is that, in the case of reading the TSR start time, the two casts will add a small amount of time to the stop reading. Thoughts.

ltratt:
We do the checks after we've read the final values I think? Besides, the *tiny* extra maths is trivial compared to the huge number of cycles we normally execute.

vext01:
No, you need the original `u_int64_t` at the C level. Once you return it back as a double, it is lost.

vext01:
But I agree that the work is probably tiny.

ltratt:
What I mean is: we can do the checks on the start value *after* we've finished running the benchmark and read the final TSR I think?

vext01:
Yes, but we would have to stash another array at the C-level. I don't think storing another 2000 doubles is a great idea.

vext01:
Er u_int64_t's sorry

vext01:
Wait, we would only need to store the last start reading. Let me think more.

ltratt:
You don't need to put them in an array. Something like:

```c
st_tsr = tsr();
bench();
end_tsr = tsr();
if (overflowed(st_tsr) || overflowed (end_tsr))
  err();
```

vext01:
You can't do that because `st_tsr` is a float with precision potentially already lost. You need to take the original unsigned64, cast that to a double and back. Of course if you take an integral double, cast it to unsigned64 and back it will always match.

ltratt:
Ah, I think I see why we might be talking at cross purposes. I thought the *C* code would do the check, but you're suggesting the iteration runner does the check?

vext01:
No no, I'm saying the iteration runner cannot do the check. The above code you showed is the iteration runner (in e.g. lua), and tsr() is implemented in C code.

ltratt:
Yes, you're right. Ignore me.

cfbolz:
I think the overhead of the two casts and a comparison is acceptable.

ltratt:
Yes, I think we're fine. We should note it in the paper, but also point out that an extra 10-20 cycles out of a few million isn't too horrible.

vext01:
```
#include <stdio.h>
#include <stdlib.h>
#include <inttypes.h>

u_int64_t
read_ts_reg() {
    u_int32_t hi, lo;

    __asm volatile("rdtsc" : "=d" (hi), "=a" (lo));
    return ((u_int64_t) hi << 32) | (u_int64_t) lo;
}

int
main(int argc, char **argv)
{
    u_int64_t x = read_ts_reg();
    u_int64_t y = (u_int64_t) (double) x;

    printf("%" PRIu64 "\n", y);
    return (EXIT_SUCCESS);
}
```

Makes this code (-O3). I annotated the cast logic.

```
            // Set up stack frame
            0x00000580      4883ec08       sub rsp, 8                                                                    

            // Get TSR
            0x00000584      0f31           rdtsc                                                                         

            // Reconstitute unsigned 64 TSR value from two 32-bit regs
            0x00000586      48c1e220       shl rdx, 0x20                                                                 
            0x0000058a      89c0           mov eax, eax                                                                  
            0x0000058c      4809d0         or rax, rdx                                                                   

         // start cast code
        ,=< 0x0000058f      7849           js 0x5da                    ;[1]                                              
        |   0x00000591      660fefc0       pxor xmm0, xmm0                                                               
        |   0x00000595      f2480f2ac0     cvtsi2sd xmm0, rax                                                            
       .--> 0x0000059a      f20f100da604.  movsd xmm1, qword [rip + 0x1004a6] ; [0x100a48:8]=0x43e0000000000000          
       ||   0x000005a2      660f2ec1       ucomisd xmm0, xmm1                                                            
      ,===< 0x000005a6      731a           jae 0x5c2                   ;[2]                                              
      |||   0x000005a8      f2480f2cf0     cvttsd2si rsi, xmm0                                                           
     .----> 0x000005ad      488d3d8c0410.  lea rdi, [rip + 0x10048c]   ; 0x100a40 ; section..rodata ; "%llu." @ 0x100a40 
           // end cast code

            // printf
     ||||   0x000005b4      31c0           xor eax, eax                                                                  
     ||||   0x000005b6      e8b5fdffff     call sym.imp.printf         ;[3]                                              
```

vext01:
Admittedly, this is more code than I expected, but it's still a mere drop in the ocean compared to the benchmark payload.

ltratt:
Agreed on both points!

vext01:
So I will implement checks in the next round of PRs if that is OK, Keen to merge this large PR (not yet, a couple of tweaks to go in).

ltratt:
OK, I'll nominate myself to do the merge. [Need to check a couple of things first.]

vext01:
So, in a future PR I will:

 * Check double precision using the cast method.
 * Use the asm snippets from the intel manual.

Anything I missed in this PR (or the accompanying warmup_experiment PR) though?

libkrun/libkruntime.c
ltratt: Have we fixed this?

ltratt:
I have one comment which needs a quick check; but otherwise, it looks like we're OK.

libkrun/libkruntime.c
vext01: I did yeah.

ltratt:
OK, so I think we're ready for squashing.

vext01:
I think I'm going to have to splat this all into one. There's too much to pick apart into units.

ltratt:
In this case, I think that's OK.

vext01:
Ready once travis gets its act together ;)

ltratt:
OK!

PR #235, opened 2016-08-15
instrument pypy to measure the jit time directly

cfbolz:
This is probably not quite ready, but would you be up to taking a look at the approach already, @vext01 ?

### DIFF
--- a/iterations_runners/iterations_runner.py
+++ b/iterations_runners/iterations_runner.py
@@ -31,6 +31,9 @@ if __name__ == "__main__":
     iters, param, debug, instrument = \
         int(iters), int(param), int(debug) == 1, int(instrument) == 1
 
+    if instrument:
+        import pypyjit # instrumentation not supported on CPython yet anyway
+
     assert benchmark.endswith(".py")
     bench_mod_name = os.path.basename(benchmark[:-3])
     bench_mod = imp.load_source(bench_mod_name, benchmark)
@@ -44,6 +47,8 @@ if __name__ == "__main__":
     iter_times = [0] * iters
     tsr_iter_times = [0] * iters
     for i in xrange(iters):
+        if instrument:
+            start_snap = pypyjit.get_stats_snapshot()
         if debug:
             sys.stderr.write(
                 "[iterations_runner.py] iteration %d/%d\n" % (i + 1, iters))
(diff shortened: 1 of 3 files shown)


vext01:
Looks good so far :)

cfbolz:
OK, I've just tested this, seems to work fine. @vext01, can you take another look?

ltratt:
Obvious (dumb) question: should we want to instrument CPython (or, indeed, any other non-PyPy Python) in the future, will we just be able to say `if sys.python is "cpython": import someothermod` or similar?

cfbolz:
yes, right now the whole instrument code is specific to pypy. if we instrument CPython (which is hard!) we would have to generalize the code a bit, along the lines you sketched.

ltratt:
OK, so long as it's possible, I'm happy!

cfbolz:
it's possible from the POV of our runner code, yes. However, on CPython it's pretty hard to get at numbers. Here is a thread about that from 2013: https://mail.python.org/pipermail/python-dev/2013-October/129552.html (I think I'll put it into an issue so the link won't get lost).

vext01:
Looks good to me.

Is there anything else in the jit stats that might be useful?

```
>>>> x = pypyjit.get_stats_snapshot()
>>>> x.counter_times
{'TRACING': 0.02404618263244629, 'BACKEND': 0.0015478134155273438}
>>>> x.counters
{'TRACING': 5, 'BACKEND': 5, 'OPS': 1494, 'RECORDED_OPS': 384, 'GUARDS': 152, 'OPT_OPS': 378, 'OPT_GUARDS': 110, 'OPT_GUARDS_SHARED': 67, 'OPT_FORCINGS': 6, 'OPT_VECTORIZE_TRY': 0, 'OPT_VECTORIZED': 0, 'ABORT_TOO_LONG': 0, 'ABORT_BRIDGE': 0, 'ABORT_BAD_LOOP': 0, 'ABORT_ESCAPE': 0, 'ABORT_FORCE_QUASIIMMUT': 0, 'NVIRTUALS': 36, 'NVHOLES': 2, 'NVREUSED': 9, 'TOTAL_COMPILED_LOOPS': 3, 'TOTAL_COMPILED_BRIDGES': 2, 'TOTAL_FREED_LOOPS': 0, 'TOTAL_FREED_BRIDGES': 0}
```

cfbolz:
the remaining jit stats are about the traces (how many operations, how many guards, etc), they don't contain any timings. So I think not.

vext01:
Alright, so are we ready to merge?

cfbolz:
yep

PR #239, opened 2016-09-08
Core cycles support for Krun

vext01:
Hi,

Let's start looking at this, but please don't merge yet.

This change:
 * Adds core_cycles support to libkrun
 * Makes all iterations runners use libkrun.
 * renames variables to be more intuitive.
 * Ensure all iteration runner errors go to stderr.
 * more sanity checking in iterations runners

A typical iteration runner should look like this now (pseudocode):

```
libkruntime_init(); // open msr device nodes and do performance counter setup

loop {
    // timed section
    read_wallclock_time();
    read_core_cycles();
    do_benchmark();
    read_core_cycles();
    read_wallclock_time();
    // end timed section

   check_for_start_readings_greater_than_stop_readings();
}

libkruntime_done(); // close msr device nodes
```

@ltratt is this measurement orderings correct? I think you had some intuition as to which should come first based on clock resolutions etc.

There are some other optimisations and tidy ups, but I would like to do them as separate PRs, as this is already getting large.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ up correctly).
 ## Security Notes
 
 Krun is not intended to be run on a secure multi-user system, as it uses sudo
-to elevate proveleges.
+to elevate privileges.
 
 Sudo is used to:
 
@@ -447,6 +447,8 @@ Sudo is used to:
  * Set process priorities.
  * Create cgroup shields (Linux only, off by default)
  * Detect virtualised hosts.
+ * Change Linux capabilities(7) for MSR device node access.
+ * Change MSR device node filesystem permissions.
 
 Please make sure you understand the implications of this.
 
--- a/iterations_runners/iterations_runner.java
+++ b/iterations_runners/iterations_runner.java
@@ -131,9 +131,10 @@ class IterationsRunner {
         System.loadLibrary("kruntime");
     }
 
+    public static native void JNI_libkruntime_init();
+    public static native void JNI_libkruntime_done();
     public static native double JNI_clock_gettime_monotonic();
-    public static native long JNI_read_ts_reg_start();
-    public static native long JNI_read_ts_reg_stop();
+    public static native long JNI_read_core_cycles();
 
     public static void main(String args[]) throws
         ClassNotFoundException, NoSuchMethodException, InstantiationException, IllegalAccessException, java.lang.reflect.InvocationTargetException {
@@ -168,11 +169,13 @@ class IterationsRunner {
         Object instance = constructors[0].newInstance();
         BaseKrunEntry ke = (BaseKrunEntry) instance; // evil
 
-        double[] iter_times = new double[iterations];
-        Arrays.fill(iter_times, 0);
+        IterationsRunner.JNI_libkruntime_init();
+
+        double[] wallclockTimes = new double[iterations];
+        Arrays.fill(wallclockTimes, 0);
 
         /*
-         * TSR values are unsigned 64-bit, whereas Java long is *signed*
+         * Core-cycle values are unsigned 64-bit, whereas Java long is signed
          * 64-bit. We are safe since:
          *
          *   - The only arithmetic we use is subtract, which is the same
@@ -181,20 +184,19 @@ class IterationsRunner {
          *   - We print the result using toUnsignedString, thus interpreting
          *     the number as unsigned.
          */
-        long[] tsr_iter_times = new long[iterations];
-        Arrays.fill(tsr_iter_times, 0);
+        long[] cycleCounts = new long[iterations];
+        Arrays.fill(cycleCounts, 0);
 
-        // Please, no reflection inside the timed code!
         for (int i = 0; i < iterations; i++) {
             if (debug) {
                 System.err.println("[iterations_runner.java] iteration: " + (i + 1) + "/" + iterations);
             }
 
-            double startTime = IterationsRunner.JNI_clock_gettime_monotonic();
-            long tsrStartTime = IterationsRunner.JNI_read_ts_reg_start();
+            long cyclesStart = IterationsRunner.JNI_read_core_cycles();
+            double wallclockStart = IterationsRunner.JNI_clock_gettime_monotonic();
             ke.run_iter(param);
-            long tsrStopTime = IterationsRunner.JNI_read_ts_reg_stop();
-            double stopTime = IterationsRunner.JNI_clock_gettime_monotonic();
+            double wallclockStop = IterationsRunner.JNI_clock_gettime_monotonic();
+            long cyclesStop = IterationsRunner.JNI_read_core_cycles();
 
             // Instrumentation mode emits a JSON dict onto a marker line.
             if (instrument) {
@@ -205,23 +207,37 @@ class IterationsRunner {
                 sb.setLength(0);  // clear
             }
 
-            iter_times[i] = stopTime - startTime;
-            tsr_iter_times[i] = tsrStopTime - tsrStartTime;
+            if (wallclockStart > wallclockStop) {
+                System.err.println("wallclock start is greater than stop");
+                System.err.println("start=" + wallclockStart + " stop=" + wallclockStop);
+                System.exit(1);
+            }
+
+            if (Long.compareUnsigned(cyclesStart, cyclesStop) > 0) {
+                System.err.println("cycle count start is greater than stop");
+                System.err.print("start=" + Long.toUnsignedString(cyclesStart) + " ");
+                System.err.println("stop=" + Long.toUnsignedString(cyclesStop) + " ");
+                System.exit(1);
+            }
+
+            wallclockTimes[i] = wallclockStop - wallclockStart;
+            cycleCounts[i] = cyclesStop - cyclesStart;
         }
+
+        IterationsRunner.JNI_libkruntime_done();
 
         // wall clock
         System.out.print("[[");
         for (int i = 0; i < iterations; i++) {
-            System.out.print(iter_times[i]);
+            System.out.print(wallclockTimes[i]);
 
             if (i < iterations - 1) {
                 System.out.print(", ");
             }
         }
         System.out.print("], [");
-        // TSR
         for (int i = 0; i < iterations; i++) {
-            System.out.print(Long.toUnsignedString(tsr_iter_times[i]));
+            System.out.print(Long.toUnsignedString(cycleCounts[i]));
 
             if (i < iterations - 1) {
                 System.out.print(", ");
--- a/iterations_runners/iterations_runner.rb
+++ b/iterations_runners/iterations_runner.rb
@@ -1,5 +1,7 @@
-class AssertionError < RuntimeError
-end
+# The VM needs to be patched to offer up:
+#   libkruntime_init()
+#   libkruntime_done()
+#   read_core_cycles()
 
 # defined this way so we don't measure the conditional platform check.
 if /linux/ =~ RUBY_PLATFORM then
@@ -12,17 +14,12 @@ else
     end
 end
 
-def assert(cond)
-    if not cond
-        raise AssertionError
-    end
-end
-
 # main
 if __FILE__ == $0
     if ARGV.length != 5
-        puts "usage: iterations_runner.rb <benchmark> <# of iterations> "\
-             "<benchmark param> <debug flag> <instrument flag>\n"
+        STDERR.puts "usage: iterations_runner.rb <benchmark> "\
+                    "<# of iterations> <benchmark param> <debug flag> "\
+                    "<instrument flag>\n"
         Kernel.exit(1)
     end
 
@@ -31,11 +28,12 @@ if __FILE__ == $0
     param = Integer(param)
     debug = Integer(debug) > 0
 
-    assert benchmark.end_with?(".rb")
     require("#{benchmark}")
 
-    iter_times = [0] * iters
-    tsr_iter_times = [0] * iters
+    wallclock_times = [0] * iters
+    cycle_counts = [0] * iters
+
+    libkruntime_init();
 
     for iter_num in 0..iters - 1 do
         if debug then
@@ -43,26 +41,40 @@ if __FILE__ == $0
             STDERR.flush  # JRuby doesn't flush on newline.
         end
 
-        start_time = clock_gettime_monotonic()
-        tsr_start_time = read_ts_reg_start()
+        cycles_start = read_core_cycles()
+        wallclock_start = clock_gettime_monotonic()
         run_iter(param)
-        tsr_stop_time = read_ts_reg_stop()
-        stop_time = clock_gettime_monotonic()
+        wallclock_stop = clock_gettime_monotonic()
+        cycles_stop = read_core_cycles()
 
-        iter_times[iter_num] = stop_time - start_time
-        tsr_iter_times[iter_num] = tsr_stop_time - tsr_start_time
+        if wallclock_start > wallclock_stop
+            STDERR.puts "wallclock start greater than stop"
+            STDERR.puts "start=#{wallclock_start} stop=#{wallclock_stop}"
+            exit 1
+        end
+
+        if cycles_start > cycles_stop
+            STDERR.puts "cycles start greater than stop"
+            STDERR.puts "start=#{cycles_start} stop=#{cycles_stop}"
+            exit 1
+        end
+
+        wallclock_times[iter_num] = wallclock_stop - wallclock_start
+        cycle_counts[iter_num] = cycles_stop - cycles_start
     end
+
+    libkruntime_done();
 
     STDOUT.write "[["
     for iter_num in 0..iters - 1 do
-        STDOUT.write String(iter_times[iter_num])
+        STDOUT.write String(wallclock_times[iter_num])
         if iter_num < iters - 1 then
             STDOUT.write ", "
         end
     end
     STDOUT.write "], ["
     for iter_num in 0..iters - 1 do
-        STDOUT.write String(tsr_iter_times[iter_num])
+        STDOUT.write String(cycle_counts[iter_num])
         if iter_num < iters - 1 then
             STDOUT.write ", "
         end
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -442,7 +442,13 @@ class UnixLikePlatform(BasePlatform):
 
         # Create fresh user
         debug("Create krun user")
-        args = self.change_user_args("root") + ["useradd", "-m", BENCHMARK_USER]
+        args = self.change_user_args("root") + ["useradd", "-m"]
+
+        # On Linux, the krun user has to be in the root group for rmsr access
+        if isinstance(self, LinuxPlatform):
+            args += ["-g", "root"]
+        args += [BENCHMARK_USER]
+
         run_shell_cmd(" ".join(args))
 
         # This should now not raise
@@ -1132,6 +1138,23 @@ class LinuxPlatform(UnixLikePlatform):
         # Needed on Linux because the default mail spool ownership causes
         # non-zero userdel exit status.
         return ["-f"]  # force
+
+    def set_msr_dev_permissions(self, who=None):
+        """Change filesystem perimssions on Linux msr device nodes so that 'who'
+        (or if None the current user) can read and write.
+
+        Be aware that capabilities(7) also protects these device nodes."""
+
+        if who is None:
+            who = os.getuid()
+        who = str(who)
+
+        for dev in glob.glob("/dev/cpu/*/msr"):
+            args = self.change_user_args(user="root") + ["chmod", "660", dev]
+            run_shell_cmd(" ".join(args))
+
+            args = self.change_user_args(user="root") + ["chgrp", who, dev]
+            run_shell_cmd(" ".join(args))
 
 
 class DebianLinuxPlatform(LinuxPlatform):
--- a/krun/results.py
+++ b/krun/results.py
@@ -18,8 +18,8 @@ class Results(object):
 
         # Maps key to results:
         # "bmark:vm:variant" -> [[e0i0, e0i1, ...], [e1i0, e1i1, ...], ...]
-        self.data = dict()      # wall-clock times
-        self.tsr_data = dict()  # timestamp register times
+        self.data = dict()              # wall-clock times
+        self.core_cycles_data = dict()  # core cycles all cores
 
         self.reboots = 0
 
@@ -69,7 +69,7 @@ class Results(object):
                 for variant in vm_info["variants"]:
                     key = ":".join((bmark, vm_name, variant))
                     self.data[key] = []
-                    self.tsr_data[key] = []
+                    self.core_cycles_data[key] = []
                     self.instr_data[key] = defaultdict(list)
                     self.eta_estimates[key] = []
 
@@ -92,7 +92,7 @@ class Results(object):
         to_write = {
             "config": self.config.text,
             "data": self.data,
-            "tsr_data": self.tsr_data,
+            "core_cycles_data": self.core_cycles_data,
             "instr_data": self.instr_data,
             "audit": self.audit.audit,
             "reboots": self.reboots,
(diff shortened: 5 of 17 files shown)


vext01:
Oh and I need to fix tests.

vext01:
And I need to make sure this didn't break BSD.

README.md
```diff
@@ -435,7 +435,7 @@ up correctly).
 ## Security Notes
 
 Krun is not intended to be run on a secure multi-user system, as it uses sudo
-to elevate proveleges.
+to elevate priveleges.
```
ltratt: This is still spelt wrong ;)

ltratt:
Good question about the ordering. When we were using the TSR, it made sense to do that immediately around the benchmark, as it was a much faster operation than reading wall clock time. However, reading the wall clock time will clearly be faster than the MSR dance (which involves file opening etc. etc.). So I'm inclined to do:

```c
    read_core_cycles();
    read_wallclock_time();
    do_benchmark();
    read_wallclock_time();
    read_core_cycles();
```

I can be convinced otherwise though!

vext01:
reading the MSR does not require opening, in case that sways you. I open the file outside the timed loop and hold it open.

ltratt:
Ah, didn't realise that. Then I don't know what the overhead is. We could use CF's "empty" experiment with wall-clock time to compare that vs. the MSR read?

vext01:
A little C program would probably do, but it's not likely to prove much, as when you come to time the MSR read, you will also have the overhead of the monotonic clock mixed in(?)

vext01:
Let me experiment a bit.

vext01:
@ltratt have a look at this. I *think* it shows that reading MSRs are more expensive than reading the clock (on bencher3). Please check my reasoning and my brain approaches mush very soon:


```
void                                                                            
test_msr_time(void)                                                             
{                                                                               
    double t1, t2, t3, t4, delta1, delta2;                                      
    uint64_t c1, c2;                                                            
                                                                                
    // time doing "nothing"                                                     
    t1 = clock_gettime_monotonic();                                             
    t2 = clock_gettime_monotonic();                                             
    delta1 = t2 - t1;                                                           
                                                                                
    // time two msr reads                                                       
    t3 = clock_gettime_monotonic();                                             
    c1 = read_core_cycles();                                                    
    c2 = read_core_cycles();                                                    
    t4 = clock_gettime_monotonic();                                             
    delta2 = t4 - t3;                                                           
                                                                                
    printf("monotonic_start_nothing= %f\n", t1);                                
    printf("monotonic_stop_nothing = %f\n", t2);                                
    printf("monotonic_delta_nothing= %f\n", delta1);                            
                                                                                
    printf("monotonic_start_msrs= %f\n", t3);                                   
    printf("monotonic_stop_msrs = %f\n", t4);                                   
    printf("cycles_u64_start = %" PRIu64 "\n", c1);                             
    printf("cycles_u64_stop = %" PRIu64 "\n", c2);                              
    printf("monotonic_delta_msrs %f\n", delta2);                                
```

3 reps:

```
monotonic_start_nothing= 1474602.459761
monotonic_stop_nothing = 1474602.459762
monotonic_delta_nothing= 0.000002
monotonic_start_msrs= 1474602.459764
monotonic_stop_msrs = 1474602.459950
cycles_u64_start = 165803
cycles_u64_stop = 336592
monotonic_delta_msrs 0.000186
```

```
monotonic_start_nothing= 1474615.221965
monotonic_stop_nothing = 1474615.221966
monotonic_delta_nothing= 0.000002
monotonic_start_msrs= 1474615.221968
monotonic_stop_msrs = 1474615.222154
cycles_u64_start = 166181
cycles_u64_stop = 338595
monotonic_delta_msrs 0.000187
```

```
monotonic_start_nothing= 1474627.868010
monotonic_stop_nothing = 1474627.868012
monotonic_delta_nothing= 0.000002
monotonic_start_msrs= 1474627.868014
monotonic_stop_msrs = 1474627.868201
cycles_u64_start = 165661
cycles_u64_stop = 338391
monotonic_delta_msrs 0.000188
```

vext01:
@cfbolz laurie is away for the rest of the afternoon -- what do you think? Am I sane?

cfbolz:
Agreed. In addition it shows that reading the cycles is not that cheap: 0.2 ms or so

vext01:
You mean reading the clock?

vext01:
Sometimes the "nothing delta" is 0.0 :P So I guess it varies.

cfbolz:
No. Reading the clock (once) is contained in the first variant, so costs about 2*0.000002s=0.004ms

cfbolz:
But reading the msrs costs 0.000188s which is about 0.2ms, two orders of magnitudes larger.

vext01:
Yep, and reading the core cycles is done via an msr.

vext01:
I just pushed a test which encapsulates our newly acquired wisdom.

ltratt:
OK, I think we have out answer -- the monotonic timer should be the inner measurement :)

vext01:
Testing this on OpenBSD now... Anything else that needs to be addressed?

libkrun/test/test_libkruntime.py
```diff
@@ -83,6 +83,21 @@ def test_cycles_u64_double_ratio(self):
...
+        dct = parse_keyvals(out, True)
         # Depends on speed of CPU, but should be very close to 1
         assert 0.95 <= dct["monotonic_delta"] <= 1.05
+
+    def test_msr_time(self):
+        rv, out, _ = invoke_c_prog("msr_time")
+        assert rv == 0
+        dct = parse_keyvals(out)
+
+        # I expect reading MSRs to be more than 70x slower than reading MSRs.
```
ltratt: I think this slowdown is purely because of Linux's file-system mechanism for reading MSRs. I don't think it *has* to be that slow! I think we should drop the 70x thing (which is pretty arbitrary, and may or may not hold on other machines), and just assert that the MSRs are slower than the monotonic timer. We may also want to make this test Linux only? I doubt reading MSRs will be as slow on OpenBSD. Perhaps we can address that later, though.

So we might want to say "Reading an MSR on Linux is *much* slower than reading wall-clock time. If this were to not be the case, we might want to reorder the way that time/MSRs are read."

ltratt:
I think I will have to wait a day or two until I can do a full review of this. Sorry!

libkrun/test/test_libkruntime.py
vext01: The 70x is somewhat arbitrary, but I wanted to give it some room for manauver. We can try > and see what happens.

The test does not run on OpenBSD as we have no way to read core cycles yet, hence it is guarded:
```
@pytest.mark.skipif(not CORECYCLES_SUPPORT, reason="would not make sense")
```

vext01:
No worries. I can amuse myself :)

iterations_runners/iterations_runner.java
```diff
@@ -168,11 +169,13 @@ public static void main(String args[]) throws
...
-        double[] iter_times = new double[iterations];
-        Arrays.fill(iter_times, 0);
+        IterationsRunner.JNI_libkruntime_init();
+
+        double[] wallclockTimes = new double[iterations];
+        Arrays.fill(wallclockTimes, 0);
 
         /*
-         * TSR values are unsigned 64-bit, whereas Java long is *signed*
+         * Core-cycle values are unsigned 64-bit, whereas Java long is *signed*
```
ltratt: I was a bit distracted by the emphasis on `*signed*`. Maybe drop the `*`s?

iterations_runners/iterations_runner.rb
```diff
@@ -31,38 +28,57 @@ def assert(cond)
     param = Integer(param)
     debug = Integer(debug) > 0
 
-    assert benchmark.end_with?(".rb")
+    if ! benchmark.end_with?(".rb")
```
ltratt: Is there any particular reason to a) make this something we check b) make it more complex than an assertion?

libkrun/libkruntime.c
```diff
@@ -1,5 +1,10 @@
 /*
- * C support functions.
+ * C support functions for timing.
+ *
+ * Although this is a library, any error will exit indicating failure. Krun
```
ltratt: I'm not entirely sure I understand this first sentence.

libkrun/libkruntime.c
```diff
@@ -1,5 +1,10 @@
...
+ * will detect this and report appropriately to the user.
+ *
+ * Whole file assumes x86 for now.
```
ltratt: "This whole file..."?

libkrun/libkruntime.c
```diff
@@ -21,6 +28,41 @@
...
+/*
+ * Bitfields of MSR_IA32_FIXED_CTR_CTRL related to fixed counter 1.
+ * AKA, CPU_CLK_UNHLATED.CORE in the Intel manual.
+ */
+// Enable couting in ring 0
+#define EN1_OS      1 << 4
+// Enable counting in higer rings
+#define EN1_USR     1 << 5
+// Enable counting for all core threads (if any)
+#define EN1_ANYTHR 1 << 6
```
ltratt: Slightly inconsistent indentation of `1 << 6`?

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
-Java_IterationsRunner_JNI_1read_1ts_1reg_1stop(JNIEnv *e, jclass c)
+#if defined(__linux__) && !defined(TRAVIS)
+int
+open_msr_node(int core, int flags)
 {
-    return read_ts_reg_stop();
+    char path[MAX_MSR_PATH];
+    int msr_node;
+
+    /* Note this is not the default msr(4) device! */
```
ltratt: Maybe be clearer in the comment that we're tracking `rmsr`, not `msr` (at the moment, the comment requires the reader to know that).

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+    if (snprintf(path, MAX_MSR_PATH, "/dev/cpu/%d/rmsr", core) < 0) {
+        perror("snprintf");
+        exit(EXIT_FAILURE);
+    }
+
+    msr_node = open(path, flags);
+    if (msr_node < 0) {
```
ltratt: I think the only negative number that can be returned is -1. So `if (msr_node == -1)` is probably clearer.

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+
+u_int64_t
+read_msr(int core, long addr)
+{
+    u_int64_t msr_val;
+    int msr_node;
+
+    msr_node = msr_r_nodes[core];
+
+    if (lseek(msr_node, addr, SEEK_SET) < 0) {
```
ltratt: Same -1 comment as above.

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+}
+
+void
+write_msr(int core, long addr, uint64_t msr_val)
+{
+    int msr_node;
+
+    msr_node = msr_w_nodes[core];
+
+    if (lseek(msr_node, addr, SEEK_SET) < 0) {
```
ltratt: Same -1 comment as above.

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+        config_fixed_ctr1(core, enable);
+        write_msr(core, PCTR_IA32_PERF_FIXED_CTR1, 0); // reset
+    }
+}
+
+void
+close_fd(int fd)
+{
+    int ok = close(fd);
+    if (ok < 0) {
```
ltratt: Same -1 comment as above.

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+ * Configure fixed-function counter 1 to count all rings and threads
  */
+void
+config_fixed_ctr1(int core, int enable)
+{
+    u_int64_t msr_val = read_msr(core, MSR_IA32_FIXED_CTR_CTRL);
+    if (enable) {
+        msr_val |= (EN1_OS | EN1_USR | EN1_ANYTHR);
+    } else {
+        msr_val &= ~((u_int64_t) EN1_OS | EN1_USR | EN1_ANYTHR);
```
ltratt: Why does this calculation need the cast but line 176 not?

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+     *   5-12: width of counters
+     */
+    num_fixed_ctrs = edx & 0x1f;
+    fixed_ctr_width = (edx & 0x1fe0) >> 5;
+
+    /* eax
+     *   0-7:  architectural counter version
+     *   8-31: reserved
+     */
+    arch_ctr_vers = eax & (0xff);
```
ltratt: `(0xff)` -- brackets are pointless here.

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+    }
+
+    msr_w_nodes = calloc(num_msr_nodes, sizeof(int));
+    if (msr_w_nodes == NULL) {
+        perror("calloc");
+        exit(EXIT_FAILURE);
+    }
+
+    for (core = 0; core < num_msr_nodes; core++) {
+        msr_r_nodes[core] = open_msr_node(core, O_RDONLY);
```
ltratt: Dumb question: can we / should we open up a read/write file handles, rather than separate read only and write only file handles?

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
+        msr_w_nodes[core] = open_msr_node(core, O_WRONLY);
+    }
+
+    /* Configure and reset CPU_CLK_UNHALTED.CORE on all CPUs */
+    config_fixed_ctr1_all_cores(1);
+#else
+    /*
```
ltratt: I wonder whether, to avoid shooting ourselves in the foot in the future, we might better say something like:
```c
#if defined(__linux__) && !defined(TRAVIS)
    ...
#elif defined(__linux__) && defined(TRAVIS)
   // explicitly say that we're doing nothing
#else
   #error "Unsupported platform"
#endif
```

libkrun/libkruntime.c
```diff
@@ -37,86 +79,254 @@ clock_gettime_monotonic()
...
 {
-    u_int32_t hi, lo;
+#if defined(__linux__) && !defined(TRAVIS)
+    int core;
+    uint64_t cycles = 0, val;
+
+    for (core = 0; core < num_msr_nodes; core++) {
+        val = read_msr(core, PCTR_IA32_PERF_FIXED_CTR1) & pctr_val_mask;
+
+        // Note that overflow is impossible on most platforms. E.g. on an i7,
```
ltratt: "most" is a worrying word here! Should we add an assert along the lines of:
```c
assert(cycles <= INT_MAX - val)
```
?

ltratt:
OK, I have a number of small comments, but nothing particularly significant.

iterations_runners/iterations_runner.rb
vext01: I was thinking of removing it altogether actually. None of the other runners do such a check and it will be pretty obvious what has gone wrong if you pass something with a weird filename.

ltratt: OK, let's nuke it then.

libkrun/libkruntime.c
vext01: It is impolite to exit() in a library. Usually you would return an error condition and allow the consumer to decide what to do.

I can put words to this effect if that makes sense to you.

ltratt: I think it just needs rephrasing to be a bit clearer.

vext01: I think it's ok to exit in libkrun as it is not designed for mainstream use.

ltratt: I'm fine with that -- it's just the English isn't very easy to parse at the moment.

libkrun/libkruntime.c
vext01: nice spot.

libkrun/libkruntime.c
vext01: Because those constants default to a signed 32.

In the first case, when you or `msr_val` (an unsigned 64) with the signed 32 (which is known to be +ve), the compiler will zero extend it, which is fine.

In the second case, we want to turn of all bits other than `EN1_OS | EN1_USR | EN1_ANYTHR`. This means we need the high bits (32-63) to be set in the mask. Without the cast, the 32-bit value would be zero extended and the top 32 bits would not be set.

libkrun/libkruntime.c
vext01: indeed.

libkrun/libkruntime.c
vext01: That's the next PR. I'm not 100% it will work due to the hacky nature of the msr driver, but we can but try.

libkrun/libkruntime.c
vext01: Hrm, I'm not sure what value it adds. I can be convinced otherwise.

libkrun/libkruntime.c
ltratt: OK (this is not a bit of C I'm very familiar with!).

libkrun/libkruntime.c
ltratt: At the moment, if you compile on an unsupported platform, things will silently compile, but without telling you that you're missing support for a feature that your platform may support. It's much easier, IMHO, to have the code raise an error which says "look at this bit!" rather than having to grep it later (assuming you even know what to grep for).

libkrun/libkruntime.c
vext01: Yeah we could if we are happy with that being in the timed section. Should be insignificant I would have thought. 

I would not use `assert` in C bit rather `printf` and `exit`.

FWIW, I did a "back of a napkin" calculation and figured you would need several thousand cores to trigger an overflow ;)

ltratt: Even for a really long-running benchmark like CPython? If that's the case then the assert is pretty pointless.

libkrun/libkruntime.c
vext01: Actually I think I am wrong!

```
int                                                                             
main(int argc, char **argv)                                                     
{                                                                               
    // turn on only the bottom bit                                              
    printf("%0" PRIX64 "\n", ALL_ZEROS | 0x1);                                  
                                                                                
    // turn off only bottom bit                                                 
    printf("%0" PRIX64 "\n", ALL_ONES & ~0x1);                                  
                                                                                
                                                                                
    return (EXIT_SUCCESS);                                                      
}                                                                               
```

prints:

```
1
FFFFFFFFFFFFFFFE
```

I was expecting `~0x1` to make `0x00000000fffffffe`, then anding with `0xffffffffffffffff` to make `0x00000000fffffffe`, bit it seems that C is aware than I am anding with a 64-bit. Today I learned.

Please check my working.

libkrun/libkruntime.c
vext01: i see

libkrun/libkruntime.c
ltratt: I *think* (but this is not something I'm an expert in), that it will automatically cast the RHS up to 64 bits so that the cast isn't needed.

libkrun/libkruntime.c
vext01: Do you think we should use this pattern for every instance of `#elif defined(__linux__) && defined(TRAVIS)`, or just in the case you commented above?

ltratt: My inclination would be to do it everywhere. It makes subsequent porting much easier IME.

vext01:
@ltratt like the last commit with the portability improvements? Need to test on BSD.

vext01:
I can already see a missing openbsd case >.<

ltratt:
@vext01 Yep, this is starting to look good. Thanks!

vext01:
And also for cases like:

```
#ifdef WITH_JAVA                                                                
#include <jni.h>                                                                
#endif
```

or just on the new #error?

ltratt:
I guess we should consistently do the # indent thing, but you don't necessarily need to fix them all in one go.

vext01:
The thing is, in some cases it would be quite weird to do the indent, at-least not without also indenting the globally scope C stuff too. If you see what I mean.

I tend to always put macros hard against the left margin, but I do understand that it can be hard to read.

vext01:
And that test failure is bogus, it was an SSL failure in apt-get.

ltratt:
I don't feel super-super strongly about the indentation thing.

vext01:
OK, so I think we can leave the indent as is, and if you want to experiment with different styles, we can do it all at once in another PR.

ltratt:
Sounds like a plan.

vext01:
alright, I expanded the comment a little anyway.

vext01:
And this all seems to be OK on BSD.

krun/platform.py
```diff
@@ -442,7 +442,13 @@ def make_fresh_krun_user(self):
 
         # Create fresh user
         debug("Create krun user")
-        args = self.change_user_args("root") + ["useradd", "-m", BENCHMARK_USER]
+        args = self.change_user_args("root") + ["useradd", "-m"]
+
+        # On Linux, the krun user has to me in the root group for rmsr access
```
cfbolz: typo, should be "has to be"

krun/results.py
```diff
@@ -18,8 +18,8 @@ def __init__(self, config, platform, results_file=None):
 
         # Maps key to results:
         # "bmark:vm:variant" -> [[e0i0, e0i1, ...], [e1i0, e1i1, ...], ...]
-        self.data = dict()      # wall-clock times
-        self.tsr_data = dict()  # timestamp register times
+        self.data = dict()              # wall-clock times
+        self.core_cycles_data = dict()  # core cycles all cores
```
cfbolz: those two could just be ``defaultdict(list)``, then you don't need to initialize them below.

vext01: I think we discussed this in the past and decided that it would be beneficial to use a normal dict because it ensures that use of a malformed key make an error.

cfbolz: ah, makes sense!

iterations_runners/iterations_runner.java
```diff
@@ -181,20 +184,20 @@ public static void main(String args[]) throws
          *   - We print the result using toUnsignedString, thus interpreting
          *     the number as unsigned.
          */
-        long[] tsr_iter_times = new long[iterations];
-        Arrays.fill(tsr_iter_times, 0);
+        long[] cycleCounts = new long[iterations];
+        Arrays.fill(cycleCounts, 0);
 
         // Please, no reflection inside the timed code!
```
cfbolz: hm, this is just an unrelated observation, but we should remove that comment, we haven't been using reflection since ages. Relatedly, we can probably remove some of the exceptions from the declaration of main. Should I make an issue so we don't forget?

vext01: The runner does use reflection to get the named benchmark:

```
        // reflectively call the benchmark's run_iter                           
        Class<?> cls = Class.forName(benchmark);                                
        java.lang.reflect.Constructor<?>[] constructors = cls.getDeclaredConstructors();
        assert constructors.length == 1;                                        
        Object instance = constructors[0].newInstance();                        
        BaseKrunEntry ke = (BaseKrunEntry) instance; // evil                    
```

That said, I do wonder of the worth of the comment. Maybe we should remove it.

vext01:
fixed those bits @cfbolz thanks.

cfbolz:
Then imo this is looking pretty good!

ltratt:
Let's squash this one, then I can merge at the appropriate point.

vext01:
Do you mind if I splat this all into one?

ltratt:
As you wish.

vext01:
Check that out for a monster diff!

PR #240, opened 2016-09-14
Core cycles msr improvements

vext01:
Here are a couple of miscellaneous improvements to the work we merged yesterday.

See individual commit messages.

Thanks

### DIFF
--- a/iterations_runners/iterations_runner.c
+++ b/iterations_runners/iterations_runner.c
@@ -54,8 +54,8 @@ main(int argc, char **argv)
     int     (*krun_bench_func)(int); /* func ptr to benchmark entry */
     double    krun_wallclock_start = -1, krun_wallclock_stop = -1;
     double   *krun_wallclock_times = NULL;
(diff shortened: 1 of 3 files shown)


ltratt:
If `u_int64_t` is internal, should we also delete its definition?

ltratt:
Otherwise, this looks good and ready for merge.

vext01:
I think undefining `u_int64_t` would be asking for trouble. On some systems the system headers use them, but in general you cannot guarantee their existence:

http://stackoverflow.com/questions/5090737/datatypes-used-in-c

vext01:
(I would like to squash the two type changes into one commit)

ltratt:
Ah, so we don't define u_int64_t ourselves?

vext01:
No, it's defined in the system headers on some systems.

vext01:
(C is a mess)

ltratt:
I'm with you! OK, let's squash then I can merge.

vext01:
Splat.

PR #244, opened 2016-09-16
Dont load config twice

cfbolz:
Previously the config file was loaded twice when resuming, which could lead to funny effects. Now, I instead check whether there are differences to the previously loaded file, and complain if there are.


vext01:
Looks good to me. I will test next.

vext01:
The diff is a nice touch, but can we tweak the formatting?

```
[2016-09-16 14:32:36 ERROR] the config file test.krun on disk was different than the resumed config file:
--- 

+++ 

@@ -107,7 +107,7 @@

 
 
 BENCHMARKS = {
-     'binarytrees': 25,
+    'binarytrees': 25,
     'richards': 500,
     'spectralnorm': 3,
     'nbody': 15,

```

 * A fair amount of spurious whitespace, e.g. between `---` and `+++` and again later
 * You usually include an old and new file name in the `---` and `+++` lines. For use we could use "blah.krun" and "cached".

What do you think?

krun/config.py
```diff
@@ -41,16 +41,13 @@ def _fatal_exception_execing_config(self, exc_info):
...
-        self.__dict__.update(config_dict)
-        self.filename = ""
-        self.text = config_str
+    def check_config_consistency(self, config_str):
+        import difflib
+        if self.text != config_str:
+            diff = "\n".join(difflib.unified_diff(self.text.splitlines(),
+                                                  config_str.splitlines()))
+            fatal("the config file %s on disk was different than "
+                  "the resumed config file:\n%s" % (self.filename, diff))
```
vext01: nitpick: both config files are on disk. Can we say: ```The config file blah.krun was changed mid experiment``` followed by the diff.

vext01: Sorry, typo, `mid-experiment` with a hyphen.

vext01:
Works for me. Minor comments.

krun/config.py
cfbolz: Sorry, I find "mid-experiment" less precise. How about "the config file ... is different than the stored copy in ..._result.json.bz2"

vext01: You are right, we could be more precise. I'm trying to give an error message that will also make sense to a user who doesn't know that we cache the config file (I had forgotten, for example).

How about:

```
The experiment is in an inconsistent state as config file x.krun has changed since it was initially cached in x_results.json.bz2
```

(Maybe I am over-thinking)

cfbolz: fine

cfbolz:
the extra empty lines is a strange feature of difflib, do we really care to remove them? I'd be happy to add filenames

krun/config.py
```diff
@@ -41,16 +41,13 @@ def _fatal_exception_execing_config(self, exc_info):
...
-        except Exception as e:
-            self._fatal_exception_execing_config(sys.exc_info())
-
-        self.__dict__.update(config_dict)
-        self.filename = ""
-        self.text = config_str
+    def check_config_consistency(self, config_str):
+        import difflib
+        if self.text != config_str:
+            diff = "\n".join(difflib.unified_diff(self.text.splitlines(),
```
vext01: Does passing `lineterm=""` fix the whitespace?

cfbolz: ah, yes. but I'll do it differently then.

vext01:
Want to squash?

vext01:
Thanks for that.

PR #246, opened 2016-09-19
Amperf

vext01:
Let's start looking at this.

This adds support for {A,M}PERF to Krun.

Please don't merge, as there may be a bug! APERF deltas are coming back larger than MPERF deltas. This suggests either a bug, or we misunderstood the meaning of these counters.

Maybe code review will spot a bug.

(There will be a `warmup_experiment` PR to accompany this one later.)

Branch also needs to be synced.

### DIFF
--- a/iterations_runners/iterations_runner.c
+++ b/iterations_runners/iterations_runner.c
@@ -52,10 +52,18 @@ main(int argc, char **argv)
     int       krun_debug = 0;
     void     *krun_dl_handle = 0;
     int     (*krun_bench_func)(int); /* func ptr to benchmark entry */
+
     double    krun_wallclock_start = -1, krun_wallclock_stop = -1;
     double   *krun_wallclock_times = NULL;
+
     uint64_t krun_cycles_start = 0, krun_cycles_stop = 0;
     uint64_t *krun_cycle_counts = NULL;
+
+    uint64_t krun_aperf_start, krun_aperf_stop;
+    uint64_t *krun_aperf_counts;
+
+    uint64_t krun_mperf_start, krun_mperf_stop;
+    uint64_t *krun_mperf_counts;
 
     if (argc != 6) {
         printf("usage: iterations_runner_c "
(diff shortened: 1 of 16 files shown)


ltratt:
I wonder if we should really be resetting APERF/MPERF to 0... They're probably used by the kernel, so it may be that resetting them to 0 causes weird things (possibly the problem you note).

vext01:
Maybe, I'm just boiling down a minimal program that shows the problem. Hopefully I will spot a bug.

vext01:
Interesting.

This is what seems to be happening.

The APERF and MPERF values are almost identical, but we read APERF outermost, and the difference introduced by a non-atomic APERF then MPERF read is enough to make aperf always larger.

If we swap the nesting order around, mperf is always larger.

I'll now get lunch and think.

vext01:
By the way, the fact that the values are almost identical is a wonderful thing for us.

ltratt:
Haha, yes, that's good news!

vext01:
I think we are OK to swap the order of the aperf and mperf measurements, thoughts?

ltratt:
I'm fine with swapping the order, though I think that doesn't fix the fundamental problem: we won't ever quite end up with a round number. I suspect that the better fix is simply to take the APERF/MPERF numbers as-is (whichever way around), and then round the resulting division (to, say, a couple of DP, which I think is what Linux does). Very minor differences in the cycle count should then get lost in the noise.

vext01:
I tend to agree, but I would also like to swap the order.

Note that the division does not happen in Krun, and raw figures are simply dumped into the result file.

If you are good with the previous commit, then I will sync the branch.

When reviewing, please look for 'a's which should be 'm's and 'start's which should be 'stop's etc. Most of this change is all copy/paste/modify and is error prone.

ltratt:
OK, I'm fine with that.

BTW, notice we have a merge conflict that needs fixing.

vext01:
Yep, that's the next thing to fix. Likely Carl's changes.

vext01:
Synced branch. Once reviewed I will rebase and squash a bit.

ltratt:
OK, this looks good (I went through it quite carefully and nothing jumped out at me). Squash and then I'll merge.

vext01:
Here we are.

ltratt:
OK, will wait for tests then merge.

PR #249, opened 2016-09-29
refactor careful reading of stdout/err of benchmarking processes

cfbolz:
- move the code from vm_defs.py to a new helper function
read_popen_output_carefully in util.py
- add a new function run_shell_cmd_bench that makes use of that function
(we can't change run_shell_cmd, because for we need a platform object in
read_popen_output_carefully, but run_shell_cmd is used in setting up a
platform).

### DIFF
--- a/krun/util.py
+++ b/krun/util.py
@@ -1,10 +1,23 @@
 import json
 import os
 import re
+import select
 from subprocess import Popen, PIPE
 from logging import error, debug, info
 
 FLOAT_FORMAT = ".6f"
+
+SELECT_TIMEOUT = 1.0
+
+# Pipe buffer sizes vary. I've seen reports on the Internet ranging from a
+# page size (Linux pre-2.6.11) to 64K (Linux in 2015). Ideally we would
+# query the pipe for its capacity using F_GETPIPE_SZ, but this is a) not
+# portable between UNIXs even, and b) not exposed by Python's fcntl(). For
+# now, we use a "reasonable" buffer size. If it is larger than the pipe
+# capacity, then no harm done; if it is smaller, then we may do more reads
+# than are strictly necessary. In either case we are safe and correct.
+PIPE_BUF_SZ = 1024 * 16
+
 
 DIR = os.path.abspath(os.path.dirname(__file__))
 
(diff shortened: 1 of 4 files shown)


vext01:
Looks good to me.

If I had to nitpick (which I don't have to ;) ), I would add a test the stderr return value on the new function.

krun/util.py
```diff
@@ -76,6 +104,89 @@ def run_shell_cmd_list(cmds, failure_fatal=True, extra_env=None):
...
+            emit = d[startindex:nl]
+            info("stderr: " + "".join(stderr_partial_line) + emit)
+            stderr_partial_line = []
+            startindex = nl + 1
+
+def read_popen_output_carefully(process, platform, print_stderr=True):
+    """ helper function: given a process, read new data whenever it is
+    available, to make sure that the process is never blocked when trying to
+    write to stdout/stderr. """
+    from krun.platform import detect_platform
```
ltratt: [I know this code is sort-of copy and pasted, but sometimes a fresh look at things raises new questions!]

I think the `import` statement should be inside the `if` since that's the only place that `detect_platform` is called. Also... I wonder why this function even *does* platform detection itself. Shouldn't users be forced to do this?

ltratt:
I have one question (that's not directly part of the PR in a sense), but am otherwise OK with this PR.

krun/util.py
cfbolz: yes, I think we should just kill that

ltratt: Should we just replace it with `assert platform is not None` then?

vext01: Looks bogus yeah.

ltratt:
I'm fine with the updates. Assuming Edd is too, then squash as desired, and I'll merge?

vext01:
Looks good.

I wonder why I put detect platform inside if no-one ever passes a None platform. Maybe someone used to...

cfbolz:
squashed

ltratt:
We have a failure on Travis from one of the new tests?

cfbolz:
eh, simply fails locally too. fixing

ltratt:
It's still failing? Can we just push an update commit without a force push, to see what's happening?

ltratt:
OK, so I think we're ready to merge now?

cfbolz:
I think so. @vext01?

PR #251, opened 2016-10-04
Per-core measurements

vext01:
This overhauls krun to support per-core measurements for core-cycles, aperf and mperf.

Not ready to merge just yet, as I need to test on BSD (and fix some tests it seems), but we can start reviewing.

It's a big PR, I'm afraid. Nevertheless, please check super-carefully.


iterations_runners/iterations_runner.c
```diff
@@ -136,61 +156,40 @@ main(int argc, char **argv)
...
+        krun_wallclock_stop = krun_get_wallclock(1);
 
-        if (krun_wallclock_start > krun_wallclock_stop) {
-            fprintf(stderr, "wallclock time start greater than stop\n");
-            fprintf(stderr, "start=%f, stop=%f\n",
-                krun_wallclock_start, krun_wallclock_stop);
-            exit(EXIT_FAILURE);
-        }
+        krun_wallclock_times[krun_iter_num] =
+            krun_wallclock_stop - krun_wallclock_start;
```
ltratt: I can't help feel that the intermediate variables are a bit confusing. Can we not just say:

```krnu_wallclock_times[krun_iter_num] = krun_get_wallclock(POST) - krun_get_wallclock(PRE);```

or similar?

File: iterations_runners/iterations_runner.c [hunk 2]
```diff
@@ -136,61 +156,40 @@ main(int argc, char **argv)
...
 
-        if (krun_aperf_start > krun_aperf_stop) {
-            fprintf(stderr, "aperf start greater than stop\n");
-            fprintf(stderr, "start=%" PRIu64 ", stop=%" PRIu64 "\n",
-                krun_aperf_start, krun_aperf_stop);
-            exit(EXIT_FAILURE);
-        }
+        /* Same for per-core measurements */
+        for (krun_core = 0; krun_core < krun_num_cores; krun_core++ ) {
```
ltratt: Similar comment to earlier. I don't think the intermediate variables make things clearer (given that they're only used once).

iterations_runners/iterations_runner.java
```diff
@@ -131,12 +131,35 @@ public void getInstJSON(StringBuilder sb, int iterNum) {
...
+
+    /* Prints signed longs for the per-core measurements */
+    private static void emitPerCoreResults(String name, int numCores, long[][] ary) {
+        System.out.print("\"" + name + "\": [");
+
+        for (int core = 0; core < numCores; core++) {
+            System.out.print("[");
+            for (int i = 0; i < ary[core].length; i++) {
+                System.out.print(Long.toUnsignedString(ary[core][i]));
+
```
ltratt: Extraneous blank line.

krun/results.py
```diff
@@ -183,17 +191,18 @@ def strip_results(self, key_spec):
     def append_exec_measurements(self, key, measurements):
         """Unpacks a measurements dict into the Results instance"""
 
+        # Has already been check earlier, but let's be defensive.
```
ltratt: "checked"

ltratt: Though if it has been checked earlier, doing it a second time seems a bit pointless, unless it could have changed?

krun/util.py
```diff
@@ -92,16 +96,32 @@ def check_and_parse_execution_results(stdout, stderr, rc):
...
         err_s += "stderr:\n%s\n%s\n%s\n" % (rule, stderr, rule)
         raise ExecutionFailed(err_s)
 
-    # [wall-clock times, core-cycle counts, aperf_counts, mperf_counts]
-    assert len(json_data) == 4
+    # Check we have the right keys
+    key_set = set(json_data.keys())
+    if key_set != EXPECT_JSON_KEYS:
+        err_s = ("Benchmark emitted unexpected JSON keys\n")
+        err_s += ("Expected: %s, got: %s" % (EXPECT_JSON_KEYS, key_set))
```
ltratt: Why the brackets on these two lines?

libkrun/libkruntime.c
```diff
@@ -29,15 +29,70 @@
 #define ACTUAL_CLOCK_MONOTONIC    CLOCK_MONOTONIC
 #endif
 
+#define CORE_BOUNDS_CHECK(C)                                                \
```
ltratt: I still can't help but feel these should be functions and not macros -- is there a good reason for them being macros? C macros are dangerous, so they always require more careful thought than normal functions.

libkrun/libkruntime.c
```diff
@@ -342,126 +419,213 @@ libkruntime_done(void)
...
     return 0;
 #else
 #error "Unsupported platform"
 #endif
 }
 
-#if defined(__linux__) && !defined(TRAVIS)
 /*
- * Sum the values of an MSR across all cores.
+ * Check for double precision loss.
```
ltratt: This comment could do with tightening and shortening: it's a bit repetitive at the moment. I'd like to keep the technical info in, though, which really helps readers.

libkrun/libkruntime.c
```diff
@@ -342,126 +419,213 @@ libkruntime_done(void)
...
 
-        // Check for overflow
-        if (sum > UINT64_MAX - val) {
-            fprintf(stderr, "Sum overflow!: "
-                    "%" PRIu64 " + %" PRIu64 "\n", sum, val);
-            exit(EXIT_FAILURE);
+#if defined(__linux__) && !defined(TRAVIS)
+    // We support per-core readings
+
+    // We do it this way to ensure correct nesting of reads
```
ltratt: I think we can add a bit more to this. Something like "The quickest, and most important, reading is krun_read_core_cycles, which is the innermost call before and after benchmarking. Although APERF/MPERF have separate functions, they are really one operation, so they are done in the same order before/after benchmarking (not, as you might expect, in reverse order after benchmarking)." Well, that's a bit rubbish, but you get the idea I hope!

libkrun/libkruntime.c
```diff
@@ -342,126 +419,213 @@ libkruntime_done(void)
...
-        sum += val;
+        krun_check_mdata();
     }
-
-    return sum;
-}
 #elif defined(__linux__) && defined(TRAVIS)
-// function not used on non-linux or travis
+    data->wallclock = krun_clock_gettime_monotonic();
+    if (mdata_idx == 1) {
```
ltratt: Why is the `if` inside a platform check? Shouldn't it be always executed and thus outside the platform check?

libkrun/libkruntime.c
```diff
@@ -342,126 +419,213 @@ libkruntime_done(void)
...
+            bail = 1;
+        }
+
+        if (krun_mdata[0].mperf[core] > krun_mdata[1].mperf[core]) {
+            fprintf(stderr, "mperf error: "
+                    "start=%" PRIu64 ", stop=%" PRIu64 "\n",
+                    krun_mdata[0].mperf[core], krun_mdata[1].mperf[core]);
+            bail = 1;
+        }
+
```
ltratt: Extraneous blank line.

libkrun/libkruntime.c
```diff
@@ -342,126 +419,213 @@ libkruntime_done(void)
...
- * For languages like Lua, where there is suitible integer type
+ * Check all of the measurements for issues.
  */
-double
-read_core_cycles_double()
+void
+krun_check_mdata(void)
 {
-    return u64_to_double(read_core_cycles());
+    int bail = 0, core;
```
ltratt: `bail` seems pointless: just call exit (or do a goto) at the appropriate point, rather than printing out lots of errors?

ltratt:
Not that it's your fault as such, but this PR is way too big to make careful review possible. All my comments tend towards the shallow side because of that. If there are some bits you *really* would like us to focus on, you could explicitly point us at them?

iterations_runners/iterations_runner.c
```diff
@@ -136,61 +156,40 @@ main(int argc, char **argv)
         }
 
         /* Start timed section */
-        krun_mperf_start = read_mperf();
-        krun_aperf_start = read_aperf();
-        krun_cycles_start = read_core_cycles();
-        krun_wallclock_start = clock_gettime_monotonic();
-
+        krun_measure(0);
```
vext01: We can yeah, but I don't see it as completely necessary. The reason I used 0 and 1 is because I was not certain all the languages support symbolic constants, so we might still end up with some 0s and 1s for some languages.

FWIW, I don't think of the 0s and 1s as start and stop, but as an array index, hence the parameter name is `mdata_index`.

ltratt: OK.

krun/results.py
vext01: I was just being defensive.

libkrun/libkruntime.c
vext01: It would appear in a #else then, but I thought you wanted #else to be the #error case?

krun/util.py
vext01: Probably was a long line once. Will fix.

libkrun/libkruntime.c
vext01: I can make them functions. I think I was trying to ensure no extra calls in the timed section, but it's probably negotiable.

krun/results.py
ltratt: Is it useful to do so? Too much checking can be confusing IME. But this is a judgement call, and I'm happy with whatever you decide here.

libkrun/libkruntime.c
vext01: Yeah. I don't think we would ever care for the results once one error is printed.

libkrun/libkruntime.c
ltratt: The compiler will almost certainly inline them, but I agree, the tiny performance difference either way is completely irrelevant -- let's favour readability.

vext01:
Just look for incorrect mdata indicies, aperf that should be mperfs etc.

iterations_runners/iterations_runner.java
```diff
@@ -131,12 +131,35 @@ public void getInstJSON(StringBuilder sb, int iterNum) {
...
+    public static native void JNI_krun_done();
+    public static native void JNI_krun_measure(int mindex);
+    public static native double JNI_krun_get_wallclock(int mindex);
+    public static native long JNI_krun_get_core_cycles(int mindex, int core);
+    public static native long JNI_krun_get_aperf(int mindex, int core);
+    public static native long JNI_krun_get_mperf(int mindex, int core);
+    public static native int JNI_krun_get_num_cores();
+
+    /* Prints signed longs for the per-core measurements */
+    private static void emitPerCoreResults(String name, int numCores, long[][] ary) {
```
vext01: array

vext01: I'll rename it.

iterations_runners/iterations_runner.java
```diff
@@ -188,32 +212,26 @@ public static void main(String args[]) throws
...
-        long[] mperfCounts = new long[iterations];
-        Arrays.fill(mperfCounts, 0);
+        long[][] cycleCounts = new long[numCores][iterations];
+        long[][] aperfCounts = new long[numCores][iterations];
+        long[][] mperfCounts = new long[numCores][iterations];
+
+        /* longs default to zero, but we explicitely zero to force allocation */
+        for (int core = 0; core < numCores; core++) {
+            Arrays.fill(cycleCounts[core], 0);
+            Arrays.fill(mperfCounts[core], 0);
```
snim2: This is a rather anal comment, but in other places you have `aperf` after `cycles` and `mperf` last; to avoid copy and paste errors it might be better to be consistent about that.

vext01: A valid comment. Will fix :)

krun/util.py
```diff
@@ -92,16 +96,32 @@ def check_and_parse_execution_results(stdout, stderr, rc):
...
+        raise ExecutionFailed(err_s)
+
+    # Check lengths
+    # (All lists apart from wallclock_times are per-core)
```
snim2: Not all lists, I think. `audit` isn't per-core

snim2:
I can't see anything obviously wrong here, but I can't promise that I haven't missed anything!

krun/util.py
vext01: All measurement lists -- good point though.

vext01:
I think that addresses everything.

With the variable folding, if the computation didn't fit on one line, I have favoured lining up each call on it's own line -- I find it makes checking the mdata indices easier. Hope that is OK.

iterations_runners/iterations_runner.py
```diff
@@ -86,24 +86,19 @@
...
-
-            cycle_counts[core][i] = cycles_stop - cycles_start
-            aperf_counts[core][i] = aperf_stop - aperf_start
-            mperf_counts[core][i] = mperf_stop - mperf_start
+            cycle_counts[core][i] = (
+                krun_get_core_cycles(1, core) -
+                krun_get_core_cycles(0, core))
+            aperf_counts[core][i] = (
+                krun_get_aperf(1, core) -
+                krun_get_aperf(1, core))
```
vext01: oops!

ltratt:
Modulo the one remaining comment, these all look good to me!

ltratt:
If Sarah's happy, then please squash as you see fit, and I'll merge.

snim2:
+1 from me

vext01:
Eh failing test.

Oh, I need to bsd test.

vext01:
Fixed the tests.

GCC notices that a static function that is not used in the current compilation unit is useless, so I have made those functions disappear in travis. One function is not designed to be used on travis, but must be exposed for tests in general. Had to workaround this...

vext01:
You OK with that last change? The logic is a bit fiddly...

ltratt:
Fine with me.

vext01:
Right so just openbsd testing now.

vext01:
With the above commit, this works on OpenBSD.

Shall we start squashing this down?

ltratt:
Yes, once you've squashed, I think I can merge.

vext01:
Big splat.

PR #252, opened 2016-10-05
Fix pypy instr parser

vext01:
OK?

### DIFF
--- a/krun/vm_defs.py
+++ b/krun/vm_defs.py
@@ -558,13 +558,15 @@ class PyPyVMDef(PythonVMDef):
                 current_node = parent_stack.pop()
                 continue
 
-        # When we are done, we should be at nesting level 0 and we should see a
-        # lonely jit-summary event.
-        assert current_node[0] == "root" and len(parent_stack) == 0 and \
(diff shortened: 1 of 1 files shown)


vext01:
It asserts some things that shouldn't happen.

ltratt:
Ah, hang on, are you saying that the problem before was that an `assert` fired when it shouldn't have?

vext01:
Yep.

I had assumed that no events were emitted after the end of the last iteration, but ofcourse the iterations ruunner still has some work to do, like emit results, during which time VM event can occur.

The assertion was: "ensure that the only node left in the event tree is a jit-summary node" <-- to strong.

And is now: "Ensure there exists a JIT summary node in the remaining VM events"

ltratt:
OK, that makes sense. I suggest you squash, we wait for the tests to complete, then I'll merge.

vext01:
splat

vext01:
splat

vext01:
Don't forget this btw :)

vext01:
Sorry. Posted my comment, then the "ltratt merged" thing appeared. Ignore.

Issue #253, opened 2016-10-05
Remove non-reboot mode?


vext01:
We should remove non-reboot mode. If you don't want to actually reboot, fake reboots is enough.

vext01:
Might limit the scope for testing however...

snim2:
Discussed on #257

snim2:
This was fixed in #256

PR #256, opened 2016-10-05
Use a manifest file so as to not have results in memory during benchmarking.

vext01:
NOT EVEN CLOSE TO BEING READY :)

I suppose cookie support looks a bit like this. Please comment on the rough approach.

I've realised that anyone wishing to use --resume manually, will have to make sure the right cookie is in place.

I also need to try to detect a stale cookie. I guess it could happen.


krun/scheduler.py
```diff
@@ -149,6 +148,10 @@ def jobs_until_eta_known(self):
     def add_job(self, job):
         self.work_deque.append(job)
 
+    def clear_jobs(self):
+        del self.work_deque
+        self.work_deque = deque
```
ltratt: Reference of an undefined variable `deque`?

ltratt:
Apart from that one comment, the logic makes sense to me.

In terms of resuming, I think we just need to say "if you interrupt Krun, you need to keep the status cookie unchanged. Removing or editing it leads to undefined behaviour."

snim2:
This seems sensible to me. One possible corner case: if something goes wrong with writing the status cookie, I think Krun will think it is perpetually on the first task in the schedule and go into an infinite loop.

ltratt:
If the write fails, I think it'll raise an exception, and then stop doing anything?

snim2:
I mean if the read fails...

krun/scheduler.py
vext01: Oops. It should be `deque()`

ltratt:
I don't see how we can do anything if the cookie keeps disappearing (which means the user is either destroying it, or the hard disk is broken). IMHO, that comes strictly under the "don't do that" heading.

vext01:
I'm happy to say "leave the cookie alone" if you want resume. But I wonder if we should kill the results stripper functionality, which clearly isn't going to be easy to use anymore.

I'm starting to think that killing little-used code in Krun can only be a good thing at this point.

ltratt:
@vext01 I'm all for deleting code, but it might be worth thinking how much testing the deletion will require.

vext01:
@snim2 Any thoughts?

snim2:
I think if there is no cookie, carry on, if the read fails for some other reason we should abort.

vext01:
About deleting the result stripper I mean.

snim2:
Well, it was you that wanted that functionality :)

I think it is OK to delete it if you feel we have good test coverage.

vext01:
Ironically, I've never used that functionality.

vext01:
I wonder if there is any merit in having the cookie contain one line for each proc. exec, and then once the proc.exec is done, the line is marked as done (or removed). This means we don't have to repeatedly regenerate the schedule after each run, which means the scary code in `build_shcedule` can go.

Thoughts? @snim2

vext01:
Correct, but we never need to decide what has already been run.

vext01:
Suggest I try it out and see how we go?

snim2:
Makes sense

vext01:
It touches the ETA code, but I hope it isn't hard to fix. (famous last words?)

vext01:
Still not ready, but here is most of the code (doesn't yet work).

On first run, Krun writes a manifest file to disk that looks like this:
```
eta_avail_idx=-1
keys
O dummy:C:default-c
O nbody:C:default-c
O dummy:CPython:default-python
O nbody:CPython:default-python
O dummy:C:default-c
O nbody:C:default-c
O dummy:CPython:default-python
O nbody:CPython:default-python
```

`keys` signifies the beginning of jobs, then each line is a `flag key` pair. Flags let us know the status of the benchmarking session:

 * O - outstanding
 * C - complete
 * E - errored
 * S - skip

The idea was to never such the whole file into memory, but to simply update the flags as required.

This means:

 * We only build the schedule once, and some of the scary logic disappears.
 * The only time you should have no manifest file is when you run for the first time, so it is easy to detect when something went wrong.

Downsides:
 * The change *is* invasive.

[The reason this doesn't work ATM is OpenBSD does not support writing to an arbitrary offset in an existing file (AFAICS). `a` mode for open always appends to the end of the file. This means I can't update the flags in the manifest in the way I intended. So we may end up having to suck the file into memory after all, boo.]

In an ideal world this change is nice, and clearly superior, but I fear it is too invasive for us now. What do you think people? Maybe after all I should just revert to the old way where we build a schedule after each proc. exec.

snim2:
3 files (minus tests) seems reasonable in terms of invasiveness, although this is the tricky bit of the code. The manifest itself looks good to me.

vext01:
So I think you are telling me that this is worthwhile to continue working on.

snim2:
I am!

ltratt:
Fine with me.

I think you should be able to seek to a position and then write a limited number of bytes without issue?

vext01:
I thought I could too. Will investigate...

vext01:
```
vext01@bencher5:/tmp$ cat a.py 
import os

with open("testfile", "w") as fh:
    fh.write("123456")
os.system("hexdump -C  testfile")

with open("testfile", "a+") as fh:
    fh.seek(3)
    fh.write("X")
os.system("hexdump -C  testfile")
vext01@bencher5:/tmp$ python2.7 a.py 
00000000  31 32 33 34 35 36                                 |123456|
00000006
00000000  31 32 33 34 35 36 58                              |123456X|
00000007
```

vext01:
This works:

```
import os

with open("testfile", "w") as fh:
    fh.write("123456")
os.system("hexdump -C  testfile")

fd = os.open("testfile", os.O_WRONLY)
os.lseek(fd, 3, os.SEEK_SET)
os.write(fd, "X")
os.close(fd)
os.system("hexdump -C  testfile")
vext01@bencher5:/tmp$ python2.7 a.py 
00000000  31 32 33 34 35 36                                 |123456|
00000006
00000000  31 32 33 58 35 36                                 |123X56|
00000006
```

ltratt:
Good ;) If it hadn't worked, I think none of the database out there would be working properly on OpenBSD!

vext01:
yeah, so file operations are not flexible enough, we have to use fds.

vext01:
[This is the same on linux even]

snim2:
Fixes #253

ltratt:
OK, I'll need to be the final approver for this one, as I'm the person who hasn't been involved ;) Let me know when it's ready. It might be worth coordinating some squishing between the two of you before that point, to make the reviewing a little easier, but that's up to you both.

vext01:
Tests fixed -- @snim2 if you do your squashes now, I will review the remainder and decide how/if to squash.

snim2:
OK @vext01 your turn!

vext01:
I've squashed this down best I can.

You OK to start reviewing this @ltratt.

(Apologies this is so large)

README.md
```diff
@@ -406,8 +368,7 @@ miscompiled, you may want to re-run all benchmarks for the troublesome VM only.
 Of course, under ideal circumstances, you would collect all results in one go.
 
 You can use the `--strip-results` mode to strip results from your result file.
```
vext01: `--strip results` was removed. We should kill all mentions of it.

etc/rc.local.openbsd
```diff
@@ -26,5 +26,4 @@ PYTHON=/usr/local/bin/python2.7
 # Don't change anything below here
 
 env PATH=${PATH}:/usr/local/bin sudo -u ${INITIAL_USER} sh -c \
-    "cd ${EXPERIMENT_DIR} && ${PYTHON} ${KRUN_DIR}/krun.py \
-   --resume --reboot --started-by-init ${CONFIG_FILE}" &
+    "cd ${EXPERIMENT_DIR} && ${PYTHON} ${KRUN_DIR}/krun.py --hardware-reboots ${CONFIG_FILE}" &
```
vext01: We need to update the example `rc.local` files, but shall we do that as a separate PR?

examples/example.krun
```diff
@@ -45,7 +45,7 @@ VMS = {
 
 BENCHMARKS = {
     'dummy': 1000,
-    'nbody': 1000,
+    #'nbody': 1000,
```
ltratt: Let's delete this line.

krun.py
```diff
@@ -192,11 +177,10 @@ def main(parser):
         util.print_session_info(config)
         return
 
-    if args.strip_results:
-        util.strip_results(config, args.strip_results)
-        return
+    on_first_invokation = not (os.path.isfile(ManifestManager.PATH) and
```
vext01: Did we somwhere check that if a manifest file exists, so should a results file, and the other way around?

krun/scheduler.py
```diff
@@ -37,6 +36,132 @@ class ScheduleEmpty(Exception):
...
+        offset = 0
+
+        # Parse manifest header
+        for line in fh:
+            offset += len(line)
+            line = line.strip()
+            if line == "keys":
+                break
+            else:
+                key, val = line.split("=")
```
ltratt: This would be clearer and shorter as:
```python
assert key == "eta_avail_idx"
self.eta_avail_idx = int(val)
```

krun.py
```diff
@@ -275,8 +252,8 @@ def inner_main(mailer, config, args):
         platform.starting_temperatures = current.starting_temperatures
     else:
         # Touch the config file to update its mtime. This is required
-        # by resume-mode which uses the mtime to determine the name of
-        # the log file, should this benchmark be resumed.
+        # by when resuming a partially complete benchmark, in which case Krun
```
vext01: A partially complete benchmark *session*, right?

krun/platform.py
```diff
@@ -654,8 +654,16 @@ def __init__(self, mailer, config):
         self.temp_sensor_map = None
         UnixLikePlatform.__init__(self, mailer, config)
         self.num_cpus = self._get_num_cpus()
+        self.num_per_core_measurements = self._get_num_per_core_measurements()
+
         self.virt_what_cmd = None  # set later
 
+    def _get_num_per_core_measurements(self):
+        # For all systems apart from travis we expect per-core measurements
+        if os.environ.get("TRAVIS") == "true":
```
vext01: Note -- if this kind of thing happens much more (I hope not), we should make a `class TravisPlatform(DebianLinuxPlatform)`. Not now though.

krun/results.py
```diff
@@ -93,10 +93,41 @@ def read_from_file(self, results_file):
...
+            # Check the length of the in-process interations in results
+            # XXX Check all core counts are the same
+            # XXX Check all exec counts are the same
+            expect_len = None
+            for exec_idx in xrange(len(execs)):
+                one_exec = execs[exec_idx]
+                if expect_len is None:
+                    expect_len = len(one_exec)
+
+                #if not all([len(x) == expect_len for x in self.core_cycle_counts[key][exec_idx]]):
```
ltratt: Delete line.

krun/results.py
```diff
@@ -93,10 +93,41 @@ def read_from_file(self, results_file):
...
+                for core in self.core_cycle_counts[key][exec_idx]:
+                    if len(core) != expect_len:
+                        fatal("inconsistent results: core_cycle_counts length")
+
```
ltratt: Extra blank line.

krun/scheduler.py
```diff
@@ -279,160 +314,178 @@ def _make_pre_post_cmd_env(self):
...
-
-            if self.reboot and len(self) > 0:
-                info("Reboot in preparation for next execution")
-                self._reboot()
+        if self.on_first_invokation:
+            info("Reboot prior to first execution")
+            self._reboot()
 
-        self.platform.save_power()
+        # If we get here, this is a real run, with a benchmark about to
```
vext01: I'm really happy with the simplification of `run()`, it's much easier to understand now.

krun/scheduler.py
```diff
@@ -350,131 +346,134 @@ def run(self):
...
-            else:
-                assert False
+            info("Reboot in preparation for next execution")
+            self._reboot(self.manifest.total_num_execs)
 
-            if self.reboot and self.manifest.num_execs_left > 0:
-                info("Reboot in preparation for next execution")
-                self._reboot(self.manifest.total_num_execs)
+        elif self.manifest.num_execs_left < 0:
+            assert False
```
ltratt: This is is not a good `assert False` -- it should be somewhere else (e.g. before the `if`).

krun/scheduler.py
```diff
@@ -350,131 +346,134 @@ def run(self):
...
-        self.platform.save_power()
+        elif self.manifest.num_execs_left ==  0:
+            self.platform.save_power()
 
-        info("Done: Results dumped to %s" % self.config.results_filename())
-        err_msg = "Errors/warnings occurred -- read the log!"
-        if self.results.error_flag:
-            warn(err_msg)
+            info("Done: Results dumped to %s" % self.config.results_filename())
+            err_msg = "Errors/warnings occurred -- read the log!"
```
ltratt: This doesn't make sense to me. a) if There are no execs why would we possibly raise a "error occured" message? b) If `error_flag` isn't set, why construct `err_msg`?

krun/tests/test_scheduler.py
```diff
@@ -46,13 +43,12 @@ def dummy_reboot(self, _):
 
     reboots = 0
     while True:
-        resume = False
+        on_first_invokation = True
```
ltratt: "invocation"

snim2: oops

krun.py
```diff
@@ -206,36 +190,29 @@ def main(parser):
     mailer = Mailer(mail_recipients, max_mails=config.MAX_MAILS)
 
     try:
-        inner_main(mailer, config, args)
+        inner_main(mailer, on_first_invokation, config, args)
     except util.FatalKrunError as e:
         subject = "Fatal Krun Exception"
         mailer.send(subject, e.args[0], bypass_limiter=True)
         util.run_shell_cmd_list(config.POST_EXECUTION_CMDS)
```
ltratt: I think this should be:
```python
     except util.FatalKrunError as e:
          util.run_shell_cmd_list(config.POST_EXECUTION_CMDS)
          subject = "Fatal Krun Exception"
          mailer.send(subject, e.args[0], bypass_limiter=True)
```

ltratt: I also wonder if `run_shell_cmd_list` should have an argument `try_all` (or similar) which, even if a command fails, tries to run the subsequent commands?

ltratt:
OK, those are my initial comments.

examples/example.krun
vext01: I'm confused -- this line is not commented on the head of the branch... GH bug?

etc/rc.local.openbsd
vext01: #264

krun/results.py
vext01: fixed in a later commit.

examples/example.krun
vext01: fixed in a later commit.

krun/results.py
vext01: fixed in a later commit.

krun/scheduler.py
vext01: If there was an error at any point in the session a benchmark has failed, a message is printed right at the end. This is mostly just useful for when you are testing, as in a real benchmarking session you would have received an email.

ltratt: OK, that addresses point a) (but not b)).

krun/tests/test_scheduler.py
ltratt: Das ist England ('16).

krun.py
vext01: Up until now, we have allowed commands to fail by having `| true` in the config. Is that enough?

vext01: As for the re-order, let's do that in the pre/post hook fix branch.

ltratt: Ah, yes, the `| true` hack is fine. Good point.

ltratt: I don't think we should leave the reordering though -- it's a one line fix, and we'll forget it otherwise.

vext01:
I think that covers it. One remaining comment outstanding for @snim2 regarding the consistency of the manifest and results file existing (or not).

ltratt:
I'd like to reorder the `try/except` thing before we merge. Once that's done, I'm fine with this, and it can be squashed as needed.

[Note that tests are failing ATM though.]

ltratt:
OK, once the tests have passed, please squash, then I can merge.

vext01:
Will wait for Sarah before any squashing.

snim2:
@vext01 I'm going to answer the question here. We have a number of checks in `krun.py` that deal with all of this, but they aren't all next to each other. Let me pull them out here so you can see:

```python
    on_first_invocation = not (os.path.isfile(ManifestManager.PATH) and
                               os.stat(ManifestManager.PATH).st_size > 0)

    ...

    out_file = config.results_filename()  # This is the results file
    out_file_exists = os.path.exists(out_file)

    if out_file_exists and not os.path.isfile(out_file):
        util.fatal(
            "Output file '%s' exists but is not a regular file" % out_file)

    if out_file_exists and on_first_invocation:
        util.fatal("Output results file '%s' already exists. "
                   "Move the file away before running Krun." % out_file)

    if not out_file_exists and not on_first_invocation:
        util.fatal("No results file to resume. Expected '%s'" % out_file)

    ...

    if not on_first_invocation:
        # output file must exist, due to check above
        assert(out_file_exists)
        current = Results(config, platform, results_file=out_file)
```

vext01:
Alright -- will squash now. Thanks

vext01:
I think that's the best squashing I can do without pain.

ltratt:
We can definitely squash the "in an error case" commit away easily. The others are OK as-is though.

vext01:
I kept it separate as it's not related to the manifest changes.

ltratt:
Point taken.

PR #257, opened 2016-10-06
Remove duplicate run of post-command hook.

### LINKED ISSUE #248: Fix bugs in PRE/POST hooks
* If a post hook command fails, this triggers a second run of the post hooks.
 * When sanity checks run,  post hooks are run, but not pre hooks -- suggest running neither.

snim2:
Not yet ready for review.

Fixes #248

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -6,9 +6,8 @@ Benchmark, running many fresh processes.
 usage: runner.py <config_file.krun>
 """
 
-import argparse, logging, os, sys
+import argparse, locale, logging, os, sys, traceback
 from logging import debug, info, warn
-import locale
 
 import krun.util as util
 from krun.config import Config
@@ -191,10 +190,14 @@ def main(parser):
 
     try:
         inner_main(mailer, on_first_invocation, config, args)
-    except util.FatalKrunError as e:
-        util.run_shell_cmd_list(config.POST_EXECUTION_CMDS)
+    except Exception:
+        error_info = sys.exc_info()
         subject = "Fatal Krun Exception"
-        mailer.send(subject, e.args[0], bypass_limiter=True)
+        lines = ["Fatal Krun error: %s\n" % str(error_info[1])]
+        for frame in traceback.format_tb(error_info[2]):
+            lines.append(frame)
+        msg = "".join(lines)
+        util.log_and_mail(mailer, debug, subject, msg, bypass_limiter=False)
         raise e
 
 
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -9,6 +9,7 @@ import os, subprocess, sys, time
 # Wait this many seconds for the init system to finish bringing up services.
 STARTUP_WAIT_SECONDS = 2 * 60
 
+
 def mean(seq):
     if len(seq) == 0:
         raise ValueError("Cannot calculate mean of empty sequence.")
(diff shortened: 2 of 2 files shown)


krun.py
```diff
@@ -210,7 +210,6 @@ def main(parser):
     except util.FatalKrunError as e:
         subject = "Fatal Krun Exception"
         mailer.send(subject, e.args[0], bypass_limiter=True)
-        util.run_shell_cmd_list(config.POST_EXECUTION_CMDS)
```
snim2: This is the source of the duplicate call when a post-command fails, and `util.run_shell_command` calls `fatal()`. So, do you want to introduce a new exception type? `JustAFleshWoundKrunError`?

ltratt:
I wonder if we're being too clever for our own good here. I think the simplest solution might in fact be the right one. Here are the rules I think we have to observe:
 * If the pre commands are run, the post commands *must* be run
 * If *any* of the pre commands are run, the post commands *must* be run
I don't think we've articulated that second point before, but it now seems obvious to me. If, for example, an early pre command brings the network down, and then a later pre command fails, we still want the post commands to be executed (to bring the network back up).

On that basis, I presume that all we need is:
```python
try:
  exec_pre_cmds()
  <everything else>
finally:
  exec_post_cmds()
```

vext01:
In addition, the pre-and post commands need to happen before and after each process execution even if there is no error. As it stands, in reboot mode, the finally would not be reached if no error occurred because Krun reboots the system.

ltratt:
I thought it raised an exception to indicate that it wanted to reboot? If not... can we make it raise an exception :)

Then it can be something like:

```python
reboot = False
try:
  exec_pre_cmds()
  <everything else>
except RebootException:
  reboot = True
finally:
  exec_post_cmds()

if reboot:
  self.do_reboot()
```

vext01:
That would work for reboot mode, but not for no-reboot mode.

In another PR, I suggested removed no-reboot mode might not be a bad idea.

ltratt:
OK, so let's deprecate no-reboot mode? Then this PR could simply put in an error saying "no reboot mode no longer supported" and a separate PR can actually do the ripping out?

vext01:
Sounds good.

Makes sense @snim2?

vext01:
If there is one thing I have learned from Krun, it is, don't add so many knobs and whistles that it becomes hard to test. This is what I inadvertently did with Krun.

ltratt:
Amen to simplicity.

snim2:
I think it might be simpler to do both things in one PR, TBH

ltratt:
I'm not a big of big PRs. The ripping-out PR will quite probably obscure the (very important) logic this PR needs to concentrate on.

vext01:
Let's also add an exception handler for *all* other kinds of exception, so that we will get an email if (e.g.) and assertion fires.

snim2:
I am going to bring this up to date with master and force push, since we have made substantial changes.

ltratt:
OK.

vext01:
So where do we stand with this?

ltratt:
I don't think anything's changed relative to my suggestions above?

snim2:
I've rebased and I'll push when I have some new commits.

snim2:
I've had a first go at this. Note that if a pre- or post-execution command fails, Krun now stops. I'm not sure if that is what was intended in the discussion above?

krun/scheduler.py
```diff
@@ -345,56 +345,72 @@ def run(self):
...
-        )
+        # Run the pre-exec commands, the benchmark and the post-exec commands.
+        # These are wrapped in a try/except, so that the post-exec commands
+        # are always executed, even if an exception has occurred. We only
+        # reboot /after/ the post-exec commands have completed.
+        try:
+            # Run the user's pre-process-execution commands We can't put an ETA
+            # estimate in the environment for the pre-commands as we have not
+            # (and should not) load the results file into memory yet.
+            util.run_shell_cmd_list(
```
ltratt: Should be on one line.

krun/scheduler.py
```diff
@@ -345,56 +345,72 @@ def run(self):
...
+        except Exception:
+            error_info = sys.exc_info()
+            lines = ["Krun error during benchmark execution: %s\n" % str(error_info[1])]
+            for frame in traceback.format_tb(error_info[2]):
+                lines.append(frame)
+            msg = "".join(lines)
+            util.log_and_mail(self.mailer, debug,
+                              "Krun error during benchmark execution",
+                              msg, bypass_limiter=False)
+
```
ltratt: Add `sys.exit(1)` here?

krun.py
```diff
@@ -192,7 +192,6 @@ def main(parser):
     try:
         inner_main(mailer, on_first_invocation, config, args)
     except util.FatalKrunError as e:
```
vext01: Should this be a generic except?

vext01: (I'm not sure)

krun/scheduler.py
vext01: extra line?

krun/scheduler.py
```diff
@@ -345,56 +345,72 @@ def run(self):
...
+            if self.platform.hardware_reboots:
+                # Add time taken to wait for system to come up if we are in
+                # hardware-reboot mode.
+                eta_info += STARTUP_WAIT_SECONDS
+            self.add_eta_info(job.key, eta_info)
+
+            # We dump the json after each process exec so we can monitor the
+            # JSON file mid-run. It is overwritten each time.
+            self.manifest.update(flag)
+
```
snim2: I was following the structure of your code here, but can lose a few lines

krun/scheduler.py
snim2: Surely the post-exec commands in the `finally` won't happen if we `sys.exit(1)` here?

ltratt: I think `sys.exit` itself raises an exception? [It's been a while, so I don't remember fully.]

ltratt:
I'm fine with just reraising, so that's fine. Can we lose the blank lines before the `except` and `finally` please?

snim2:
Ready to squash?

ltratt:
Let's rebase against master, squash, and then I can merge.

PR #263, opened 2016-10-13
Rmsr autoload

### LINKED ISSUE #260: rmsr autoload and documentation
rmsr is needed for krun on linux. And the user running tests must be in the root group to access the device nodes.

snim2:
Ready for review.

Fixes #260

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ PASS_DOWN_ARGS =	ENABLE_JAVA=${ENABLE_JAVA} JAVAC=${JAVAC} \
 
 .PHONY: libkrun vm-sanity-checks clean all
 
-all: iterations-runners libkrun vm-sanity-checks platform-sanity-checks
+all: iterations-runners libkrun vm-sanity-checks platform-sanity-checks rmsr/rmsr.ko
 
 iterations-runners: libkrun
 	cd iterations_runners && ${MAKE} ${PASS_DOWN_ARGS}
@@ -22,8 +22,17 @@ vm-sanity-checks:
 platform-sanity-checks:
 	cd platform_sanity_checks && ${MAKE} ${PASS_DOWN_ARGS}
 
+rmsr/rmsr.ko:
+ifeq ($(shell uname -s),Linux)
+ifneq ("${TRAVIS}","true")
+	cd rmsr && ${MAKE}
+endif
+endif
+
+
 clean:
 	cd iterations_runners && ${MAKE} clean
 	cd libkrun && ${MAKE} clean
 	cd vm_sanity_checks && ${MAKE} clean
 	cd platform_sanity_checks && ${MAKE} clean
+	cd rmsr && ${MAKE} clean
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -22,6 +22,7 @@ from krun.vm_defs import BENCHMARK_USER
 NICE_PRIORITY = -20
 DIR = os.path.abspath(os.path.dirname(__file__))
 LIBKRUNTIME_DIR = os.path.join(DIR, "..", "libkrun")
+RMSR_DIR = os.path.join(DIR, "..", "rmsr")
 SYNC_SLEEP_SECS = 30  # time to wait for sync() to finish
 
 
@@ -764,6 +765,7 @@ class LinuxPlatform(UnixLikePlatform):
         self._check_cpu_governor()
         self._check_cpu_scaler()
         self._check_perf_samplerate()
+        self._check_rmsr_loaded()
         if not self.no_tickless_check:
             self._check_tickless_kernel()
         self._check_aslr_disabled()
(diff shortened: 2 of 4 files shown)


krun/platform.py
```diff
@@ -1006,6 +1008,26 @@ def _check_cpu_scaler(self):
...
+            debug("Running on Travis-CI so not calling insmod.")
+            return
+        if not self.config.AUTOLOAD_RMSR:
+            fatal("rmsr module not loaded and AUTOLOAD_RMSR is False in config.")
+        # rmsr git repo should have been cloned and compiled by 'make all'.
+        debug("Auto-loading 'rmsr' module.")
+        cmd = "%s rmmod msr" % self.change_user_cmd
+        stdout, stderr, rc = run_shell_cmd(cmd, failure_fatal=False)
+        module = os.path.join(RMSR_DIR, "rmsr.ko")
+        cmd = "%s insmod %s" % (self.change_user_cmd, module)
```
snim2: Lines 1024-5 do this, I think

examples/example.krun
```diff
@@ -73,3 +73,6 @@ TEMP_READ_PAUSE = 1
 
 # CPU pinning (off by default)
 #ENABLE_PINNING = False
+
+# Autoload rmsr kernel module (on by defualt)
+# AUTOLOAD_RMSR = True
```
vext01: Hrm, I was imagining a config var to specify the location of rmsr. They way you locate rmsr is fine (assume the krun build system built it) is absolutel fine, but now I wonder if there would ever be a situation where `AUTOLOAD_RMSR = False` would be useful. Is there? Or should we kill the config variable altogether?

krun/platform.py
vext01: Ah yes.

examples/example.krun
snim2: I don't have any good intuition on this; might it be possible to be running Krun on a machine that doesn't support msr, or a kernel that doesn't?

ltratt: Good point. At the moment, turning it off on Linux would just cause later things to go wrong. So we can remove the knob for the time being.

ltratt: On Linux, it currently makes no sense not to load rsmr (as other code depends on it), so I suggest removing the knob. Less knobs is good!

snim2:
OK, are we good to squash now?

Makefile
```diff
@@ -1,4 +1,5 @@
 JAVAC ?= javac
+GIT ?= git
```
ltratt: There's no real use to git being a variable (the chances of us wanting e.g. 2 versions of it are vanishingly small), so let's simplify and remove it.

Makefile
```diff
@@ -22,8 +23,19 @@ vm-sanity-checks:
 platform-sanity-checks:
 	cd platform_sanity_checks && ${MAKE} ${PASS_DOWN_ARGS}
 
+rmsr/rmsr.ko:
+ifeq ($(shell uname -s),Linux)
+	# Allow git to fail if rmsr has already been cloned.
+	${GIT} clone https://github.com/softdevteam/rmsr.git | true
```
ltratt: We can't do this: it will also fail if you run `git` and e.g. there isn't a network connection.

Doing this stuff in a makefile is always a mess, because the syntax is so awful for it. It would be easier to put this into build.sh, and then we have easier access to shell commands, and we can say things like "if the rsmr directory doesn't exist, clone it".

snim2: There isn't a `build.sh` in Krun. We can push this out to a separate shell script (in which case it still needs to be run from the Makefile) or make the `rmsr` repository a submodule of Krun.

I can't think of any other options...

vext01: This target is only used in the build stage, so I think this is OK.

ltratt: Good point. Let's have the warmup experiment build this then. Doing this stuff in a makefile is a recipe for unreadable, unmaintainable, and incorrect code. [I speak from bitter experience.]

vext01: I disagree. rmsr is a dependency of krun, not of the warmup experiment. If we want rmsr to be external, then we can re-instate a config variable that allows the user to  specify where the module is, then build.sh in the warmup_experiment can build it, but I don't really see any need.

ltratt: OK, so heretic thought: let's integrate the rsmr repo into this one? I mean, it could be a subdir, right?

ltratt: [I *really* don't want `git clone` in a makefile... we will get it wrong]

vext01: I suppose the correct way would be to to use a git submodule.

vext01: I don't mind looking into that, as I really do need to learn how submodules work ;)

ltratt: Good idea, we can try that. I don't really have any experience of submodules, but it feels like it might be the right way to go.

snim2: I'll do it. You can read about submodules here: https://github.com/blog/2104-working-with-submodules

vext01: OK, I will try that out and push the result!

vext01:
Or Sarah could beat me to it ;)

vext01:
Note that you will need to clone with `--recursive` or, if you already have a clone, do `git submodule init`.

We should make warmup_experiment clone recursively.

ltratt:
So long as we update `build.sh` in the warmup_experiment repo to do a recursive clone, all should be good, I think.

vext01:
Wait, I have a commit to go in.

snim2:
`make all` works on my laptop

vext01:
Works on bencher5.

snim2:
Tests are passing now (although we have to wait for the test from the last commit). Are we ready to squash?

PR #268, opened 2016-10-17
Separate json files for instrumentation data

vext01:
Don't merge yet please.

How about something like this?

If it looks good, I want to add a couple of tests to check I am computing the next exec index correctly from the manifest file.

```
$ python -mpdb ./krun/krun.py test.krun --quick --debug debug
...
[2016-10-17 11:40:58 INFO] Done: Results dumped to test_results.json.bz2
[2016-10-17 11:40:58 INFO] Session completed. Log file at: 'test_20161017_114049.log'
Don't forget to disable Krun at boot.
$ ls test_
test_20161017_114049.log  test_instr_data/          test_results.json.bz2     
vext01@bencher5:~/research/warmup_experiment$ ls test_instr_data/
empty_PyPy_default-python__0.json.bz2  empty_PyPy_default-python__5.json.bz2
empty_PyPy_default-python__1.json.bz2  empty_PyPy_default-python__6.json.bz2
empty_PyPy_default-python__2.json.bz2  empty_PyPy_default-python__7.json.bz2
empty_PyPy_default-python__3.json.bz2  empty_PyPy_default-python__8.json.bz2
empty_PyPy_default-python__4.json.bz2  empty_PyPy_default-python__9.json.bz2
```

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -202,6 +202,9 @@ def inner_main(mailer, on_first_invocation, config, args):
     out_file = config.results_filename()
     out_file_exists = os.path.exists(out_file)
 
+    instr_dir = util.get_instr_json_dir(config)
+    instr_dir_exists = os.path.exists(instr_dir)
+
     if out_file_exists and not os.path.isfile(out_file):
         util.fatal(
             "Output file '%s' exists but is not a regular file" % out_file)
@@ -209,6 +212,10 @@ def inner_main(mailer, on_first_invocation, config, args):
     if out_file_exists and on_first_invocation:
         util.fatal("Output results file '%s' already exists. "
                    "Move the file away before running Krun." % out_file)
+
+    if instr_dir_exists and on_first_invocation:
+        util.fatal("Instrumentation dir '%s' exists (from an older run?)" %
+                   (instr_dir))
 
     if not out_file_exists and not on_first_invocation:
         util.fatal("No results file to resume. Expected '%s'" % out_file)
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -9,13 +9,6 @@ import os, subprocess, sys, time
 # Wait this many seconds for the init system to finish bringing up services.
 STARTUP_WAIT_SECONDS = 2 * 60
 
-EMPTY_MEASUREMENTS = {
-    "wallclock_times": [],
-    "core_cycle_counts": [],
-    "aperf_counts": [],
-    "mperf_counts": [],
-}
-
 def mean(seq):
     if len(seq) == 0:
         raise ValueError("Cannot calculate mean of empty sequence.")
@@ -41,6 +34,7 @@ class ManifestManager(object):
         self.total_num_execs = 0
         self.eta_avail_idx = 0
         self.outstanding_exec_counts = {}
+        self.completed_exec_counts = {}  # including errors
         self.skipped_keys = set()
         self.non_skipped_keys = set()
 
@@ -83,6 +77,8 @@ class ManifestManager(object):
             flag, key = line.strip().split(" ")
             if key not in self.outstanding_exec_counts:
                 self.outstanding_exec_counts[key] = 0
+            if not key in self.completed_exec_counts:
+                self.completed_exec_counts[key] = 0
 
             if flag in ["S", "E", "C"]:  # skip, error, completed
                 pass
@@ -92,6 +88,7 @@ class ManifestManager(object):
                     self.next_exec_key = key
                     self.next_exec_flag_offset = offset
                     self.next_exec_idx = exec_idx
+
                 self.num_execs_left += 1
             else:
                 assert False  # bogus flag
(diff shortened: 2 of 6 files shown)


snim2:
This looks sensible to me!

vext01:
There will need to be changes to plotting scripts too. Is that OK @snim2?

snim2:
Yes, so long as I have a clear idea of where all the data is I can raise an issue for it.

vext01:
If you have a config file named `test.krun`, you get a dir called `test_instr_data` with files prefixed with the job key (with `:` replaced with `__`) and then a process exec number suffix + `.json.bz2`.

See above for examples.

snim2:
OK, raised in warmup issue 201.

vext01:
Here are some more robust tests.

OK to squash?

snim2:
yup!

vext01:
splat

vext01:
Let's merge this?

snim2:
@ltratt happy?

krun/scheduler.py
```diff
@@ -189,6 +182,15 @@ def __init__(self, sched, vm_name, vm_info, benchmark, variant, parameter):
         # Used in results JSON and ETA dict
         self.key = "%s:%s:%s" % (self.benchmark, self.vm_name, self.variant)
 
+        # When a benchmark fails, you get this:
+        per_core = [[]] * sched.platform.num_per_core_measurements
```
ltratt: This isn't safe:

```python
>>> a = [[]] * 2
>>> a[0].append(10)
>>> a
[[10], [10]]
```
Even if, in this case, the sub-lists aren't used, the idiom is a classic shoot-in-the-footer. Best to do something like:
```python
per_core = []
for _ in range(sched.platform.num_per_core_measurements: per_core.append([])
```

krun/scheduler.py
```diff
@@ -228,17 +230,22 @@ def run(self, mailer, dry_run=False):
...
             if vm_def.instrument:
-                instr_data = vm_def.get_instr_data()
-                for k, v in instr_data.iteritems():
-                    assert len(instr_data[k]) == in_proc_iters
+                if flag != "E":
+                    instr_data = vm_def.get_instr_data()
+                    for k, v in instr_data.iteritems():
+                        assert len(instr_data[k]) == in_proc_iters
+                else:
+                    # Although there may be some instrumentation data, it is
```
ltratt: I think this comment might be clearer as "As an error happened, we have to assume that any instrumentation data is compromised; we thus ignore it."

krun/util.py
```diff
@@ -332,3 +333,23 @@ def get_git_version():
...
+def dump_instr_json(key, exec_num, config, instr_data):
+    """Write per-execution instrumentation data to a separate JSON file"""
+
+    instr_json_dir = get_instr_json_dir(config)
+    if not os.path.exists(instr_json_dir):
+        os.mkdir(instr_json_dir)
+
+    filename = "%s__%s.json.bz2" % (key.replace(":", "_"), exec_num)
+    path = os.path.join(instr_json_dir, filename)
+    assert not os.path.exists(path)
```
ltratt: Presumably somewhere before we've checked that the directory doesn't exist? Otherwise, won't we run one whole benchmark, discover there's already a file of that name and then die (with a not-very-helpful message)?

krun/scheduler.py
vext01: You are right that the lists are never mutated, but will refactor anyway.

krun/util.py
ltratt: OK, can we add a comment to that effect? Something like "Although we've tested earlier in function `blahblah` that the directory containing these files didn't exist, this assert is designed to stop the user doing something really stupid." [Well, that's not a great comment, but you get the idea.]

vext01: There seems to be confusion. We are asserting that the instr *file* (not dir) does not exist. We check the directory does not exist when the benchmark session is started.

krun.py
```diff
@@ -210,6 +213,10 @@ def inner_main(mailer, on_first_invocation, config, args):
         util.fatal("Output results file '%s' already exists. "
                    "Move the file away before running Krun." % out_file)
 
+    if instr_dir_exists and on_first_invocation:
```
vext01: @ltratt here is the dir test.

krun/util.py
ltratt: Yes, I understand :) I'm just saying add a comment explaining *where* that test happens, otherwise future readers will have the same initial comment I did.

vext01: Try that

ltratt:
I'm happy (and assuming Sarah's not been made unhappy by any of the changes!). If you squash and rebase this, I can then merge.

snim2:
lgtm

vext01:
splat

ltratt:
Thanks. Will merge once tests have finished.

vext01:
This is good to go @ltratt

PR #273, opened 2016-10-19
More robust exn handler

vext01:
VM testing has raised a couple of other issues.

See individual commit messages.

OK?

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -372,9 +372,14 @@ class ExecutionScheduler(object):
         # are always executed, even if an exception has occurred. We only
         # reboot /after/ the post-exec commands have completed.
         try:
-            # Run the user's pre-process-execution commands We can't put an ETA
-            # estimate in the environment for the pre-commands as we have not
-            # (and should not) load the results file into memory yet.
+            # Run the user's pre-process-execution commands. We can't put an
+            # ETA estimate in the environment for the pre-commands as we have
+            # not (and should not) load the results file into memory yet.
+            #
+            # It might seem tempting to move this outside the try block, to
+            # ensure that post-hooks are only run if pre-hooks ran. We don't,
+            # thus avoiding the case where only *part* of the pre-hooks run,
+            # but the post-hooks then don't run.
             util.run_shell_cmd_list(self.config.PRE_EXECUTION_CMDS,)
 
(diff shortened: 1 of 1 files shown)


krun/scheduler.py
```diff
@@ -367,16 +367,16 @@ def run(self):
         job = ExecutionJob(self, vm, self.config.VMS[vm], bench, variant,
                            self.config.BENCHMARKS[bench])
 
+        # Run the user's pre-process-execution commands We can't put an ETA
```
ltratt: Missing "." before "we".

ltratt:
The first commit opens up a nasty race where *part* of the pre hooks fail, but then none of the post hooks are run. For example, the network might be brought down, then something else goes wrong in the pre hooks; with this commit, the post hook networking-bringer-uper won't be run. This was something I was very conscious of when I suggested the layout of the try/finally, so I'd need to be convinced that it should be changed. I do think that we should explicitly document that the post hooks are run even if some of the pre hooks failed though.

snim2:
The second commit looks sensible to me, the first needs some discussion.

vext01:
The network down command is last in the pre-hook, so we are grand. Right?

vext01:
It's not just about the network. Point taken.

snim2:
I think if we are careful about the network and `| true` in the right places then this looks OK. Let's wait for @ltratt before we squash though.

vext01:
@snim2 laurie is right.

Suppose we have pre-hooks:
```
rcctl stop postfix
rcctl stop cron
```

and post hooks:
```
rcctl start postfix
rcctl start cron
```

And `rcctl stop cron` fails. Postfix would not be brought up.

My last push reverts the old behaviour and explains the placement in a comment.

vext01:
If it's OK, I would like to revise some orthogonal comments in nearby code. Can I add this change here?

snim2:
Sure. BTW the comment about dumping JSON on lines 412-413 of `scheduler.py` does not seem to apply to the line of code beneath it.

vext01:
That's the very line of code I want to change ;)

vext01:
Here we go. OK to squash?

snim2:
OK, lgtm

vext01:
splat

PR #274, opened 2016-10-19
Ignore desmg lines relating to bringing the network down/up on Linux.

vext01:
Tested with real reboots.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -1162,6 +1162,16 @@ class LinuxPlatform(UnixLikePlatform):
                 "isolcpus", "isolcpus should not be in the kernel command line"
             )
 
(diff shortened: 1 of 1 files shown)


snim2:
I can't see where that new method is called...

vext01:
```
    def check_dmesg_for_changes(self):                                          
        new_dmesg = self._collect_dmesg_lines()                                 
        patterns = self.get_allowed_dmesg_patterns()                            
```

I'm overriding a more generic implementation.

snim2:
OK, I see it! This seems OK to me. Do you want @ltratt to review also?

vext01:
This one is pretty trivial, so we can go ahead I think. The other one, I would like his input on.

PR #275, opened 2016-10-20
Log the environment

vext01:
Not ready to merge.

Laurie asked last week if we could log the environment of each process execution, so that we could check that each benchmark with the same key (so that's vm + benchmark pairings) ran with the same environment.

This PR makes a start on that. I'm using a line of shell script, as it's the innermost place before the VM itself is invoked.

Questions:
 * Should this only be done when the instrument flag is on?
 * This change introduces I/O just before the benchmark. Should we move the sync wait into the wrapper script (after the new I/O) in the form of  `sync && sleep N`?

If the general approach is OK, then I will need to do some magic with the log file locations. I couldn't write into the experiment dir, as the krun user does not have access. At the moment, I write to /tmp. Likley a good idea is to write first to /tmp, then have krun move it after the execution finishes.

FWIW, I don't see any environment discrepancies using just the example config file. A typical env looks like this:
```
SUDO_GID=0                                                                      
MAIL=/var/mail/krun                                                             
USER=krun                                                                       
HOME=/home/krun                                                                 
SUDO_UID=0                                                                      
LOGNAME=krun                                                                    
TERM=unknown                                                                    
USERNAME=krun                                                                   
PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/X11R6/bin:/usr/local/bin                
SUDO_COMMAND=/usr/local/bin/dash /tmp/krun_wrapper.dash                         
SHELL=/bin/ksh                                                                  
SUDO_USER=root                                                                  
PWD=/home/edd/krun/examples                                                     
```

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -208,6 +208,9 @@ def inner_main(mailer, on_first_invocation, config, args):
     instr_dir = util.get_instr_json_dir(config)
     instr_dir_exists = os.path.exists(instr_dir)
 
+    envlog_dir = util.get_envlog_dir(config)
+    envlog_dir_exists = os.path.exists(envlog_dir)
+
     if out_file_exists and not os.path.isfile(out_file):
         util.fatal(
             "Output file '%s' exists but is not a regular file" % out_file)
@@ -217,8 +220,10 @@ def inner_main(mailer, on_first_invocation, config, args):
                    "Move the file away before running Krun." % out_file)
 
     if instr_dir_exists and on_first_invocation:
-        util.fatal("Instrumentation dir '%s' exists (from an older run?)" %
-                   (instr_dir))
+        util.fatal("Instrumentation dir '%s' exists." % instr_dir)
+
+    if envlog_dir_exists and on_first_invocation:
+        util.fatal("Env log dir '%s' exists." % envlog_dir)
 
     if not out_file_exists and not on_first_invocation:
         util.fatal("No results file to resume. Expected '%s'" % out_file)
--- a/krun/vm_defs.py
+++ b/krun/vm_defs.py
@@ -5,6 +5,11 @@ import fnmatch
 import re
 import json
 import getpass
+import tempfile
+import pwd
+import shutil
+import util
+import stat
 from abc import ABCMeta, abstractmethod
 
 from logging import info, debug, warn
@@ -34,6 +39,9 @@ if DASH is None:
 
 class BaseVMDef(object):
     __metaclass__ = ABCMeta
+
+    # Read/write for user and group
+    ENVLOG_MODE = stat.S_IRUSR | stat.S_IRGRP | stat.S_IWUSR | stat.S_IWGRP
 
     def __init__(self, iterations_runner, env=None, instrument=False):
         self.iterations_runner = iterations_runner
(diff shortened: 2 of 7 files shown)


ltratt:
OK, this looks like a good start.

We could have the parent process pass a first argument to the shell script which is temporary directory, so that we don't have any race conditions? Then we can log all of them easily.

I have two thoughts. Either we move the sync and wait into this shell script *or* this script simply commits the environment to an internal buffer, runs the benchmarks, and *then* writes the file out. That latter approach might be easier?

vext01:
> We could have the parent process pass a first argument to the shell script which is temporary directory, so that we don't have any race conditions? Then we can log all of them easily.

Yep, could do, although I would probably suggest we use the PID of the krun process, as it's easier to implement. Note alsp that some other areas of krun (e.g. the wrapper script filename itself) are guilty of not using unique filenames. I don't think it is "racy" per se though, since you are not supposed to run more than one krun instance anyway. Any thoughts?

> I have two thoughts. Either we move the sync and wait into this shell script or this script simply commits the environment to an internal buffer, runs the benchmarks, and then writes the file out. That latter approach might be easier?

Yes, the latter is easier.

Something like:

```
ENVLOG=`env`
<blah>
echo ${ENVLOG} > file
```

Right?

vext01:
>We could have the parent process pass a first argument to the shell script which is temporary directory, so that we don't have any race conditions? Then we can log all of them easily.

Ah wait. Depending what you mean with this, we might be ok. Note that the env file is derived from the job key and is suffixed with the execution number.

vext01:
Try this.

Afterwards we have have directory (whose name is derived from the config file name) containing env logs, one per exec.

E.g. after running `example.krun`:

```
$ ls example_envlogs/
dummy__CPython__default-python__0.env             dummy__C__default-c__1.env                        nbody__C__default-c__0.env
dummy__CPython__default-python__1.env             nbody__CPython__default-python__0.env             nbody__C__default-c__1.env
dummy__C__default-c__0.env                        nbody__CPython__default-python__1.env
```

vext01:
(Will fix tests)

vext01:
Not sure why travis is taking so long here. It's been over an hour...

vext01:
https://www.traviscistatus.com/incidents/m41c9md5dg9z

ltratt:
Shall we restart the tests or wait?

vext01:
Travis is fixed, but has a backlog. Let's just wait.

ltratt:
OK.

vext01:
Feel free to review though.

krun.py
```diff
@@ -220,6 +223,10 @@ def inner_main(mailer, on_first_invocation, config, args):
         util.fatal("Instrumentation dir '%s' exists (from an older run?)" %
                    (instr_dir))
 
+    if envlog_dir_exists and on_first_invocation:
+        util.fatal("Env log dir '%s' exists (from an older run?)" %
```
ltratt: I think we can just say "Env log dir '%s' already exists. Please remove it."

ltratt:
This looks OK except I really don't like the racy /tmp/-with-PID hack. I know we've done that sort of thing elsewhere in Krun, but it's caused us problems, and I can't see that it's a good plan to repeat the same idiom ;) Why can't we just use `mktemp -d` and pass that name around?

vext01:
Using a PID in the filename is an improvement on what we did in the past (I have a branch to fix this).

But I don't understand how this is racy. A PID is unique and we are single threaded.

ltratt:
Is there a technical reason that we can't use `mktemp -d`? If so, we should do that, because the OS guarantees there can't be any problems. That way we absolve ourselves from worrying about single-threadedness, or existing directories (PIDs can be reused) etc.

vext01:
We can use `mktemp` (why -`d?`), but it requires some plumbing.

vext01:
(whereas it was easy to know the filename to remove later when it uses a PID, we would have to store the filename somplace so we could pick it up later).

vext01:
A slight variation upon you suggestion would be to store our temp files in a unique directory in /tmp, rather than the files themselves with unique names directly in /tmp. This would be easier to plumb (I think).

vext01:
Scratch that, the files we would want to group together are owned by different users and one one the users may not exists yet.

I'll come up with something...

ltratt:
It's fine to create a directory with `mktemp -d` and then chmod it to g+w (or a+w).

vext01:
The user is created and deleted a lot, so that itself may be racy. I have another plan. Watch this space.

krun.py
vext01: I'm going to stick with just "Env log dir '%s' already exists". If this fires, then krun's internal consistency is broken and the user is on their own to decide what to do.

vext01:
I think this is ready for review.

I wanted to make a top-level unique directory under /tmp which we could use a scratch area for temp files, but this is awkward, as the permissions need to be right. The krun user is created fresh many times in one invocation (due to sanity checks), and we have to guarantee that the uid will be the same each time.

I've fallen back to just using uniquely named temp files directly under /tmp. This is free of the above problem. Keep it simple.

If this is good, my next PR is to do the same to the wrapper script filename.

vext01:
[I also learned that group write permission is not enough to redirect into a file owned by someone else, or to remove a file owned by someone else.]

krun/vm_defs.py
```diff
@@ -108,18 +116,40 @@ def run_exec(self, entry_point, benchmark, iterations, param, heap_lim_k,
                  stack_lim_k, force_dir=None, sync_disks=True):
         pass
 
-    @staticmethod
-    def make_wrapper_script(args, heap_lim_k, stack_lim_k):
-        """Make lines for the wrapper script.
+    def make_wrapper_script(self, args, heap_lim_k, stack_lim_k):
+        """Write the wrapper script.
         Separate for testing purposes"""
```
ltratt: I think we need to add something like "returns a filename for the env file"?

ltratt:
One minor comment only, I think. Looks good.

vext01:
OK, that's fixed.

One question from earlier outstanding. This is always on, right? Or only for instrumentation mode? I think the former is OK due to the way we have implemented it, right?

ltratt:
OK, if you squash, then I can merge.

vext01:
One more minor commit. OK?

vext01:
Argh, I screwed that up. Moment.

vext01:
Try that. Basically, I was confusing myself because group perms came before user perms.

ltratt:
Are there new commits? I'm confused!

vext01:
Yes, but GH has screwed up the order of the comments and commits.

The last two commits are new.

ltratt:
I'm with it! OK, let's squash.

vext01:
Splat.

ltratt:
OK, waiting for tests to pass...

PR #278, opened 2016-10-24
Big SA sweep.

vext01:
Quickly ran static analysis over the code-base and fixed only the most obviously and agreeably (I hope) broken parts. I'm not going to obsess about this, as I could spend all day on this otherwise.

Some nice finds though, including some weird indents, a failing test being shadowed.

OK?


ltratt:
To me it's more of a lint than "full" static analysis, but that's just me being pedantic :) I agree, they're worthwhile fixes.

PR #279, opened 2016-10-24
Only clean rmsr if on Linux.

vext01:
A part of the build cleans rmsr, even on OpenBSD, causing a crash.

OK?


ltratt:
Waiting for the tests now...

PR #281, opened 2016-10-25
Mail tweaks

vext01:
This fixes a couple of issues with the mailer. See individual commits.

It turned out to be quite a bit more involved that I thought.

Sarah, is there any reason for the tests you wrote for the manifest manager to write it's manifests in a non-default place? I think I could simplify the tests a lot without the need to do that. I suppose it means running the tests would mean any existing krun.manifest would not be touched... I suppose that's a good thing, but it's hard to ensure no test ever writes a manifest in the default place. Any thoughts?

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -331,8 +331,7 @@ if __name__ == "__main__":
     debug("arguments: %s"  % " ".join(sys.argv[1:]))
     parser = create_arg_parser()
     setup_logging(parser)
-
-    try:
-        main(parser)
-    except:
-        sys.exit(1)
+    main(parser)
+    # All fatal exceptions (FatalKrunError, AssertionError, ...) end up here.
+    # Although Some do get caught deeper in the stack, (to try to recover from
+    # various failures) but they are always re-raised.
--- a/krun.py
+++ b/krun.py
@@ -176,8 +176,9 @@ def main(parser):
         util.print_session_info(config)
         return
 
-    on_first_invocation = not (os.path.isfile(ManifestManager.PATH) and
-                               os.stat(ManifestManager.PATH).st_size > 0)
+    manifest_filename = ManifestManager.get_filename(config)
+    on_first_invocation = not (os.path.isfile(manifest_filename) and
+                               os.stat(manifest_filename).st_size > 0)
 
     attach_log_file(config, not on_first_invocation)
     debug("Krun invoked with arguments: %s" % sys.argv)
@@ -197,7 +198,7 @@ def main(parser):
         for frame in traceback.format_tb(error_info[2]):
             lines.append(frame)
         msg = "".join(lines)
-        util.log_and_mail(mailer, debug, subject, msg, bypass_limiter=False)
+        util.log_and_mail(mailer, debug, subject, msg, bypass_limiter=True)
         raise exn
 
 
--- a/krun/mail.py
+++ b/krun/mail.py
@@ -18,16 +18,18 @@ SENDMAIL = "/usr/sbin/sendmail"
 
 
 class Mailer(object):
-    def __init__(self, recipients, max_mails):
-        self.recipients = recipients
+    def __init__(self, recipients=None, max_mails=5):
+        if recipients is not None:
+            self.recipients = recipients
+        else:
+            self.recipients = []
         self.hostname = socket.gethostname()
         self.short_hostname = self.hostname.split(".")[0]
 
-        # After sending max_mails emails, we stop sending more so as
-        # not to spam. Some emails however, you will always want to send.
+        # After sending the maximum number of emails, we stop sending more so
+        # as not to spam. Some emails however, you will always want to send.
         # For these use send(..., bypass_limiter=True).
         self.max_mails = max_mails
-        self.n_mails_sent = 0
 
     def set_recipients(self, recipients):
         self.recipients = recipients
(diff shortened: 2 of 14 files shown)


snim2:
I think I was just trying to keep all the test files in the tests directory, so as not to overwrite anything a user may have written.

vext01:
I think other tests will already write into the default location. I agree that this is not ideal, but how can we ensure that the tests never use the default path?

snim2:
Is the default path stored in a variable or something?

vext01:
Yes, as an instance variable (since this PR).

Perhaps the manifest filename should be derived from the config name?

snim2:
Well, if we have the path in a variable, we can change it before each test runs, then change it back afterwards. 

Deriving the path from the config filename seems sensible.

vext01:
How would we do it for every test though? If we do it manually in each test file, it's too easy to forget/get wrong.

vext01:
krun.py assumes that the manifest can not move:
```
    on_first_invocation = not (os.path.isfile(ManifestManager.PATH) and         
                               os.stat(ManifestManager.PATH).st_size > 0)       
```

Ugh, I've made a mess of this.

snim2:
The `_setup()` method in the manifests already does this...

vext01:
Sure, but what about other tests, which just happen to make a manifest in a more implicit manner? Lots do.

vext01:
I wonder if really the manifest filename should be derived from the config filename.

snim2:
Ah, sorry, I had not realised that you were talking about the other tests. The other way around this would be to write a test fixture.

Deriving the manifest name from the config would be good (you could have two Kruns running on the same machine, for testing). It would also mean that the "test manifests" are always stored with the test configs.

vext01:
I'd still have to manually apply the fixture to all tests.

I'll try deriving the manifest filename from the config filename. That way the manifest naturally drops into the test directory, as that is where our test config files are.

snim2:
sounds good to me

vext01:
This is ready for review.

vext01:
OpenBSD tests also pass.

krun.py
```diff
@@ -330,8 +331,4 @@ def attach_log_file(config, resume):
     debug("arguments: %s"  % " ".join(sys.argv[1:]))
     parser = create_arg_parser()
     setup_logging(parser)
-
-    try:
-        main(parser)
-    except:
-        sys.exit(1)
+    main(parser)
```
snim2: Is an exception here still caught somewhere else? Might be worth a comment...

vext01: This is the top-level. The problem was, if an exception occurred (say) one line before this line, krun would exit silently. And this caused me some pain.

krun/mail.py
```diff
@@ -18,33 +18,47 @@
...
+            # anyway.
+            assert bypass_limiter  # Krun can't know how many mails were sent
+            num_mails_sent = 0
+
         if not self.recipients:
             # Don't bother mailing if there are no recipients
             return
 
-        if self.n_mails_sent < self.max_mails or bypass_limiter:
+        if num_mails_sent < self.max_mails or bypass_limiter:
```
snim2: Trivial comment, but I think this would read more easily as `if bypass_limiter or num_mails_sent < self.max_mails`

krun/mail.py
```diff
@@ -18,33 +18,47 @@
...
+             manifest=None):
+        if manifest is not None:
+            num_mails_sent = manifest.num_mails_sent
+        else:
+            # It's OK to call this without a manifest (e.g. outside the
+            # scheduler loop, where there is no manifest to speak of), but
+            # without a manifest we can't know how many emails have been sent.
+            # So, the only time this is OK is if we are skipping the limiter
+            # anyway.
+            assert bypass_limiter  # Krun can't know how many mails were sent
```
vext01: Basically, you can't do that. The assertion will be caught by the exception handler, and result in an email.

krun/mail.py
vext01: ok

krun/mail.py
vext01: (The only time you don't have a manifest, is in initialisation, so it's rare)

krun.py
vext01: I don't think I have anything to say in a comment. What would it say?

snim2:
OK, other than those small comments this looks good to me.

vext01:
Addressed one comment, questions on the others.

krun.py
snim2: Well, I was thinking that if the exn was caught elsewhere you would say that. So if Krun just bails at this point maybe say `# Krun can fail here if parsing fails` or some such.

krun/mail.py
snim2: OK, so this is caught at the top level? Maybe say that in the comment?

krun.py
vext01: All exceptions end up here. As the deeper handlers re-raise.

vext01: Maybe that is the comment.

snim2: Yes, I think so. What we've done is sensible but it won't be obvious how the exceptions flow for a first-time reader of this code (or me in five minutes time)

krun/mail.py
vext01: Depending where it is called, it may go via a couple of intermediate handlers, but yes, it will arrive at the top level, as for all AssertionErrors. We wouldn't want to comment every single assertion, so what makes this one special?

vext01: I've tried to cover this with the comment I just pushed.

vext01:
OK to squash?

snim2:
Sure

vext01:
Squashed -- with skill ;)

PR #283, opened 2016-10-25
Add more filtering into the Linux dmesg filter.

vext01:
This fixes the constant dmesg changes we were seeing on our last run.

OK?


ltratt:
Waiting for the tests then will merge...

PR #284, opened 2016-10-26
Put the filename of the manifest file in the post-exec hook env.

vext01:
(also rename the function, as it is never used for pre-hooks).

OK?

### DIFF
--- a/krun/tests/test_scheduler.py
+++ b/krun/tests/test_scheduler.py
@@ -6,6 +6,7 @@ from krun.tests import BaseKrunTest
 import krun.util
 
 import os, pytest
+import re
 from krun.tests import TEST_DIR
 
 
@@ -221,7 +222,7 @@ class TestScheduler(BaseKrunTest):
     def test_pre_and_post_exec_cmds0002(self, monkeypatch, mock_platform):
         config = Config(os.path.join(TEST_DIR, "one_exec.krun"))
         path = os.path.join(TEST_DIR, "shell-out")
-        cmd = "echo ${KRUN_RESULTS_FILE}:${KRUN_LOG_FILE} > %s" % path
+        cmd = "echo ${KRUN_RESULTS_FILE}:${KRUN_LOG_FILE}:${KRUN_MANIFEST_FILE} > %s" % path
         config.POST_EXECUTION_CMDS = [cmd]
 
         krun.util.assign_platform(config, mock_platform)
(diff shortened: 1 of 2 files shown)


vext01:
Please wait on merging this.

vext01:
Ready.

vext01:
The failure was a network glitch in the CI. Unrelated to the change...

snim2:
Have you restarted that build?

vext01:
Yeah. Didn't know you could do that!

snim2:
The only drawback is that you lose the previous log, so you don't want to use that feature for a "real" bug

krun/tests/test_scheduler.py
```diff
@@ -237,8 +238,13 @@ def test_pre_and_post_exec_cmds0002(self, monkeypatch, mock_platform):
...
         elems = got.split(":")
-        assert elems[0].endswith(".json.bz2")
-        assert elems[1].endswith(".log")
+        assert os.path.basename(elems[0]) == "one_exec_results.json.bz2"
+        assert re.match("one_exec_[0-9]{8}_[0-9]{6}.log", os.path.basename(elems[1]))
+        assert os.path.basename(elems[2]) == "one_exec.manifest"
+
+        # all paths should be in the same dir
+        dirnames = [os.path.dirname(x) for x in elems]
+        assert all([dirnames[i] == dirnames[0] for i in [1, 2]])
```
ltratt: This `assert` feels weird, particularly with the only-for-the-assert variable above it. Could we not do:

```python
assert os.path.dirname(dirnames[0]) == os.path.dirname(dirnames[1])
assert os.path.dirname(dirnames[0]) == os.path.dirname(dirnames[2])
```

vext01:
Try this. OK to squash?

ltratt:
Ah, that's a bit neater. Now if we do:

```python
assert os.path.dirname(elems[0]) == os.path.dirname(elems[1]) == os.path.dirname(elems[2])
```

I think we have a winner.

vext01:
Here we go. (I find this less readable)

ltratt:
I prefer this because now there's not a variable floating around that's only used by an `assert`. But I don't feel *that* strongly about it. Squash as you feel fit, and feel free to use whichever version of the `assert` you prefer.

vext01:
Squashed.

ltratt:
Will merge when tests have run.

vext01:
Once this is merged, there is a partner commit here: https://github.com/softdevteam/warmup_experiment/pull/209

PR #289, opened 2016-11-23
Checksum experiment - do not merge

vext01:
1/2


iterations_runners/iterations_runner.java
```diff
@@ -223,14 +223,15 @@ public static void main(String args[]) throws
...
+        long checksum = 0;
         for (int i = 0; i < iterations; i++) {
             if (debug) {
                 System.err.println("[iterations_runner.java] iteration: " + (i + 1) + "/" + iterations);
             }
 
             // Start timed section
             IterationsRunner.JNI_krun_measure(0);
-            ke.run_iter(param);
+            checksum += ke.run_iter(param, checksum);
```
ltratt: If we pass in `checksum` does it need to be `checksum +=`? Can it not be `checksum =`?

iterations_runners/iterations_runner.py
```diff
@@ -81,7 +82,7 @@
 
         # Start timed section
         krun_measure(0)
-        bench_func(param)
+        checksum += bench_func(param, checksum)
```
ltratt: Ditto to above.

vext01: Let me check these. I fiddled with these a fair bit to avoid a fix point.

iterations_runners/iterations_runner.java
vext01: For this one, we can use `checksum =` but the checksum will then grow linearly, instead of exponentially, thoughts?

vext01:
An outstanding question is, should i move the abs and mod into the top level iteration runner loop. I suppose this would minimise the cost of the evolving checksum.

ltratt:
Let's try it as-is. If things change a lot, we can then try moving things to the top-level and retry.

vext01:
Ok, so one outstanding query above before I try this out. linear vs non...

ltratt:
Let's give it a go! I suggest a full 10x2000 so that we can see.

PR #290, opened 2016-11-23
Allow dmesg entry for gfx card going into powersave.

vext01:
This graphics card dmesg entry is still causing issues. My theory is that this message appears in dmesg when a monitor is absent for X seconds, and this happens after krun collects its pre-exec dmesg.

The only potential issue is that the message could happen mid-process execution.

The change below ignores the message, if that is the behaviour we want.

Thoughts?


ltratt:
I tend to think we should ignore it. So merge?

vext01:
Sure.

Issue #292, opened 2016-12-09
Post-exec commands


snim2:
The post-exec commands run directly after every benchmark has completed, and typically include commands to bring the network back up and scp the current data and logs to a different machine. 

Because post-execs run *directly* after the benchmark is complete, the log that is tarred up and scp'd does not contain some very useful information (such as ETAs) which are computed between the benchmark completing and the next reboot.

Additionally, data from the *last* benchmark to run is not included in the scp tarball, which is confusing for the user.

snim2:
Fixed in #293 and related PRs.

PR #293, opened 2016-12-12
Dump results earlier, so that the post-execution hooks see new results.

vext01:
For discussion only at this point.

This is a minimal fix to ensure that the results file is up to date at the time of the post-execution commands running.

There was talk of the log being confusing since the last entry in an mid-run log file is not: "reboot in preparation for next exec". That would entail having the "i'm rebooting" message being logged before the post-execution commands have run, which I find a bit confusing, and would require some more involved changes which we would need to consider carefully.

I think we would be wise to play it safe, and just be aware that the log file you see on bencher2 is at the time of the post-execution hook.

If anyone really feels strongly against, we can reconsider.

Thoughts?

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -151,7 +151,7 @@ class ManifestManager(object):
         fh.seek(self.num_mails_sent_offset)
         new_val = self.num_mails_sent + 1
         assert 0 <= new_val <= self.num_mails_maxout
-        fh.write(ManifestManager.NUM_MAILS_FMT % (self.num_mails_sent + 1))
+        fh.write(ManifestManager.NUM_MAILS_FMT % (new_val))
         fh.close()
 
         self._reset()
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -22,13 +22,17 @@ class ManifestManager(object):
     NUM_MAILS_BYTES = 4  # number of bytes used for the (ASCII) field
     NUM_MAILS_FMT = "%%0%dd" % NUM_MAILS_BYTES
 
+    NUM_REBOOTS_BYTES = 8
+    NUM_REBOOTS_FMT = "%%0%dd" % NUM_REBOOTS_BYTES
+
     def __init__(self, config, new_file=False):
         """If new_file is True, write a new manifest file to disk based on the
         contents of the config file, otherwise parse the (existing) manifest
         file corresponding with the config file."""
 
-        # The max number the field can accommodate
+        # Maximum values for mutible header fields
         self.num_mails_maxout = 10 ** ManifestManager.NUM_MAILS_BYTES - 1
+        self.num_reboots_maxout = 10 ** ManifestManager.NUM_REBOOTS_BYTES - 1
 
         self.path = ManifestManager.get_filename(config)
         if new_file:
@@ -59,6 +63,8 @@ class ManifestManager(object):
         self.completed_exec_counts = {}  # including errors
         self.skipped_keys = set()
         self.non_skipped_keys = set()
+        self.num_reboots = -1
+        self.num_reboots_offset = None
 
     def _open(self):
         debug("Reading status cookie from %s" % self.path)
@@ -92,6 +98,9 @@ class ManifestManager(object):
                 elif key == "num_mails_sent":
                     self.num_mails_sent = int(val)
                     self.num_mails_sent_offset = offset + len(key) + 1  # +1 to skip '='
+                elif key == "num_reboots":
+                    self.num_reboots = int(val)
+                    self.num_reboots_offset = offset + len(key) + 1
                 else:
                     util.fatal("bad key in the manifest header: %s" % key)
                 offset += len(line)
@@ -143,6 +152,20 @@ class ManifestManager(object):
         new_val = self.num_mails_sent + 1
         assert 0 <= new_val <= self.num_mails_maxout
         fh.write(ManifestManager.NUM_MAILS_FMT % (self.num_mails_sent + 1))
+        fh.close()
+
+        self._reset()
+        self._parse()  # update stats
+
+    def update_num_reboots(self):
+        """Updates the reboot count header in the manifest file."""
+
+        debug("Increment reboot count in manifest")
+        fh = self._open()
+        fh.seek(self.num_reboots_offset)
+        new_val = self.num_reboots + 1
+        assert 0 <= new_val <= self.num_reboots_maxout
+        fh.write(ManifestManager.NUM_REBOOTS_FMT % (new_val))
         fh.close()
 
         self._reset()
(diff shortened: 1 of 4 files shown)


ltratt:
Given that no lines are deleted, won't this write out everything twice?

vext01:
Read the comment in the diff.

ltratt:
The comment doesn't change the fact that this behaviour is surely wrong.

snim2:
No new results are being appended to the list of results before the JSON is written out, so I think the `results.write...` will just write out two identical JSON files.

vext01:
@ltratt It's slightly inefficient, but not wrong per se.

@snim2 The results are appended a few lines before:
`self.results.append_exec_measurements(job.key, measurements)`

ltratt:
If things are written out twice, how can that not be wrong? I'm willing to be convinced otherwise, but I need some sort of reason to do so!

snim2:
@vext01 exactly, so the results are not appended twice, and the results file will always be correct. regardless of writing it to disk twice.

vext01:
A second write doesn't mean you get a duplicate result. Remember it's JSON, so you can't operate on it incrementally. The order of events is:

 * run exec
 * load results from json
 * append result
 * write json to disk
 * do some book-keeping
 * increment reboot count in json
 * dump json (only difference is the reboot count at this point)
 * reboot

ltratt:
Why don't we just remove the second `write_to_file` call? Is there any reason at all to keep it around?

vext01:
Hrm, it probably could be removed, yeah, but it would mean moving the increment of the reboot count further away from the reboot. What do you think?

(In an ideal world, that reboot count belongs in the manifest).

snim2:
How hard would it be to move the reboot count into the manifest?

vext01:
@ltratt The only reason we write a second time is to ensure that the mutated reboot count is written to disk. That is the crux of it.

@snim2 Probably not super-hard, but as you can probably see, I'm trying to make conservative changes so that I can have a stress-free xmas break ;)

That said, I'm willing to give it a shot if only to see the extent of the changes.

ltratt:
Let's give it a shot. We've still got plenty of time to do a few small test runs and see what breaks.

vext01:
OK.

I'll append to this branch if that is OK?

snim2:
:+1:

ltratt:
Out of curiosity, do we even need to keep a reboot counter? Or is that what's used to determine where in the manifest we are? If the latter, can we not automatically increment it by 1 on load, thus neatly solving the issue?

[And yes, in this branch would be good please.]

vext01:
The reboot counter is part of the boot-loop detector that @snim2 implemented. It's worth keeping.

vext01:
@ltratt And did you have any comment on the log file sequencing? 

> There was talk of the log being confusing since the last entry in an mid-run log file is not: "reboot in preparation for next exec". That would entail having the "i'm rebooting" message being logged before the post-execution commands have run

ltratt:
I can live with the log situation (it's not ideal, and I'd prefer that it's fixed, but it's not a deal breaker).

vext01:
To make the ordering of log messages appear in that manner means making the order in the code somewhat unintuitive. Also note that when the experiment is finished, the log file copied to bencher2 will still be incomplete, as there will be more log messages to come after the last post-execution hook.

Either way, I tend to agree that this is a separate issue. The main thing is that we have a complete results file.

ltratt:
I'm happy to leave it to later. [Personally I'd just remove the "rebooting" log message and move everything else to just before where we execute the post hooks. But that's me.]

vext01:
Take a look at this.

We now take the reboot count out of the manifest (instead of from the results file) and update it in the same way as the flags. This means have no business writing the results file twice any more.

So your typical manifest file looks like this now:
```
eta_avail_idx=4                                                              
num_mails_sent=0000                                                                                      
num_reboots=00000000                                                                                     
keys                                                                                                     
O dummy:Java:default-java
...
```

After this, I have one more change to make to to make the manifest parser a little more robust. Separate PR though.

snim2:
This is a really very stupid question, but what is the maximum number of reboots we can store in the manifest? It looks like 255, but I think I must have misunderstood the code.

In the current experiment we have (I think) `27` VM/benchmark pairs. We will be running `30` process executions of each, that makes `30 x 27` reboots, or `810` in total, right?

vext01:
The maximum is `99999999`. It depends upon the (constant) value of `NUM_REBOOTS_BYTES` (currently set to 8 (ascii) bytes, hence eight `9`s is the max)

vext01:
Maybe you are thinking of binary bits, not ascii bytes?

snim2:
Ah, indeed I was!

ltratt:
Am I correct in thinking the manifest file is supposed to be of fixed size? If so, should `eta_avail_idx=4` be `=00004` or something like that?

krun/scheduler.py
```diff
@@ -148,6 +156,18 @@ def update_num_mails_sent(self):
         self._reset()
         self._parse()  # update stats
 
+    def update_num_reboots(self):
+        """Updates the reboot count header in the manifest file."""
+
+        debug("Increment reboot count in manifest")
+        fh = self._open()
+        fh.seek(self.num_reboots_offset)
+        fh.write(ManifestManager.NUM_REBOOTS_FMT % (self.num_reboots + 1))
```
ltratt: Should there be an `assert` here that `num_reboots` can't overflow `NUM_REBOOTS_FMT`? Or does Python's stringf do that implicitly?

vext01:
Yeah, and that field is never mutated, so it can't change the size of the file after it is written the first time.

ltratt:
Aha, OK! Then I'm fine with this I think. Has it been tested?

vext01:
It's been unit tested and travis has run the example scenario. I'll quickly run a dry run of the real experiment.

ltratt:
Can we run a few small benchmarks with a low number of iterations after the dry run? It would be good to check that we really haven't broken anything and that we've fixed what we hoped to fix.

snim2:
Just FYI the example that Travis uses runs one pexec of Java/C/Python nbody and a dummy benchmark (of course, it's nice to do a bit more for a potentially tricky PR).

ltratt:
The thing is we can't actually tell from the tests/Travis whether this actually does what we want it to. And, at this stage, I'd prefer to be very sure that we haven't accidentally broken some weird internal invariant.

vext01:
Dry run for the warmup experiment (with just PyPy) works fine.

I also checked that in the final tarball it copied, the data is the correct length now:
```
(u'nbody:PyPy:default-python', 10)
(u'fannkuch_redux:PyPy:default-python', 10)
(u'fasta:PyPy:default-python', 10)
(u'binarytrees:PyPy:default-python', 10)
(u'spectralnorm:PyPy:default-python', 10)
(u'richards:PyPy:default-python', 10)
```

krun/scheduler.py
vext01: Probably does need an overflow check.

vext01: Fixed.

vext01:
This contributes to #292

ltratt:
Right, so it looks like we need to do 1) do a real test (with reboots) (on a subset of benchmarks) 2) report on its sucess (or otherwise) 3) squash and merge. Sound like a plan?

vext01:
Can I use bencher3 or 5, where I still have the old vms around?

ltratt:
Sounds like a plan to me.

vext01:
Which one?

snim2:
My benchmarks on b5 have finished, if that helps

ltratt:
I don't think it matters.

vext01:
I'm taking 3.

vext01:
I ran a 2*30 run on bencher3 overnight (with the old VMs). No crashes and the results file is complete.

ltratt:
Excellent! If you squash, then I can merge.

vext01:
Squashed.

vext01:
(The latter two commit, although related, are separate, so I left them apart)

PR #294, opened 2016-12-13
When loading a manifest file, check all headers are present.

vext01:
A small change to make the manifest parser more robust.

OK?

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -25,6 +25,8 @@ class ManifestManager(object):
     NUM_REBOOTS_BYTES = 8
     NUM_REBOOTS_FMT = "%%0%dd" % NUM_REBOOTS_BYTES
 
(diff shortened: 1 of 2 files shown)


krun/scheduler.py
```diff
@@ -84,6 +86,7 @@ def _parse(self):
         self._reset()
         fh = self._open()
         offset = 0
+        seen_headers = []
```
ltratt: Why not make this a set?

ltratt: Yes, but that's less efficient (and less obvious). Might as well bite the bullet upfront!

vext01: There you go.

ltratt:
Perfect! Let's squash and merge.

vext01:
splat.

PR #295, opened 2016-12-14
Revise MALLOC_FLAGS for OpenBSD-6.0.

vext01:
On OpenBSD>=6.0:

```
$ MALLOC_OPTIONS=sfghjpru date
malloc() warning: unknown char in MALLOC_OPTIONS
malloc() warning: unknown char in MALLOC_OPTIONS
Wed Dec 14 14:46:24 GMT 2016
```

The offending flags are `h` and `p`, which have been removed. From a 5.8 machine:

```
     H       ``Hint''.  Pass a hint to the kernel about pages we don't use.
             If the machine is paging a lot this may help a bit.
...
     P       ``Move allocations within a page.'' Allocations larger than half
             a page but smaller than a page are aligned to the end of a page
             to catch buffer overruns in more cases.  This is the default.
```

While I was here, I reviewed (and sorted) the remaining flags. Our choice of `f` is awkward:

```
     F       “Freeguard”.  Enable use after free detection.  Unused pages on
             the freelist are read and write protected to cause a segmentation
             fault upon access.  This will also switch off the delayed freeing
             of chunks, reducing random behaviour but detecting double free(3)
             calls as early as possible.  This option is intended for
             debugging rather than improved security (use the U option for
             security).
```

On the one hand, turning this on reduces nondeterminism at the cost of computation (extra `mprotect()`s to protect the freelist pages (?)). What do you think?

Now we have:

```
$ MALLOC_OPTIONS=fgjrsu date
Wed Dec 14 15:08:03 GMT 2016
```

(No warnings).

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -467,10 +467,12 @@ class OpenBSDPlatform(UnixLikePlatform):
     GET_SETPERF_CMD = "sysctl hw.setperf"
 
     # flags to set OpenBSD MALLOC_OPTIONS to. We disable anything that could
-    # possibly introduce non-determinism.
-    # 's' first acts to reset free page cache to default.
+    # possibly introduce non-determinism or extra computation.
     # See malloc.conf(5) and src/lib/libc/stdlib/malloc.c for more info.
-    MALLOC_OPTS = "sfghjpru"
+    # Although 'F' would improve determinism, it seems to impact performance
+    # (enough for us to care).
+    MALLOC_OPTS = "cfgrux"
+    MALLOC_CONF = "/etc/malloc.conf"
 
     def __init__(self, mailer, config):
         UnixLikePlatform.__init__(self, mailer, config)
@@ -613,6 +615,10 @@ class OpenBSDPlatform(UnixLikePlatform):
 
     def _malloc_options_sanity_check(self):
         """Checks MALLOC_OPTIONS are set"""
+
+        if os.path.lexists(OpenBSDPlatform.MALLOC_CONF):
+            fatal("%s exists. Please remove it" %
+                  OpenBSDPlatform.MALLOC_CONF)
 
         from krun import EntryPoint
         ep = EntryPoint("check_openbsd_malloc_options.so")
--- a/platform_sanity_checks/check_openbsd_malloc_options.c
+++ b/platform_sanity_checks/check_openbsd_malloc_options.c
@@ -2,7 +2,7 @@
 #include <stdlib.h>
 #include <string.h>
 
-#define EXPECT_OPTS     "sfghjpru"
+#define EXPECT_OPTS     "cfgrux"
 
 void
 run_iter(int param)


ltratt:
If we're turning *some* options off (lower-case), why not turn *all* of them off, just to be safe? e.g. it's not clear to me what not specifying `c` or `C` means.

vext01:
Please check this. As I understand: `c` means turn something off, `C` means turn it on. If `C` appears in the manual it means `c` is the default.

Looking through he list, the only one that is not off by default is `j` (but note `j` and `J` are special).

I'd rather not sledgehammer the flags all off without an audit of the meanings. I'll do that now, can you do the same and we can see if we agree.

vext01:
I'm not sure what the best thing to do with `f` is. Perhaps because determinism (and nto performance) is king for us, we should turn this ON.

That aside, I think we could turn everything off.

ltratt:
How do you know which flags are on by default? I can't see any obvious discussion of this anywhere...

My initia thought is that we want something like `cdfgrsux`.

vext01:
I think the defaults are determined by whether the manual page describes the option with an upper case or lower case character. An upper case character describes option which is off by default but can be switched on, and vica versa. It could be clearer.

`J` and `j` are special as they are mutators for a integer counter.

vext01:
Can you justify your decision to leave `f` off? I proposed the opposite above, but can be convinced otherwise.

ltratt:
I don't think the "if the manual uses an upper-case" thing is correct, since no option except `j` has a lower-case description.

W.r.t. `f`, why would we want it on? It can only slow things down (a bit), and make the comparison with Linux a bit less meaningful.

vext01:
OK. I'm going to have to read the code tomorrow to know if my theory is corrct about the defaults, although I'm not sure what `j` being the only lower-case switch has to do with it. If I am right, it just means that all of the flags start off (apart from junking which starts at 1, and can be incremented or decremented in the range [0, 2]).

As for `f`, it's a tradeoff between performance and determinism:

```
     F       “Freeguard”.  Enable use after free detection.  Unused pages on
             the freelist are read and write protected to cause a segmentation
             fault upon access.  This will also switch off the delayed freeing
             of chunks, reducing random behaviour but detecting double free(3)
             calls as early as possible.  This option is intended for
             debugging rather than improved security (use the U option for
             security).
```

Note this part: `This will also switch off the delayed freeing of chunks, reducing random behaviour`

So we need to decide if performance or determinism is king. Determinism, right?

ltratt:
Good question. Probably determinism, though I suppose it partly depends on its impact on performance. Maybe we could quickly try a couple of the C benchmarks with `F`/`f`?

vext01:
Good idea. I'll get that going tomorrow at the thrash.

ltratt:
Our shared understanding after a conversation is that we want `cdfgrux`, and we want Krun to check that `/etc/malloc.conf` doesn't exist (so that `j` and `<` are just the defaults).

vext01:
Oddly, `d` makes a warning!

```
$ MALLOC_OPTIONS=d date       
malloc() warning: unknown char in MALLOC_OPTIONS
```

I think i need to read the code to know what's going on here.

ltratt:
Ah, let's go for `cfgrux`.

vext01:
Ah!

```
#ifdef MALLOC_STATS
	case 'd':
		mopts.malloc_stats = 0;
		break;
	case 'D':
		mopts.malloc_stats = 1;
		break;
#endif /* MALLOC_STATS */
```

It does actually say that in the man page, but I missed it.

So `cfgrux` it is.

ltratt:
OK.

vext01:
Ready for review.

krun/platform.py
```diff
@@ -467,10 +467,12 @@ class OpenBSDPlatform(UnixLikePlatform):
...
     # flags to set OpenBSD MALLOC_OPTIONS to. We disable anything that could
-    # possibly introduce non-determinism.
-    # 's' first acts to reset free page cache to default.
+    # possibly introduce non-determinism or extra computation.
     # See malloc.conf(5) and src/lib/libc/stdlib/malloc.c for more info.
-    MALLOC_OPTS = "sfghjpru"
+    # Although 'F' would improve determinism, it seems to impact performance
+    # (enough for us to care).
+    MALLOC_OPTS = "cfgrux"
+    MALLOC_CONF = os.sep + os.path.join("etc", "malloc.conf")
```
ltratt: I think we can just say `MALLOC_CONF = "/etc/malloc.conf"` -- OpenBSD is not going to change what `os.sep` is any time soon ;)

vext01: Good point.

ltratt:
Let's squash then I can merge.

krun/platform.py
vext01: fixed.

vext01:
splat.

ltratt:
Ut-oh, Travis error. Admittedly, I don't think it's related to this PR, so I'm inclined to merge this one, and suggest we make a separate PR to fix the Travis error. Thoughts?

vext01:
It's just an intermittent travis CI network error. I've kicked the build, and they look to be doing ok now.

ltratt:
Aha, OK!

vext01:
Yeah, we've seen this before occasionally. You just have to hit the "restart build" button on the travis details page to run CI again.

PR #297, opened 2016-12-20
Ensure sampling of cold temperature is enclosed in pre/post hooks and put temperatures into manifest.

vext01:
This moves the sampling of the cold sensors into the exception handler so that it is subject to pre/post hooks. The starting temperatures have also been moved into the manifest.

I'm not really happy with this. The change is huge.

Does anyone have any suggestions? I wonder if instead of moving the sampling inside the execution handler, we should add another "critical section" (wrapped in hooks) which deals only with reading the sensors.

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -499,6 +499,12 @@ class ExecutionScheduler(object):
                 results = Results(self.config, self.platform)
                 results.write_to_file()
 
+                # We still need to run the post-execution commands
+                util.run_shell_cmd_list(
+                    self.config.POST_EXECUTION_CMDS,
+                    extra_env=self._make_post_cmd_env(results)
+                )
+
                 info("Reboot prior to first execution")
                 self._reboot()
             else:
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -25,12 +25,19 @@ class ManifestManager(object):
     NUM_REBOOTS_BYTES = 8
     NUM_REBOOTS_FMT = "%%0%dd" % NUM_REBOOTS_BYTES
 
+    # Start temperature stored in-manifest as YYYYY.ZZ
+    START_TEMPERATURE_BYTES = 8
+    START_TEMPERATURE_FMT = "%08.2f"
+
+    # Mandatory manifest header fields (others exist, e.g. start temperatures)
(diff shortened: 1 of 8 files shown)


krun/scheduler.py
```diff
@@ -112,7 +125,7 @@ def _parse(self):
             assert False
 
         # Check we saw all the necessary header fields
-        assert seen_headers == ManifestManager.HEADER_FIELDS
+        assert ManifestManager.HEADER_FIELDS.issubset(seen_headers)
```
snim2: I'm not sure I understand this, it seems to say that the header fields in the manifest should be a subset of the headers that have been read so far by the parser. Is that right?

krun/scheduler.py
```diff
@@ -193,6 +206,17 @@ def update(self, flag):
         self._reset()
         self._parse()  # update stats
 
+    def set_starting_temperatures(self, dct):
```
snim2: I had expected the manifest object to do the actual writing out.

krun/scheduler.py
```diff
@@ -360,41 +386,39 @@ def __init__(self, config, mailer, platform, dry_run=False,
...
-            self.results = Results(self.config, self.platform)
+            self.manifest = ManifestManager(config, platform, new_file=True)
         else:
-            self.manifest = ManifestManager(self.config)
-            self.results = None
+            self.manifest = ManifestManager(self.config, platform)
 
-    def get_estimated_exec_duration(self, key):
-        previous_exec_times = self.results.eta_estimates.get(key)
+    def get_estimated_exec_duration(self, results, key):
```
snim2: Why does the results object need to be passed in to these methods?

vext01: The ETA info is in the results file. This was stored in self, but I removed it so we could more easily see the extent of its use.

krun/scheduler.py
vext01: There's a variable number of sensor headers that will vary per system

snim2: OK, but I'm not sure this is a very strong assertion. Perhaps another way would be to have the per-platform fields stored in a set inside each `Platform` class, then you could check whether the seen headers were exactly correct.

vext01: Ok, i'll apply a fix in the other branch, as this code is shared there.

PR #298, opened 2016-12-20
Ensure collection of start temperatures is guarded by hooks.

vext01:
This ensures that the temperature collection part of krun is wrapped in pre/post hooks, so as to bring down the network etc.

In doing so, i've untangled the initial reboot from the main benchmarking loop. @snim2 and I discussed this earlier and thought this would make the code more comprehensible.

I also added tests for parts of the code which make me nervous ;)

My next PR will kill the `self.results` in `ExecutionScheduler` so as to make it's scope more explicit, and as laurie suggested, use a flag to ensure results are never loaded in before a benchmark is run.

Testing with reboots on bencher7, tested without reboots on bencher6.

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -265,7 +265,12 @@ def inner_main(mailer, on_first_invocation, config, args):
             util.fatal(error_msg)
 
         debug("Using pre-recorded initial temperature readings")
-        platform.starting_temperatures = current.starting_temperatures
+        manifest = ManifestManager(config, platform)
+
+        platform_temps = {}
+        for sensor, tup in manifest.starting_temperatures.iteritems():
+            platform_temps[sensor] = tup[1]
+        platform.starting_temperatures = platform_temps
     else:
         # Touch the config file to update its mtime. This is required
         # by when resuming a partially complete benchmark session, in which
@@ -274,13 +279,42 @@ def inner_main(mailer, on_first_invocation, config, args):
         if rc != 0:
             util.fatal("Could not touch config file: " + args.filename)
 
-        info(("Wait %s secs to allow system to cool prior to "
-             "collecting initial temperature readings") %
-             config.TEMP_READ_PAUSE)
-        platform.sleep(config.TEMP_READ_PAUSE)
+        manifest = ManifestManager(config, platform, new_file=True)
+        if manifest.num_execs_left == 0:
+            # No executions, or all skipped
+            fatal("Empty schedule!")
 
-        debug("Taking fresh initial temperature readings")
-        platform.starting_temperatures = platform.take_temperature_readings()
+        try:
+            info(("Wait %s secs to allow system to cool prior to "
+                 "collecting initial temperature readings") %
+                 config.TEMP_READ_PAUSE)
+
+            # This part is wrapped in hooks, so that if daemons or networking are
+            # taken down for process executions, then the initial temperature
+            # reading gets the same treatment.
+            util.run_shell_cmd_list(config.PRE_EXECUTION_CMDS,)
+            platform.sleep(config.TEMP_READ_PAUSE)
+
+            debug("Taking fresh initial temperature readings")
+            platform.starting_temperatures = platform.take_temperature_readings()
+            manifest.set_starting_temperatures(platform.starting_temperatures)
+
+            # Write out an empty results file. After the initial reboot Krun
+            # will expect this to exist.
+            results = Results(config, platform)
+            results.write_to_file()
+        except:
+            raise
+        finally:
+            util.run_shell_cmd_list(config.POST_EXECUTION_CMDS,)
+
+        log_path = config.log_filename(resume=False)
+        util.log_and_mail(mailer, debug,
+                          "Benchmarking started",
+                          "Benchmarking started.\nLogging to %s" %
+                          log_path, bypass_limiter=True)
+
+        util.reboot(manifest, platform)
 
     # Assign platform to VM defs -- needs to happen early for sanity checks
     util.assign_platform(config, platform)
@@ -291,8 +325,7 @@ def inner_main(mailer, on_first_invocation, config, args):
     sched = ExecutionScheduler(config,
                                mailer,
                                platform,
-                               dry_run=args.dry_run,
-                               on_first_invocation=on_first_invocation)
+                               dry_run=args.dry_run)
     sched.run()
 
 
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -311,10 +311,6 @@ class ExecutionJob(object):
                 self.key == other.key and
                 self.parameter == other.parameter)
 
-    def add_exec_time(self, exec_time):
-        """Feed back a rough execution time for ETA usage"""
-        self.sched.add_eta_info(self.key, exec_time)
-
     def run(self, mailer, dry_run=False):
         """Runs this job (execution)"""
 
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -25,12 +25,19 @@ class ManifestManager(object):
     NUM_REBOOTS_BYTES = 8
     NUM_REBOOTS_FMT = "%%0%dd" % NUM_REBOOTS_BYTES
 
+    # Start temperature stored in-manifest as YYYYY.ZZ
+    START_TEMPERATURE_BYTES = 8
+    START_TEMPERATURE_FMT = "%08.2f"
+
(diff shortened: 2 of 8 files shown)


krun.py
```diff
@@ -277,11 +281,34 @@ def inner_main(mailer, on_first_invocation, config, args):
...
+        # This part is wrapped in hooks, so that if daemons or networking are
+        # taken down for process executions, then the initial temperature
+        # reading gets the same treatment.
+        util.run_shell_cmd_list(config.PRE_EXECUTION_CMDS,)
         platform.sleep(config.TEMP_READ_PAUSE)
 
         debug("Taking fresh initial temperature readings")
         platform.starting_temperatures = platform.take_temperature_readings()
 
+        # Update the starting temperatures in the manifest
```
snim2: Does this only happen on first invocation? (I think GH has not given us much context)

krun.py
```diff
@@ -265,7 +265,11 @@ def inner_main(mailer, on_first_invocation, config, args):
             util.fatal(error_msg)
 
         debug("Using pre-recorded initial temperature readings")
-        platform.starting_temperatures = current.starting_temperatures
+        manifest = ManifestManager(config, platform)
+        platform.starting_temperatures = {
+            sensor: tup[1] for sensor, tup in
+            manifest.starting_temperatures.iteritems()
```
vext01: I just let the pep plugin do it's thing here.

vext01:
I think i would also like to kill the `self.results` field in the execution scheduler, but perhaps as a new PR. I also have pending changes to never load the results prior to running a benchmark. It could go in that.

snim2:
Sounds sensible.

vext01:
Anything else before I test this on bencher7 and bsd?

vext01:
Ah i see a few oddities. One moment.

vext01:
OK. Any further comments?

snim2:
Not from me... test away

vext01:
I think this is ready for review again.

I'm going to update the description.

krun.py
ltratt: I agree with Sarah: this is really confusing to read at the moment. [I must admit that, seeing it for the second time, I think that I dislike the dict comprehension syntax, though that's a secondary issue.]

krun.py
```diff
@@ -274,13 +278,43 @@ def inner_main(mailer, on_first_invocation, config, args):
...
+            # This part is wrapped in hooks, so that if daemons or networking are
+            # taken down for process executions, then the initial temperature
+            # reading gets the same treatment.
+            util.run_shell_cmd_list(config.PRE_EXECUTION_CMDS,)
+            platform.sleep(config.TEMP_READ_PAUSE)
+
+            debug("Taking fresh initial temperature readings")
+            platform.starting_temperatures = platform.take_temperature_readings()
+
+            # Update the starting temperatures in the manifest
```
ltratt: Not a useful comment.

krun.py
```diff
@@ -274,13 +278,43 @@ def inner_main(mailer, on_first_invocation, config, args):
...
+            manifest.set_starting_temperatures(platform.starting_temperatures)
+
+            # Scaffold results file
```
ltratt: Not a useful comment.

krun/tests/test_scheduler.py
```diff
@@ -35,24 +38,43 @@ def dummy_do_reboot(self):
...
         raise TestReboot()
-    monkeypatch.setattr(ExecutionScheduler, '_do_reboot', dummy_do_reboot)
+    monkeypatch.setattr(util, '_do_reboot', dummy_do_reboot)
 
 
+def no_envlogs(monkeypatch):
+    def dummy_stash_envlog(tmp_filename, config, platform, key, exec_num):
+        pass
+    monkeypatch.setattr(util, 'stash_envlog', dummy_stash_envlog)
+
```
ltratt: Extraneous blank line? [It's a bit unclear to me if the standard in this file is 1 or 2 blank lines between functions.]

ltratt:
I have no deep comments, although I admit that I found it a little hard to understand the full ramifications of the PR.

vext01:
That doesn't sound good. Do you want to ask more questions while I fix up the comments?

krun/tests/test_scheduler.py
vext01: It's two lines between global functions and one between methods, although the krun source isn't very consistent. I fix them when I change lines nearby, but not otherwise.

ltratt:
No, I don't think it's a bad PR, just that I don't know some of Krun's internal logic well enough to fully understand what some of the changes are. I think we'll find out more when we run a small test invocation.

krun.py
vext01: I flattened that to a loop then.

[Note the dict comprehension syntax is almost identical to list comprehension syntax: `[x for x in y]` vs. `{k: v for k, v in y}`. Maybe it's just that this was a non-trivial comprehension that it looked odd]

krun.py
vext01: removed.

krun.py
vext01: Revised, as there is something non-obvious going on here.

vext01:
I've fixed your comments, and re-tested on bencher2.

vext01:
(Note that another PR is to follow to deal properly with Results instances -- this is a intermediate step forward)

krun.py
```diff
@@ -265,7 +265,12 @@ def inner_main(mailer, on_first_invocation, config, args):
             util.fatal(error_msg)
 
         debug("Using pre-recorded initial temperature readings")
-        platform.starting_temperatures = current.starting_temperatures
+        manifest = ManifestManager(config, platform)
+
```
snim2: Blank line?

krun.py
```diff
@@ -265,7 +265,12 @@ def inner_main(mailer, on_first_invocation, config, args):
...
+        platform_temps = {}
```
snim2: Did we say `dict()` elsewhere?

krun/scheduler.py
```diff
@@ -67,6 +74,7 @@ def _reset(self):
         self.non_skipped_keys = set()
         self.num_reboots = -1
         self.num_reboots_offset = None
+        self.starting_temperatures = {} # name -> (offset, degrees C floats)
```
snim2: `dict()`? (I don't care which it is, so long as we are consistent!)

krun/scheduler.py
```diff
@@ -112,7 +125,9 @@ def _parse(self):
             assert False
 
         # Check we saw all the necessary header fields
-        assert seen_headers == ManifestManager.HEADER_FIELDS
+        expect_extra_headers = \
+            set(["start_temp_%s" % x for x in self.platform.temp_sensors])
+        assert ManifestManager.HEADER_FIELDS | expect_extra_headers == seen_headers
```
snim2: This is nicer!

krun.py
vext01: It was deliberate, but I can kill it if you like.

krun/scheduler.py
```diff
@@ -498,6 +501,7 @@ def run(self):
             self.add_eta_info(job.key, eta_info)
             self.manifest.update(flag)
         except Exception:
+            flag = 'E'
```
snim2: This isn't set on line 483? Is that because 483 might raise the exception?

krun/tests/mocks.py
```diff
@@ -32,14 +32,15 @@ def __init__(self, mailer, config):
         self.num_cpus = 0
         self.num_per_core_measurements = 0
         self.no_user_change = True
+        self.temp_sensors = []
```
snim2: `list()`?

krun.py
vext01: I don't know. I don't see why we shouldn't use the concise syntax. You wouldn't say `dict((1, 2), (3, 4))`, you'd use `{1: 2, 3:4}`, so why `dict()`?

krun/scheduler.py
vext01: Thanks!

snim2:
OK, finished reading through, and the comments are all minor. Probably you won't need to test again on bencher2!

vext01:
OK to squash?

snim2:
+1 from me

ltratt:
If Sarah's OK with this, I think you should squash, and she should merge.

vext01:
squashed.

PR #300, opened 2016-12-21
Audit results inst

vext01:
More careful results handling.

See individual commit messages for more.

OK?

### DIFF
--- a/krun/results.py
+++ b/krun/results.py
@@ -11,7 +11,14 @@ class Results(object):
     Can be serialised to disk.
     """
 
+    # We use this to detect times where a results instance is loaded prior to a
+    # process execution. This wouuld be bad, as the results can be big and
+    # cause memory to fragment.
+    ok_to_instantiate = False
+
     def __init__(self, config, platform, results_file=None):
+        self.instantiation_check()
+
         self.config = config
         self.platform = platform
 
@@ -47,6 +54,10 @@ class Results(object):
         # Import data from a Results object serialised on disk.
         if results_file is not None:
             self.read_from_file(results_file)
+
+    def instantiation_check(self):
+        if not Results.ok_to_instantiate:
+            fatal("Results instance loaded prior to a process execution")
 
     @property
     def audit(self):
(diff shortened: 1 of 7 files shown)


krun/scheduler.py
```diff
@@ -384,37 +384,35 @@ def __init__(self, config, mailer, platform, dry_run=False):
         self.dry_run = dry_run
         self.log_path = self.config.log_filename(resume=True)
         self.manifest = ManifestManager(config, platform)
-        self.results = None
```
snim2: I think it might be good to replace this with a comment to explain why we don't do the obvious thing (just in case we forget and try to 'simplify' the code later on).

krun/scheduler.py
```diff
@@ -443,12 +441,6 @@ def run(self):
         assert self.manifest.num_execs_left > 0
         self.platform.wait_for_temperature_sensors()
 
-        # If we get here, this is a real run, with a benchmark about to
-        # run. Results should never be in memory at this point (as they
-        # grow over time, and can get very large, thus potentially
-        # influencing the benchmark)
-        assert self.results is None
```
vext01: It's kind of superseded by the `Results` change. No? I could bring it back if you still feel the same.

snim2:
Looks good to me, apart from those view comments above.

krun/results.py
```diff
@@ -11,7 +11,14 @@ class Results(object):
     Can be serialised to disk.
     """
 
+    # We use this to detect times where a results instance is loaded prior to a
+    # process execution. This wouuld be bad, as the results can be big and
+    # cause memory to fragment.
+    OK_TO_INSTANTIATE = False
```
ltratt: By making this all capitals, it suggests it's a constant -- which it isn't. I think we should make it `ok_to_instantitate` but make it clear that it's a class variable.

ltratt:
I only have one minor comment.

krun/scheduler.py
snim2: Yes

krun/results.py
vext01: You are indeed correct. I thought that the convention was to make class variables upper case, but I was wrong. Fixing.

ltratt:
OK, I'm happy. I suggest, once Sarah is happy, you squash and she merges.

krun/scheduler.py
vext01: Try that.

snim2:
Looks good! Squash away...

vext01:
Thanks. splat.

PR #303, opened 2016-12-21
Summarise progress from manifest files

snim2:
This PR adds a script written by @vext01 to source control. The script reads a manifest file and prints a summary of the progress through the current experiment, e.g.:

```bash
$ ./scripts/progress.awk warmup.manifest 
C=349  O=191  E=0   64.6296%
```

### DIFF
--- a/scripts/progress.awk
+++ b/scripts/progress.awk
@@ -1,0 +1,7 @@
+#!/usr/bin/awk -f
+BEGIN             {C=0; O=0; E=0; keys=0}
+/^keys/         {keys=1}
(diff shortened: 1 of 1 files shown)


ltratt:
Does this work with the updated manifest w/temperatures in it?

snim2:
I don't have one to try, but it only counts the relevant lines, so should do

ltratt:
Good point. Then I think we can take this in as-is.

vext01:
It ignores all lines preceeding the keys line:
```
/^keys/         {keys=1}
```

PR #304, opened 2016-12-22
Fix two bugs that prevent post-commands running in exception case.

vext01:
OK?

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -451,6 +451,11 @@ class ExecutionScheduler(object):
         job = ExecutionJob(self, vm, self.config.VMS[vm], bench, variant,
                            self.config.BENCHMARKS[bench])
 
+        # Default to error state. This is the value the finally block will see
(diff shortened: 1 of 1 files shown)


krun/scheduler.py
```diff
@@ -511,6 +511,7 @@ def run(self):
             # exception occurred in the above try block, there's a chance that
             # they have not have been loaded.
             if results is None:
+	        Results.ok_to_instantiate = True
```
snim2: The indentation looks wrong here, is that just GitHub?

krun/scheduler.py
```diff
@@ -507,6 +511,7 @@ def run(self):
...
```
vext01: funny indent?

krun/scheduler.py
vext01: fixed.

snim2:
OK, if the tests are good, let's squash

vext01:
splat.

PR #305, opened 2016-12-22
Enable mode 2 ASLR on Linux.

vext01:
Since the virtual addresses of a VM's data can be considered an internal
source of variation.

Also removed recursion, simplified, and tested.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -654,6 +654,7 @@ class LinuxPlatform(UnixLikePlatform):
     CPU_SCALER_FMT = "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_driver"
     KERNEL_ARGS_FILE = "/proc/cmdline"
     ASLR_FILE = "/proc/sys/kernel/randomize_va_space"
+    ASLR_MODE = 2
     CSET_CMD = "/usr/bin/cset"
     USER_CSET_DIR = "/cpusets/user"
 
@@ -799,7 +800,7 @@ class LinuxPlatform(UnixLikePlatform):
         self._check_rmsr_loaded()
         if not self.no_tickless_check:
             self._check_tickless_kernel()
-        self._check_aslr_disabled()
+        self._check_aslr_enabled()
 
     def _check_virt_what_installed(self):
         debug("Check virt-what is installed")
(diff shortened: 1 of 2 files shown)


krun/platform.py
```diff
@@ -1057,26 +1058,26 @@ def _check_rmsr_loaded(self):
...
         debug("rmsr loaded successfully.")
 
-    def _check_aslr_disabled(self):
-        debug("Checking ASLR is off")
-        with open(self.ASLR_FILE, "r") as fh:
-                enabled = fh.read().strip()
-        if enabled == "0":
-            return  # OK!
-        else:
-            # ASLR is off, but we can try to enable it
```
vext01: This comment was outright wrong by the way.

snim2: Yup!

krun/platform.py
```diff
@@ -1057,26 +1058,26 @@ def _check_rmsr_loaded(self):
...
-            debug("Turning ASLR off")
-            cmd = "%s sh -c 'echo 0 > %s'" % \
-                (self.change_user_cmd, self.ASLR_FILE)
-            stdout, stderr, rc = run_shell_cmd(cmd, failure_fatal=False)
+    def _check_aslr_enabled(self):
+        """Check ASLR is cranked to level 2"""
```
snim2: Maybe say what `2` means:

> 0 – No randomization. Everything is static.
>1 – Conservative randomization. Shared libraries, stack, mmap(), VDSO and heap are randomized.
> 2 – Full randomization. In addition to elements listed in the previous point, memory managed through brk() is also randomized.

vext01: Where is that text from, or is it original?

krun/platform.py
```diff
@@ -1057,26 +1058,26 @@ def _check_rmsr_loaded(self):
...
-                fatal(msg)
+        debug("Checking ASLR is on")
+        for rep in 1, 2:
+            with open(LinuxPlatform.ASLR_FILE, "r") as fh:
+                val = fh.read().strip()
+            if val == str(LinuxPlatform.ASLR_MODE):
+                return  # OK!
             else:
-                self._check_aslr_disabled()  # should work this time
+                if rep == 2:
```
snim2: I guess there isn't an easy way to test these failure modes?

vext01: Not easily, I could split out the opening of the file into a new method which could be monkeypatched. Is it worth the effort?

krun/platform.py
snim2: StackOverflow. You probably want to quote this file: https://www.kernel.org/doc/Documentation/sysctl/kernel.txt (grep for `randomize_va_space`)

krun/platform.py
snim2: No, probably not

krun/platform.py
vext01: Fixed.

snim2:
OK, nice! Do you want to squash?

vext01:
splat

PR #306, opened 2016-12-22
Kill noise in a VM sanity check.

vext01:
This sanity check prints to stderr, i'd rather it didn't.


snim2:
Do the other sanity checks print to STDERR or STDOUT?

vext01:
Only if crashing, which is OK.

snim2:
OK then...

PR #313, opened 2017-06-26
Make Krun Work on a Virtualised Linux Host Again

### LINKED ISSUE #310: Krun does not work in a virtualbox VM.
As mentioned via email, I had a hunch krun would not run in a VM as-is. My hunch was right.

```
  File "../krun.py", line 194, in main
  File "../krun.py", line 313, in inner_main
    sanity_checks(config, platform)
  File "../krun.py", line 71, in sanity_checks
    platform.sanity_checks()
  File "/home/vext01/krun/krun/platform.py", line 1120, in sanity_checks
    UnixLikePlatform.sanity_checks(self)
  File "/home/vext01/krun/krun/platform.py", line 397, in sanity_checks
    self._sanity_check_user_change()
  File "/home/vext01/krun/krun/platform.py", line 407, in _sanity_check_user_chang
e
    force_dir=PLATFORM_SANITY_CHECK_DIR)
  File "/home/vext01/krun/krun/util.py", line 259, in spawn_sanity_check
    fatal("%s sanity check failed: %s" % (check_name, e.message))
  File "/home/vext01/krun/krun/util.py", line 52, in fatal
    raise FatalKrunError(msg)

  File "../krun.py", line 358, in <module>
    main(parser)
  File "../krun.py", line 203, in main
    raise exn
krun.util.FatalKrunError: UNIX user change sanity check failed: Benchmark returned
 non-zero or emitted invalid JSON.
Exception string: No JSON object could be decoded
return code: 0
stdout:
arch pctr version >=2 is required! got 0

```

This is a failing sanity check. We use the `CPUID` x86 instruction to ask what performance counters are available. The ones we need are level 2, but the VM returns 0, which probably means none.

I may be able hack around this, as somewhere else in Krun, we are already checking for a virtualised host.

This blocks the artefact.

vext01:
Hi,

These changes ensure that Krun works on a virtualised Linux machine again.

Tested in a VM and on bare metal. On the former, the `aperf`, `mperf` and `core_cycle` JSON keys are now empty lists.

Fixes #310

### DIFF
--- a/libkrun/Makefile
+++ b/libkrun/Makefile
@@ -1,8 +1,28 @@
 LIBKRUNTIME_CFLAGS =	-Wall -shared -fPIC
 COMMON_CFLAGS = -Wall -pedantic -std=gnu99
+SUDO = sudo
 
-ifeq ("${TRAVIS}", "true")
-	COMMON_CFLAGS += -DTRAVIS
+# Sadly this tool requires root
+VIRT_WHAT = ${SUDO} /usr/sbin/virt-what
+
+# Detect if we are virtualised (Linux only for now XXX)
+ifeq ($(shell uname -s),Linux)
+	VIRTUALISED = 0  # default off
+
+	VIRT_FACTS = $(shell ${VIRT_WHAT})
+	ifneq ("${VIRT_FACTS}","")
+		VIRTUALISED = 1
+	endif
+
+	ifeq ("${TRAVIS}","true")
+		VIRTUALISED = 1
+	endif
+
+	# Under virtualised conditions, we have no performance counters.
+	# MSR-centric code is also guarded on a per-OS basis in libkruntime.c
+	ifeq (${VIRTUALISED},1)
+		COMMON_CFLAGS += -DNO_MSRS
+	endif
 endif
 
 ifeq ($(shell uname -s),Linux)
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -687,13 +687,12 @@ class LinuxPlatform(UnixLikePlatform):
         self.temp_sensor_map = None
         UnixLikePlatform.__init__(self, mailer, config)
         self.num_cpus = self._get_num_cpus()
+        self.virt_what_cmd = self._find_virt_what()
         self.num_per_core_measurements = self._get_num_per_core_measurements()
 
-        self.virt_what_cmd = None  # set later
-
     def _get_num_per_core_measurements(self):
-        # For all systems apart from travis we expect per-core measurements
-        if os.environ.get("TRAVIS") == "true":
+        # For all systems apart from virtualised hosts we expect per-core measurements
+        if self.is_virtual():
             return 0
         else:
             return self.num_cpus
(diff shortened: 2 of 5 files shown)


libkrun/Makefile
```diff
@@ -1,8 +1,28 @@
 LIBKRUNTIME_CFLAGS =	-Wall -shared -fPIC
 COMMON_CFLAGS = -Wall -pedantic -std=gnu99
+SUDO = sudo
 
-ifeq ("${TRAVIS}", "true")
-	COMMON_CFLAGS += -DTRAVIS
+# Sadly this tool requires root
```
snim2: That is unfortunate. This approach looked reasonable to me, but maybe I have missed something: https://serverfault.com/a/721591

libkrun/Makefile
```diff
@@ -1,8 +1,28 @@
...
+# Detect if we are virtualised (Linux only for now XXX)
+ifeq ($(shell uname -s),Linux)
+	VIRTUALISED = 0  # default off
+
+	VIRT_FACTS = $(shell ${VIRT_WHAT})
+	ifneq ("${VIRT_FACTS}","")
+		VIRTUALISED = 1
+	endif
+
+	ifeq ("${TRAVIS}","true")
```
snim2: Would it be sensible to rename this environment variable?

snim2:
This looks OK to me, apart from the minor comments I've made. I haven't tested it in anger though...

libkrun/Makefile
vext01: Yes, I remember this post from the time I was implementing `is_virtual()` in Krun. It's not as reliable as `virt-what`.

libkrun/Makefile
vext01: That's one that Travis itself sets. We don't have control over it.

vext01:
If you are satisfied with my response, I can squash.

libkrun/Makefile
snim2: Maybe I have missed something very obvious. We check for `TRAVIS` as an env var, we run `virt-what` and that sets `VIRTUALISED = 1/0` but then to find out whether Krun is running on a virtualise system we do this:

```python
def is_virtual(self):
    if os.environ.get("TRAVIS") == "true":
        return True
```

Shouldn't we also be checking for `VIRTUALISED` in that Python code?

vext01: That's an early return. The remainder of the function checks `virt-what`:
https://github.com/softdevteam/krun/blob/f14b9ebf659d3f41b273d296ebe699fa3e75d304/krun/platform.py#L1204

snim2: OK I see, so that wasn't in the diff. So... why do we need to run `virt_what` in the `Makefile`?

vext01: So that we know whether or not to build libkrun with or without performance counter support. I suppose that could have been runtime checked, but it could have gotten in the way of taking readings.

snim2:
OK, I am finally with you. So, do you want to squash, and also bring this branch up to date with master at the same time?

libkrun/Makefile
vext01: It wasn't in the diff becuase we had always done that check in `is_virtual`. This function however, was not considering travis (if you look at the previous hunk, there was a special case, I just opted to move that in to `is_virtual`).

vext01:
will do

vext01:
splat

PR #315, opened 2017-07-04
Sample license addition for review, before I add them in every file.

vext01:
Here's (to begin with) a sample of what the license looks like. I wanted to check it's right before adding it in lots of places.

Wrapped using `gq` in vim.

OK to continue?

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,7 +433,7 @@ Please make sure you understand the implications of this.
 
 ## Licenses
 
-The *nbody* benchmark comes from the Computer Language Benchmarks Game, which
-are published under a revised BSD license:
+Krun is licensed under the UPL license.
 
-  http://shootout.alioth.debian.org/license.php
+The nbody benchmark is licensed under a revised BSD license:
+http://shootout.alioth.debian.org/license.php
--- a/examples/benchmarks/dummy/c/bench.c
+++ b/examples/benchmarks/dummy/c/bench.c
@@ -1,3 +1,42 @@
+/*
+ * Copyright (c) 2017 King's College London
+ * created by the Software Development Team <http://soft-dev.org/>
+ *
+ * The Universal Permissive License (UPL), Version 1.0
+ *
+ * Subject to the condition set forth below, permission is hereby granted to
+ * any person obtaining a copy of this software, associated documentation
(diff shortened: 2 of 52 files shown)


ltratt:
And I.

vext01:
I think this is the lot.

vext01:
I didn't add licenses to Makefiles. Is that right? Seems excessive...

README.md
```diff
@@ -433,6 +433,9 @@ Please make sure you understand the implications of this.
 
 ## Licenses
 
+The main bulk of Krun is licensed under the UPL license, which is found at the
```
ltratt: This section is verbose and confusing. I'd suggest either removing it or saying "Krun is licensed under the UPL. However, the nbody benchmark is published under a ..." or similar.

ltratt:
Agreed, makefiles is one step too far (except in weird cases... which we don't have). I just have one comment.

README.md
vext01: Made it shorter.

vext01:
The benchmark is provided as an example.

vext01:
Wait, let me re-word slightly.

ltratt:
I don't think it's an example: it's a benchmark we use. I think we can just say "the nbody benchmark is...".

vext01:
No, it really is only an example in this repo.

ltratt:
Ah, OK. But I agree: "example" isn't important to the license.

OK, I think I'm happy. Time to squash?

vext01:
I think so.

ltratt:
OK, squash away.

vext01:
Big splat!

ltratt:
OK, thanks, will wait for the tests then merge.

vext01:
I suppose we should do the same on the warmup experiment itself (tomorrow)

ltratt:
Good point.

vext01:
There we already have in the README:
```
All original files are under the following MIT license
```

Is that still OK?

vext01:
(Obviously, add the header to all source files)

ltratt:
We should change MIT -> UPL IMHO, but otherwise yes.

PR #316, opened 2017-07-13
Online amperf ratio checks

vext01:
Let's start looking at this.

Goal: Check APERF/MPERF ratios after each process execution and restart if one of more in-process iteration has a ratio outside of a configurable bound.

Notes:

 * This duplicates a fair amount of code from the warmup_experiment repo, so I think the next step should probably be to flip the remaining uses to importing the krun versions and killing the originals.

 * We compute the outliers, and then discard them. We should consider (although in another PR) the ability to write the outliers into the results file directly, thus removing the need for the `mark_outliers_in_json` script.

 * You always receive an email if a process execution has a bad ratio.

 * After receiving and email, the system reboots without updating the reboot count, and without updating the manifest, thus re-running the process execution. 

 * There are three new config variables to control outliers and ratios.

 * Needs testing on OpenBSD. I will do that next.

 * We can kill the old amperf check script in the warmup repo too, but we are likely to need it later for the 0.9 data prior to this change. Perhaps we should add a git tag that we can get back to later?

Unit tests implemented for the routines, but system testing is awkward, so I've done that manually by adding scratch code to make ratios appear bad every so often:

```

        # XXX testing code
        import random
        xxx = random.randint(0, 39)
        if xxx == 0:
            ratio = 0.5
            norm_aval = busy_threshold * 2
        elif xxx == 1:
            ratio = 1.5
            norm_aval = busy_threshold * 2
```

Sample debug log snippet:

```
...
[2017-07-13 11:17:03: DEBUG] sync disks...                                      
[2017-07-13 11:17:03: WARNING] SIMULATED: `time.sleep(30)` (--quick)            
[2017-07-13 11:17:03: INFO] stderr: [iterations_runner.py] iteration 1/5        
[2017-07-13 11:17:04: INFO] stderr: [iterations_runner.py] iteration 2/5        
[2017-07-13 11:17:05: INFO] stderr: [iterations_runner.py] iteration 3/5        
[2017-07-13 11:17:06: INFO] stderr: [iterations_runner.py] iteration 4/5        
[2017-07-13 11:17:07: INFO] stderr: [iterations_runner.py] iteration 5/5        
[2017-07-13 11:17:08: DEBUG] Finding outliers with window size 200              
[2017-07-13 11:17:08: WARNING] APERF/MPERF ratio badness detected               
  in_proc_iter=4, core=2, type=turbo, ratio=1.5                                 
[2017-07-13 11:17:08: DEBUG] Sending email to 'vext01' subject line '[krun:bencher2] Benchmark needs to be re-run: dummy:CPython:defaul
t-python'
[2017-07-13 11:17:08: INFO] Finished 'dummy(1000)' (default-python variant) under 'CPython'
[2017-07-13 11:17:08: INFO] Rebooting to re-run previous process execution      
[2017-07-13 11:17:08: WARNING] SIMULATED: reboot (--hardware-reboots is OFF)    
[2017-07-13 11:17:08: DEBUG] Simulated reboot with args: ../krun.py --debug debug --quick example.krun --no-tickless-check
[2017-07-13 11:17:08: DEBUG] Attached log file: example_20170713_111523.log     
...
```

Sample email:

```
Subject: [krun:bencher2] Benchmark needs to be re-run:
 dummy:CPython:default-python
From: noreply@bencher2
To: vext01@bencher2.soft-dev.org
Date: Thu, 13 Jul 2017 14:04:38 +0100 (BST)

Message from krun running on bencher2:

APERF/MPERF ratio badness detected
  in_proc_iter=1, core=2, type=throttle, ratio=0.5
  in_proc_iter=1, core=3, type=throttle, ratio=0.5

The process execution will be retried until the ratios are OK.
```

Comments?

### DIFF
--- a/examples/example.krun
+++ b/examples/example.krun
@@ -1,12 +1,12 @@
 import os
-from krun.vm_defs import (PythonVMDef, NativeCodeVMDef)
+from krun.vm_defs import (PyPyVMDef, NativeCodeVMDef)
 from krun import EntryPoint
 from krun.util import fatal
 from distutils.spawn import find_executable
 
 # For a real experiment you would certainly use absolute paths
-PYTHON27_BIN = find_executable("python2.7")
-if PYTHON27_BIN is None:
+PYPY_BIN = find_executable("pypy")
+if PYPY_BIN is None:
     fatal("Python binary not found in path")
 
 # Who to mail
@@ -35,8 +35,8 @@ VMS = {
         'variants': ['default-c'],
         'n_iterations': ITERATIONS_ALL_VMS,
     },
-    'CPython': {
-        'vm_def': PythonVMDef(PYTHON27_BIN),
+    'PyPy': {
+        'vm_def': PyPyVMDef(PYPY_BIN),
         'variants': ['default-python'],
         'n_iterations': ITERATIONS_ALL_VMS,
     }
@@ -45,13 +45,13 @@ VMS = {
 
 BENCHMARKS = {
     'dummy': 1000,
-    'nbody': 1000,
+    'nbody': 15,
 }
 
 # list of "bench:vm:variant"
 SKIP=[
     #"*:C:*",
-    #"*:CPython:*",
+    #"*:PyPy:*",
 ]
 
 N_EXECUTIONS = 2  # Number of fresh processes.
--- a/examples/example.krun
+++ b/examples/example.krun
@@ -73,3 +73,11 @@ TEMP_READ_PAUSE = 1
 
 # CPU pinning (off by default)
 #ENABLE_PINNING = False
+
+# Lower and upper bound for acceptable APERF/MPERF ratios
+AMPERF_RATIO_BOUNDS = 0.995, 1.005
+
+# Rough rate of change in APERF per second above which a core is considered busy.
+# For many machines this is simply the base clock frequency, but generally
+# speaking, is undefined, so you need to check on a per-machine basis.
+AMPERF_BUSY_THRESHOLD = 3.4 * 1000 * 1000 * 1000 / 1000  # 3.4 GHz / 1000
--- a/krun/amperf.py
+++ b/krun/amperf.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python2.7
+#
+# Copyright (c) 2017 King's College London
+# created by the Software Development Team <http://soft-dev.org/>
+#
+# The Universal Permissive License (UPL), Version 1.0
+#
+# Subject to the condition set forth below, permission is hereby granted to any
+# person obtaining a copy of this software, associated documentation and/or
+# data (collectively the "Software"), free of charge and under any and all
+# copyright rights in the Software, and any and all patent rights owned or
+# freely licensable by each licensor hereunder covering either (i) the
+# unmodified Software as contributed to or provided by such licensor, or (ii)
+# the Larger Works (as defined below), to deal in both
+#
+# (a) the Software, and
+# (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+# one is included with the Software (each a "Larger Work" to which the Software
+# is contributed by such licensors),
+#
+# without restriction, including without limitation the rights to copy, create
+# derivative works of, display, perform, and distribute the Software and make,
+# use, sell, offer for sale, import, export, have made, and have sold the
+# Software and the Larger Work(s), and to sublicense the foregoing rights on
+# either these or other terms.
+#
+# This license is subject to the following condition: The above copyright
+# notice and either this complete permission notice or at a minimum a reference
+# to the UPL must be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import sys
+import os
+from logging import debug
+
+
+class AMPerfRatios(object):
+    """Per-core {A,M}PERF analysis results"""
+
+    def __init__(self, vals, violations, busy_iters):
+        self.vals = vals  # list of ratios
+        self.violations = violations  # dict: type_string -> [iter_idxs]
+        self.busy_iters = busy_iters  # list of bool
+
+    def ok(self):
+        for iters in self.violations.itervalues():
+            if len(iters) > 0:
+                return False
+        return True
+
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(
+    os.path.abspath(__file__)))))
+
+
+def check_amperf_ratios(aperfs, mperfs, wc_times, busy_threshold, ratio_bounds):
+    results = []  # one AMPerfRatios instance for each core
+
+    for core_idx in xrange(len(aperfs)):
+        core_res = check_core_amperf_ratios(core_idx, aperfs[core_idx],
+                                            mperfs[core_idx], wc_times,
+                                            busy_threshold, ratio_bounds)
+        results.append(core_res)
+    return results
+
+
+def check_core_amperf_ratios(core_idx, aperfs, mperfs, wc_times, busy_threshold,
+                             ratio_bounds):
+    assert len(aperfs) == len(mperfs) == len(wc_times)
+    ratios = []
+    busy_iters = []
+    violations = {
+        "throttle": [],
+        "turbo": [],
+    }
+
+    itr = zip(xrange(len(aperfs)), aperfs, mperfs, wc_times)
+    for iter_idx, aval, mval, wctval in itr:
+        # normalise the counts to per-second readings
+        norm_aval = float(aval) / wctval
+        norm_mval = float(mval) / wctval
+        ratio = norm_aval / norm_mval
+        ratios.append(ratio)
+
+        if norm_aval > busy_threshold:
+            # Busy core
+            busy_iters.append(True)
+            if ratio < ratio_bounds[0]:
+                violations["throttle"].append(iter_idx)
+            elif ratio > ratio_bounds[1]:
+                violations["turbo"].append(iter_idx)
+        else:
+            busy_iters.append(False)
+    return AMPerfRatios(ratios, violations, busy_iters)
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -370,8 +370,16 @@ class ExecutionJob(object):
         if not dry_run:
             try:
                 measurements = util.check_and_parse_execution_results(
-                    stdout, stderr, rc)
+                    stdout, stderr, rc, self.sched.config)
                 flag = "C"
+            except util.RerunExecution as e:
+                subject = ("Benchmark needs to be re-run: %s (exec_idx=%s)" %
+                           (self.key, self.sched.manifest.next_exec_idx))
+                util.log_and_mail(mailer, warn, subject,
+                                  e.message, manifest=self.sched.manifest,
+                                  bypass_limiter=True)
+                measurements = self.empty_measurements
+                flag = "O"  # i.e. still outstanding
             except util.ExecutionFailed as e:
                 util.log_and_mail(mailer, error, "Benchmark failure: %s" %
                                   self.key, e.message,
@@ -379,16 +387,14 @@ class ExecutionJob(object):
                 measurements = self.empty_measurements
                 flag = "E"
 
-            if vm_def.instrument:
-                if flag != "E":
-                    instr_data = vm_def.get_instr_data()
-                    for k, v in instr_data.iteritems():
-                        assert len(instr_data[k]) == in_proc_iters
-                else:
-                    # The benchmark failed, so we assume the instrumentation
-                    # data to be compromised, and thus discard it.
-                    instr_data = {}
+            # Collect instrumentation data
+            if vm_def.instrument and flag == "C":
+                instr_data = vm_def.get_instr_data()
+                for k, v in instr_data.iteritems():
+                    assert len(instr_data[k]) == in_proc_iters
             else:
+                # The benchmark either failed, needs to be re-run, or had
+                # instrumentation turned off.
                 instr_data = {}
         else:
             measurements = self.empty_measurements
@@ -402,7 +408,7 @@ class ExecutionJob(object):
              (self.benchmark, self.parameter, self.variant, self.vm_name))
 
         # Move the environment log out of /tmp
-        if not dry_run:
+        if not dry_run and flag != "O":
             key_exec_num = self.sched.manifest.completed_exec_counts[self.key]
             util.stash_envlog(envlog_filename, self.sched.config,
                               self.sched.platform, self.key, key_exec_num)
@@ -517,6 +523,11 @@ class ExecutionScheduler(object):
             exec_start_time = time.time()
             measurements, instr_data, flag = job.run(self.mailer, self.dry_run)
             exec_end_time = time.time()
+
+            # Bail early if the process execution needs to be re-run.
+            if flag == "O":
+                info("Rebooting to re-run previous process execution")
+                util.reboot(self.manifest, self.platform, update_count=False)
 
             # Store new result.
             Results.ok_to_instantiate = True
--- a/scripts/calibrate_amperf_tolerance.py
+++ b/scripts/calibrate_amperf_tolerance.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python2.7
+#
+# Copyright (c) 2017 King's College London
+# created by the Software Development Team <http://soft-dev.org/>
+#
+# The Universal Permissive License (UPL), Version 1.0
+#
+# Subject to the condition set forth below, permission is hereby granted to any
+# person obtaining a copy of this software, associated documentation and/or
+# data (collectively the "Software"), free of charge and under any and all
+# copyright rights in the Software, and any and all patent rights owned or
+# freely licensable by each licensor hereunder covering either (i) the
+# unmodified Software as contributed to or provided by such licensor, or (ii)
+# the Larger Works (as defined below), to deal in both
+#
+# (a) the Software, and
+# (b) any piece of software and/or hardware listed in the lrgrwrks.txt file if
+# one is included with the Software (each a "Larger Work" to which the Software
+# is contributed by such licensors),
+#
+# without restriction, including without limitation the rights to copy, create
+# derivative works of, display, perform, and distribute the Software and make,
+# use, sell, offer for sale, import, export, have made, and have sold the
+# Software and the Larger Work(s), and to sublicense the foregoing rights on
+# either these or other terms.
+#
+# This license is subject to the following condition: The above copyright
+# notice and either this complete permission notice or at a minimum a reference
+# to the UPL must be included in all copies or substantial portions of the
+# Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+Draw plots to help you decide a suitable APERF/MPERF tolerance.
+
+usages:
+   calibrate_amperf_tolerance.py analyse <result-file> <aperf-estimate>")
+   calibrate_amperf_tolerance.py plot-hist <amstat-file>")
+   calibrate_amperf_tolerance.py plot-dropoff <amstat-file>")
+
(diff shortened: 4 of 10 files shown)


krun/amperf.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class AMPerfRatios(object):
+    def __init__(self, vals, violations):
+        self.vals = vals  # list of ratios
+        self.violations = violations  # dict: type_string -> [iter_idxs]
+
+    def ok(self):
+        for iters in self.violations.itervalues():
+            if iters:
```
ltratt: Does this mean if a violation happens in in-process iteration 0 it won't be caught?

snim2:
Maybe I am confused, but if we use this script to place outliers in the JSON files and this is a "no-op" on OpenBSD (see MM discussion) will that mean OpenBSD files miss the outliers?

ltratt:
A couple of dummy questions:

  * Why do we care about outliers? Even if an in-process iteration is an outlier, if it violates the APERF/MPERF ratio we should rerun.
  * Am I a dummy or can I not see where the system works out what the speedstep values are?

examples/example.krun
```diff
@@ -73,3 +73,7 @@ TEMP_READ_PAUSE = 1
 
 # CPU pinning (off by default)
 #ENABLE_PINNING = False
+
+AMPERF_RATIO_BOUNDS = 0.97, 1.03
+AMPERF_BUSY_THRESHOLD = 3.4 * 1024 * 1024 * 1024  # 3.6 GHz
```
snim2: `# 3.4GHz`?

examples/example.krun
```diff
@@ -73,3 +73,7 @@ TEMP_READ_PAUSE = 1
...
+OUTLIER_WINDOW_SIZE = 200
```
snim2: I don't think there is a solution for this (in this repo), but it isn't possible to ensure that this window size is used throughout.

krun/amperf.py
```diff
@@ -0,0 +1,110 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import sys
+import os
+from logging import debug
+
+
+class AMPerfRatios(object):
+    def __init__(self, vals, violations):
```
snim2: Extra space between class and init?

vext01:
> Maybe I am confused, but if we use this script to place outliers in the JSON files and this is a "no-op" on OpenBSD (see MM discussion) will that mean OpenBSD files miss the outliers?

The outliers generated "online" are currently discarded, so at present you still have to use the marker script.

> * We compute the outliers, and then discard them. We should consider (although in another PR) the ability to write the outliers into the results file directly, thus removing the need for the mark_outliers_in_json script.

krun/amperf.py
```diff
@@ -0,0 +1,110 @@
...
+    for i in xrange(1, migration_lookback + 1):
+        prior_aperf = aperfs[iter_idx - i]
+        if prior_aperf < busy_threshold:
+            return True
+    return False
+
+
+def check_amperf_ratios(aperfs, mperfs, wcts, busy_threshold, ratio_bounds,
+                        outliers):
+    results = []  # one AMPerfRatios instance for each core
```
snim2: Do we say `[]` or `list()` in the rest of the code base?

krun/amperf.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def check_core_amperf_ratios(core_idx, aperfs, mperfs, wcts, busy_threshold,
+                             ratio_bounds, outliers):
+    assert len(aperfs) == len(mperfs) == len(wcts)
+    ratios = []
+    violations = {
+        "throttle": [],
+        "turbo": [],
+    }
```
snim2: Likewise.

vext01:
> Why do we care about outliers? Even if an in-process iteration is an outlier, if it violates the APERF/MPERF ratio we should rerun.

We can use that logic, but it's not the same as what we did in the post-processing script.

krun/amperf.py
```diff
@@ -0,0 +1,110 @@
...
+    return False
+
+
+def check_amperf_ratios(aperfs, mperfs, wcts, busy_threshold, ratio_bounds,
+                        outliers):
+    results = []  # one AMPerfRatios instance for each core
+
+    for core_idx in xrange(len(aperfs)):
+        core_res = check_core_amperf_ratios(core_idx, aperfs[core_idx],
+                                            mperfs[core_idx], wcts,
```
snim2: What does `wcts` mean?

ltratt:
I don't think we ever wanted to emulate the post-processing script as such, because that was always a post-facto hack. The good thing we can do here is to catch all hardware affected in-process iterations *before* data-processing. Removing the outlier stuff will also make this PR a lot simpler I think?

vext01:
> Am I a dummy or can I not see where the system works out what the speedstep values are?

You supply a ratio tolerance manually for now via `AMPERF_RATIO_BOUNDS`.

If we would like Krun to figure it out automatically, I will need some more time.

Also note that is different to what we did in the paper, where we used a fixed ratio tolerance across all machines, whereas this would use a tolerance per-machine depending on the available speedstep increments.

examples/example.krun
snim2: test_amperf.py uses 3.6GHz though.

vext01:
> Removing the outlier stuff will also make this PR a lot simpler I think?

Certainly will.

krun/amperf.py
vext01: wallclock times. I can expand that out if you like.

examples/example.krun
vext01: 3.4 Hz * 1024 = 3.4 KHz * 1024 = 3.4 MHz * 1024 = 3.4 GHz

krun/amperf.py
snim2: Ah! I realise you don't want a really long name, but something a little clearer would be good.

examples/example.krun
vext01: This was what I was going to ask you on MM earlier. But if we kill the outliers stuff, it's moot.

krun/amperf.py
vext01: Not sure what you mean.

krun/amperf.py
vext01: It's a mix, but I really prefer `[]`

examples/example.krun
vext01: On our machines APERF happens to increment at the CPU base frequency (but we cannot rely on this in general)

vext01:
So if we are all in agreement that any bad ratio on a busy core -- outlier or not -- should trigger a re-run, then I can simplify this greatly. All in favour?

snim2:
+1

ltratt:
Yep.

krun/amperf.py
vext01: Do you mean, does `[0]` evaluate false?

```
In [1]: bool([0])
Out[1]: True

In [2]: if [0]:
   ...:     print("yehaw")
   ...:     
yehaw
```

I think we are OK, but equally, maybe more explicit code could help here too. What do you think?

examples/example.krun
ltratt: Sarah's pointing out that the comment says 3.*6*Ghz but the constant is 3.*4*Ghz.

krun/amperf.py
ltratt: OK, I misread this:

```
self.violations = violations  # dict: type_string -> [iter_idxs]
```
It mixes notations which confused me. What you're saying is it maps names to lists of indexes? In which case, yes, the code is safe because any non-empty list evaluates to `True`.

[Might be worth updating the comment with a clearer description of the type?]

krun/amperf.py
```diff
@@ -0,0 +1,103 @@
...
+            if iters:
+                return False
+        return True
+
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(
+    os.path.abspath(__file__)))))
+
+
+def recently_migrated(aperfs, iter_idx, busy_threshold, migration_lookback):
```
ltratt: This function never seems to be called?

krun/amperf.py
ltratt: A comment somewhere explaining the name would be fine IMHO. [I'm a fan of shorter names in general, provided, at definition, it's made clear what they mean.]

krun/scheduler.py
```diff
@@ -370,25 +370,30 @@ def run(self, mailer, dry_run=False):
         if not dry_run:
             try:
                 measurements = util.check_and_parse_execution_results(
-                    stdout, stderr, rc)
+                    stdout, stderr, rc, self.sched.config)
                 flag = "C"
+            except util.RerunExecution as e:
+                util.log_and_mail(mailer, warn,
+                                  "Benchmark needs to be re-run: %s" % self.key,
```
ltratt: Would it be better to say something like "APERF/MPERF violations detected: will rerun benchmark"?

krun/amperf.py
vext01: You are right, that's garbage.

krun/scheduler.py
vext01: At this level the exception message (which makes the message body) says why. The exception type itself is intentionally generic in case other re-run scenarios arise later.

examples/example.krun
vext01: ohhhh

krun/amperf.py
vext01: You mean add one? I tend not to.

vext01:
On OpenBSD (after tweaks):

```
...
[2017-07-13 15:49:56: WARNING] SIMULATED: `time.sleep(30)` (--quick)
[2017-07-13 15:49:57: INFO] stderr: [iterations_runner.py] iteration 1/5
[2017-07-13 15:49:57: INFO] stderr: [iterations_runner.py] iteration 2/5
[2017-07-13 15:49:57: INFO] stderr: [iterations_runner.py] iteration 3/5
[2017-07-13 15:49:57: INFO] stderr: [iterations_runner.py] iteration 4/5
[2017-07-13 15:49:57: INFO] stderr: [iterations_runner.py] iteration 5/5
[2017-07-13 15:49:57: WARNING] platform does not support APERF and MPERF counts. Not checking ratios.
[2017-07-13 15:49:57: INFO] Finished 'nbody(200)' (default-python variant) under 'CPython'
...
```

Still need to test more on Linux.

krun/scheduler.py
ltratt: Not sure I understand, but I'll take your word for it!

snim2:
The new commits look sensible to me

vext01:
I managed to cause some problems by fiddling the governor:

```
[2017-07-13 16:09:44: DEBUG] sync disks...
[2017-07-13 16:09:44: WARNING] SIMULATED: `time.sleep(30)` (--quick)
[2017-07-13 16:09:44: INFO] stderr: [iterations_runner.c] iteration 1/5
[2017-07-13 16:11:03: INFO] stderr: [iterations_runner.c] iteration 2/5
[2017-07-13 16:12:47: INFO] stderr: [iterations_runner.c] iteration 3/5
[2017-07-13 16:14:52: INFO] stderr: [iterations_runner.c] iteration 4/5
[2017-07-13 16:16:41: INFO] stderr: [iterations_runner.c] iteration 5/5
[2017-07-13 16:18:45: WARNING] APERF/MPERF ratio badness detected
  in_proc_iter=0, core=1, type=throttle, ratio=0.301292164458
  in_proc_iter=3, core=1, type=throttle, ratio=0.287855155116
  in_proc_iter=0, core=2, type=throttle, ratio=0.372847391035
  in_proc_iter=1, core=2, type=throttle, ratio=0.280866627931
  in_proc_iter=2, core=2, type=throttle, ratio=0.235294117635
  in_proc_iter=3, core=2, type=throttle, ratio=0.268778136622
  in_proc_iter=4, core=2, type=throttle, ratio=0.235294117586

The process execution will be retried until the ratios are OK.
```

Will now retry without the fiddling to see if the effect was present before.

krun/scheduler.py
vext01: Put differently. The exception type just says "we need to re-run", the exception message says why :)

ltratt: OK.

vext01:
Hrm, pretty much every execution gets restarted on bencher2, but then again, we already suspect bencher2 is wacky WRT power management.

I mean, look at the available frequencies:
```
  available frequency steps: 3.40 GHz, 3.40 GHz, 3.20 GHz, 3.00 GHz, 2.80 GHz, 2.70 GHz, 2.50 GHz, 2.30 GHz, 2.10 GHz, 1.90 GHz, 1.70 GHz, 1.50 GHz, 1.40 GHz, 1.20 GHz, 1000 MHz, 800 MHz
```

3.40 is listed twice, and each even has a separate bucket in the stats:

```
  cpufreq stats: 3.40 GHz:68.39%, 3.40 GHz:0.00%, 3.20 GHz:0.00%, 3.00 GHz:0.00%, 2.80 GHz:0.00%, 2.70 GHz:0.00%, 2.50 GHz:0.00%, 2.30 GHz:0.00%, 2.10 GHz:0.00%, 1.90 GHz:0.00%, 1.70 GHz:0.00%, 1.50 GHz:0.00%, 1.40 GHz:0.01%, 1.20 GHz:0.01%, 1000 MHz:0.02%, 800 MHz:31.55%  (602458)
```

No sir, I don't trust it ;)

I'd like to test on another Linux machine. Bencher 5 and 7 are in-use, bencher 6 is bsd, bencher 8 is debian 9 (which won't work without hacking rmsr). Is Krish done with bencher3 @ltratt?

vext01:
Might have to bite the bullet and make a debian 8 branch in rmsr, reserving the master branch for the newest debian version (9 at the moment).

vext01:
It occurs to me these may be outliers. Maybe that is why we are filtering them out...

ltratt:
You'll need to ask Krish, but I *think* so.

vext01:
If you look at #408 in the warmup repo, then we had stuff like:

```
Total violations (across all cores): 223
Of which outliers: 216
```

https://github.com/softdevteam/warmup_experiment/pull/408

I think we filter outliers, because they nearly always also have a bad a/mperf ratio... Perhaps to do with migration mid-pexec?

Not sure what the best plan of action is now...

vext01:
I guess let's ask Krish if he's done on bencher3 and see what happens there...

ltratt:
I'm confused. The warmup paper says that 0.3% of in-process iterations are outliers i.e. ~2200. So that's not the issue. But I'm not sure what you're saying the issue is. I can certainly believe that bencher2 is ropey for this sort of thing though -- it has turbo mode on, for example IIRC.

vext01:
Oh! Turbo mode is on. Well that probably explains it!

All of the bad ratios are >1.10.

I've mailed Krish.

vext01:
OK so I have not heard from Krish, but we can still go ahead and try to infer the tolerances automatically from the speedstep increments.

But before we do, I have a question about our logic. Do we assume that there is a one to one mapping between an iteration and the speedstep value it ran at?

It occurs to be that a in-process iteration can run at any (effective) clock speed between the highest and lowest frequencies by switching between speedstep frequencies mid-iteration.

E.g. Suppose we have a CPU with only three speedstep frequencies:
 * 1.2 GHz
 * 1 GHz (base)
 * 800 MHz

What is a suitable ratio bounds? `0.8 <= 1 <= 1.2`?

But if an iteration ran at 1 GHz for half of it's duration, then at 0.8 MHz for the remainder, then the ratio would be 0.9. Is that OK?

I'm starting to wonder if the speedstep frequencies are a red herring. Should we use a fixed bound regardless?

ltratt:
The speedstep frequencies are not a red herring, but you're absolutely right that there's only so much they can help us detect, given that we need to have a tolerance in our comparison of readings (footnote 5 in the paper says exactly this!). The thing is that there is no other way short of sampling APERF/MPERF more frequently (which would bring other problems that we don't want).

In an ideal world, given your example, I think we probably want bounds 50% of the way between the lower and upper speedstep values. i.e. 0.9 & 1.1 in this example [*]. Note that we had looser bounds before where we couldn't rerun things. Given that we can rerun, we can probably afford to be more exacting. Edd, could you maybe quickly check how many pexecs on the 0.8 or 0.9 runs would be affected if we used these bounds?

[*] That said, 10% seems huge, so maybe we want to cap it at (say) 3%. There again, I don't think real chips have such huge speedstep jumps, but maybe I'm wrong.

vext01:
10% does seem big. And 3% consistently (currently) makes more sense to me, but I think I would like to mull it over a bit more.

An earlier though I had was to allow an iteration to execute for no more than X% (e.g. 25%) of it's execution in a lower/higher speedstep state. So compute this bound using the frequencies either side of the base frequency. But I'm not totally sure that makes sense. 

What bugs me is that the range of ratios we may see are almost continuous (within the constraints of the slowest/fastest frequencies) regardless of the available speedstep increments, if you see what I mean.

Let me think more...

vext01:
Ladies and gents, time to review `scripts/calibrate_amperf_tolerance.py`.

examples/example.krun
```diff
@@ -73,3 +73,9 @@ TEMP_READ_PAUSE = 1
 
 # CPU pinning (off by default)
 #ENABLE_PINNING = False
+
+# Lower and upper bound for acceptable APERF/MPERF ratios
+AMPERF_RATIO_BOUNDS = 0.97, 1.03
```
vext01: These should be updated.

vext01: Maybe in another PR? Along with switching to bencher8 for payload copies.

ltratt: Probably, although for random people's machines, we probably shouldn't set them *too* tight because they may just end up looping forever. Maybe 0.0985 and 1.1015 or something like that?

examples/example.krun
```diff
@@ -73,3 +73,9 @@ TEMP_READ_PAUSE = 1
...
+
+# Rough change in APERF per second above which a core is considered busy.
+AMPERF_BUSY_THRESHOLD = 3.4 * 1024 * 1024 * 1024 / 1000  # 3.4 GHz / 1000
```
vext01: This needs to be per-machine, but likewise, maybe in another PR?

examples/example.krun
vext01: Sorry, these are examples, disregard.

examples/example.krun
ltratt: This is the example file so I don't see a problem...

vext01: disregard this too.

vext01: I was aware that for the main experiment config, we need to add these directives, but apparently got mixed up. Yeesh. Sorry about that.

scripts/calibrate_amperf_tolerance.py
```diff
@@ -0,0 +1,246 @@
...
+from collections import OrderedDict
+
+sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
+from krun.amperf import check_amperf_ratios
+
+
+TOLERANCES = [(i + 1) * 0.0005 for i in xrange(80)]
+
+
+def compute(jsn, busy_thresh, output_filename):
```
snim2: Not a great function name!

vext01: The function does a few different things, hence it's a bit generic. Any suggestions?

scripts/calibrate_amperf_tolerance.py
```diff
@@ -0,0 +1,246 @@
...
+                    bad_iters.update(idxs)
+            if len(bad_iters) > 0:
+                num_bad_pexecs += 1
+            num_bad_iters += len(bad_iters)
+    return total_pexecs, total_iters, num_bad_pexecs, num_bad_iters
+
+
+def _collect_busy_ratios(jsn, busy_thresh):
+    ratios = []
+    for bench in jsn["wallclock_times"].iterkeys():
```
snim2: Isn't this the same as saying `for bench in jsn["wallclock_times"]:`?

scripts/calibrate_amperf_tolerance.py
```diff
@@ -0,0 +1,246 @@
...
+    ax2.plot(good_iters_xs, good_iters_ys)
+    ax2.set_title('Good In-Process Iterations')
+    ax2.set_xlabel("Tolerance")
+    ax2.set_ylabel("% iters")
+    ax2.set_ylim([0, 102])
+    ax2.grid(color='r', linestyle='--')
+    ax2.set_yticks([(x + 1) * 5 for x in xrange(20)])
+
+    print("Plotting to %s" % output_filename)
+    plt.savefig(output_filename)
```
snim2: This function uses a mixture of single and double quotes.

scripts/calibrate_amperf_tolerance.py
vext01: Yes, but it always catches me out, so I like to be explicit.

scripts/calibrate_amperf_tolerance.py
snim2: Well, the problem is that any function computes.  So maybe split the function so you can name it better, or say what it specifically computes. How about `apply_tolerances()`?

scripts/calibrate_amperf_tolerance.py
snim2: I'm not sure the rest of the code base uses `iterkeys()`, FWIW

scripts/calibrate_amperf_tolerance.py
vext01: thing is, it also computes the busy ratios.

`analyse_amperf` ok?

snim2: sounds ok, sure

vext01:
That should address your comments Sarah.

vext01:
I think I should re-test the main Krun part now, since I changed the amperf utility API.

vext01:
Working OK on bencher 2 where every pexec is rejected due to turbo mode. Taking this to bencher3 now.

vext01:
OK, I think this is ready for some (serious) squashing.

I tested on bencher5, running small example experiments with n-body, narrowing the bounds each time. Eventually (+/- 0.002) I started to get re-runs. The email now includes a process execution index (from the manifest) so we can see if the same pexec is rerunning forever.

Please check the last few commits and let's get an experiment going.

ltratt:
Looks OK to me. If Sarah's OK with it, then please squash and I'll merge.

snim2:
+1 from me...

ltratt:
OK, squash away!

vext01:
dont merge just yet please

scripts/calibrate_amperf_tolerance.py
vext01: I could have swore I fixed this....

vext01:
Forced the fix back.

OK to merge?

ltratt:
Does this need squashing or merge as-is?

vext01:
It's already squashed. We had some 30 commits before.

ltratt:
I was a bit confused by the "is it all OK" stuff. Anyway, glad to know it is! Will merge once the tests have run.

vext01:
I suppose i should squash this in too. Is that OK?

ltratt:
We should squash / rename that last commit.

vext01:
done.

PR #318, opened 2017-07-20
Ensure post-exec commands are run (once) when a rerun is requested.

vext01:
As the title says, and also fix some racy code I found whilst in there.

Opted to duplicate the call to run the post-exec commands, rather than try to thread it through to the `finally` block. It just looked too error prone.

Tested on bencher2, but will test again tomorrow. Please don't merge just yet.

Looks OK?

### DIFF
--- a/krun/util.py
+++ b/krun/util.py
@@ -486,7 +486,14 @@ def _do_reboot(platform):
         os.execv(args[0], args)  # replace myself
         assert False  # unreachable
     else:
-        subprocess.call(platform.get_reboot_cmd())
+        rc = subprocess.call(platform.get_reboot_cmd())
+        if rc != 0:
+            fatal("Failed to reboot with: %s" % platform.get_reboot_cmd())
+        else:
+            debug("hard exit")
+            # Use _exit() to stop without raising SystemExit, otherwise more
+            # Python code may run.
+            os._exit(0)
 
 
 def reboot(manifest, platform, update_count=True):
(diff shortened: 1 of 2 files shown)


krun/util.py
```diff
@@ -486,7 +486,18 @@ def _do_reboot(platform):
...
+        if rc != 0:
+            fatal("Failed to reboot with: %s" % platform.get_reboot_cmd())
+        else:
+            debug("hard exit")
+            # Really exit right now. No more Python code should execute. A regular
+            # sys.exit() would raise a 'SystemExit' which unwinds the stack
+            # including running 'finally' blocks, which is undesirable if you want
+            # to reboot. This is especially acute in the case of reboot, since the
+            # amount of 'finally' code that executes is racy depending on how quick
+            # the system reboots.
```
ltratt: OK I eventually understood this comment, but it's really long-winded. Can we shorten and clarify?

ltratt:
Looks OK except that one comment. Though I wonder if any parts of Krun implicitly relied on code executing after `reboot` is called? Should we check the call sites and check that we weren't expecting `finally` and friends to be called?

vext01:
Made the comment (much) shorter.

I'm absolutely certain that I did not intend code to run further after initiating a reboot.

ltratt:
OK, squash and I'll merge?

vext01:
squashed.

PR #320, opened 2017-07-26
Use syscalls

vext01:
Here is the initial Krun support for the custom kernel.

I need to fix the tests, but we can start reviewing.

Please don't merge yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,61 +58,14 @@ steps:
   * Add `intel_pstate=disable` to `GRUB_CMDLINE_LINUX_DEFAULT`
   * Run `sudo update-grub`
 
-### Tickless Mode Linux Kernel
+### The Krun Linux Kernel
 
-The Linux kernel can run in ``tickless'' configurations, where under certain
-conditions regular tick interrupts can be avoided for a subset of logical CPUs.
-More info here:
-https://www.kernel.org/doc/Documentation/timers/NO_HZ.txt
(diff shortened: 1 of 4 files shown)


ltratt:
This looks good to me. With the change of pushing the `for` loop into the kernel have the ratios got tighter? [My intuition is that they should have... but I've been wrong before!]

vext01:
They are still pretty tight, yeah. I'll give you a max/min once my run finishes.

ltratt:
I think they should be *tighter* than before, since we've now removed more syscall overhead.

vext01:
For my smaller run, which ran several hours, the max and min ratios we saw was:
(0.9982062780269059, 1.0008437801350047)

Pretty good going if you ask me.

ltratt:
Absolutely! 

I think we're ready to squash both PRs and merge?

vext01:
note that those are extremes.

Full log here:
http://theunixzoo.co.uk/random/amlog

ltratt:
Don't worry, I knew what you meant.

vext01:
I'd just like to plot what i collected so far to check the core-cycles.

ltratt:
OK.

vext01:
http://theunixzoo.co.uk/random/plots.pdf

Looks fine to me, but I'd like your opinion. Note these are only 200 iters long.

vext01:
splat

PR #321, opened 2017-07-27
Remove old rmsr submodule now that we use syscalls.

vext01:
Remove old rmsr submodule now that we use syscalls.

Followed this SO answer: https://stackoverflow.com/questions/1260748/how-do-i-remove-a-submodule/36593218#36593218


vext01:
Ah, I also need to hack make a bit. Please don't merge just yet.

ltratt:
OK.

vext01:
There we go.

PR #323, opened 2017-08-07
Fix crash

vext01:
This fixes the crash we saw yesterday:

```
Message from krun running on bencher5:

Fatal Krun error: 'NoneType' object has no attribute 'eta_estimates'
  File "krun/krun.py", line 231, in main
  File "krun/krun.py", line 357, in inner_main
    sched.run()
  File "/home/kruninit/warmup_experiment/krun/krun/scheduler.py", line 531, in run
    extra_env=self._make_post_cmd_env(results)
  File "/home/kruninit/warmup_experiment/krun/krun/scheduler.py", line 473, in _make_post_cmd_env
    eta_val = self.get_overall_time_estimate_formatter(results).finish_str
  File "/home/kruninit/warmup_experiment/krun/krun/scheduler.py", line 462, in get_overall_time_estimate_formatter
    self.get_estimated_overall_duration(results))
  File "/home/kruninit/warmup_experiment/krun/krun/scheduler.py", line 447, in get_estimated_overall_duration
    per_exec = self.get_estimated_exec_duration(key, results)
  File "/home/kruninit/warmup_experiment/krun/krun/scheduler.py", line 438, in get_estimated_exec_duration
    previous_exec_times = results.eta_estimates.get(key)
```

I'm not sure why my testing didn't find this when I was implementing the re-run logic, but anyway, I've re-tested carefully and I think this is good.

The first commit is the fix, the message explains what went wrong before.

The second commit is just some defensive programming.

OK?

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -524,6 +524,11 @@ class ExecutionScheduler(object):
             measurements, instr_data, flag = job.run(self.mailer, self.dry_run)
             exec_end_time = time.time()
 
+            # Only now is it OK to load the results file into memory.
+            Results.ok_to_instantiate = True
+            results = Results(self.config, self.platform,
+                              results_file=self.config.results_filename())
+
             # Bail early if the process execution needs to be re-run.
             if flag == "O":
                 util.run_shell_cmd_list(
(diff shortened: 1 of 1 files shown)


ltratt:
I'm fine with this except `assert False` is an icky idiom IMHO (asserts are removed by `-O` for example): we should do something like `raise RuntimeError`.

vext01:
I didn't know about `-O` (but it's quite amusing somehow in the case of CPython).

How about a call to `fatal` instead?

ltratt:
I'd keep it simple: since it's not *that* likely to happen, `raise RuntimeError` or equivalent seems easy enough.

vext01:
Here we go. See also #324 as there are quite a few assertions in-tree (I don't want to fix those immediately though).

ltratt:
OK, squash and I'll merge.

vext01:
splat.

vext01:
Fixed typo.

PR #326, opened 2017-08-24
Update kernel docs

vext01:
Update docs in light of this morning's `rdpmc` conversation and clarify a bit.

(No squash required)

OK?


ltratt:
I'm fine with this, but will leave @fsfod to do the final check and merge (etc.).

PR #327, opened 2017-08-31
Run on stock kernel

vext01:
**Do not merge yet please. Review only**

This adds the ability to run Krun on a stock Linux kernel.

It works as follows:

 * The user builds libkrun with `NO_MSRS=1` and this causes libkrun to *not* expect the custom kernel interface to exist. Also `krun_num_cores()` returns 0.

 * The above change alone would cause failures, since `platform.py` contains heuristics duplicated from the libkrun build system to decide how many cores worth of data to expect. Since we have not changed those heuristics in the Python code, libkrun sends 0 cores of msr data, but the Python part expects (e.g.) 4.
    * On the libkrun side, see: https://github.com/softdevteam/krun/blob/3a6bc633235e1df0799b4b228e8baaf0195b9359/libkrun/Makefile#L9 and https://github.com/softdevteam/krun/blob/3a6bc633235e1df0799b4b228e8baaf0195b9359/libkrun/libkruntime.c#L286
    * On the python side, see: https://github.com/softdevteam/krun/blob/3a6bc633235e1df0799b4b228e8baaf0195b9359/libkrun/libkruntime.c#L296

 * Rather than update the duplicated heuristics on the Python side, I've propose we use cffi to ask libkrun how many cores it will send, thus removing the need for duplication of logic altogether. This also simplifies the code a bit.

* As a downside, we have to call `krun_init()` which does some allocation: https://github.com/softdevteam/krun/blob/3a6bc633235e1df0799b4b228e8baaf0195b9359/libkrun/libkruntime.c#L296 The allocations are small and (for any given machine) fixed-size. How do you feel about that? It is unlikely to affect things in the grand scheme of things, but I felt it worth mentioning.

The NO_MSRS cases is tested on bencher8, but I'd like to test on a machine with the krun kernel installed, to see if I broke anything. Later I will set up a spare machine I have and try.

There are also a couple of doc fixes/improvements.


ltratt:
I'm fine with this. BTW, it's not clear to me that `krun_init` allocates memory, other than using stack space. Maybe I'm missing out on something?

vext01:
It calls `calloc()` via `krun_xcalloc()`.

Maybe you confused `calloc` with `alloca`?

ltratt:
Ah, no, I didn't realise that function allocates. OK, makes sense (and I'm still OK with it).

vext01:
Cool. I'm going to implement another fix before I head home to test this properly. Expect a further PR :)

snim2:
This all sounds like a good idea!

vext01:
Tested on bencher8 using the krun kernel. looks ok to me.

README.md
```diff
@@ -67,9 +70,23 @@ MSRs (sadly `IA32_APERF` or `IA32_MPERF` cannot be read from user-space via
 `rdpmc` and `rdmsr` is strictly a ring 0 operation). The kernel must also be
 configured to be tickless on all CPU cores except the boot core.
 
-Instrcutions and source code can be found here:
+Instructions and source code can be found here:
 https://github.com/softdevteam/krun-linux-kernel
 
+#### Benchmarking on a Stock Linux Kernel
+
+You can run Krun on a stock Linux Kernel, but you Krun will be unable to
```
snim2: Typo: `but you Krun`

vext01:
Fixed. OK to squash?

snim2:
Yep, no more comments from me!

vext01:
Ready to merge.

PR #329, opened 2017-08-31
--hardware-reboots is not a developer switch.

### LINKED ISSUE #328: --hardware-reboots is listed as a developer flag.
In the docs and the sources (and usage?), `--hardware-reboots` is listed as a developer flag.

This is not so :)

snim2:
Fixes #328

### DIFF
--- a/README.md
+++ b/README.md
@@ -341,7 +341,7 @@ easier.
 
   * `--no-user-change`: Without this flag, For each process execution, Krun
     will use a fresh user account called 'krun'. This involves deleting any
-    exising user account (with `userdel -r`) and creating a new user account
+    existing user account (with `userdel -r`) and creating a new user account
     (with `useradd -m`).  This switch disables the use of a fresh user account,
     meaning that `userdel` and `useradd` are not invoked, nor does Krun switch
     user; the user Krun was invoked with is used for benchmarking.
@@ -354,11 +354,20 @@ easier.
 
   * `--no-pstate-check`: Do not crash out if Intel P-states are not disabled.
 
-  * `--hardware-reboots`: Restart physical hardware before each benchmark
-    execution.
+## Benchmarking for reliable results
 
-Note that you should not collect results intended for publication with
-development switches turned on.
+You should not collect results intended for publication with development switches
+turned on.
+
+We also recommend that for 'real' benchmarking you turn on the `--hardware-reboots`
+switch to restart physical hardware before each benchmark execution.
+You will also need to ensure that Krun is restarted once the machine has
+rebooted. Depending on your OS, this might best be done by inserting a line into
+`/etc/rc.local` (that line should also pass in the `--hardware-reboots` switch
+to Krun). This will suppress some emails that Krun sends out.
+
+The `krun/etc` directory contains example `rc.local` files for some Linux
+distributions and OpenBSD, which go with the examples here.
 
 ## Unit Tests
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -354,11 +354,11 @@ easier.
 
   * `--no-pstate-check`: Do not crash out if Intel P-states are not disabled.
 
-  * `--hardware-reboots`: Restart physical hardware before each benchmark
-    execution.
-
 Note that you should not collect results intended for publication with
-development switches turned on.
+development switches turned on. We also recommend that for 'real'
```
vext01: I think this is a litle hidden here. Shall we move it into `Step 6` with some notes on trying first without reboots? We also seem to not mention rc.local anywhere.

Sorry, I just realised that.

vext01:
Just that one comment.

README.md
snim2: OK, sounds like a good idea...

snim2:
Ok, if you are both happy with those commits, this will need squashing.

README.md
```diff
@@ -354,25 +354,32 @@ easier.
...
-  * `--hardware-reboots`: Restart physical hardware before each benchmark
-    execution.
-
 Note that you should not collect results intended for publication with
 development switches turned on.
 
+We also recommend that for 'real' benchmarking you turn on the `--hardware-reboots`
+switch to restart physical hardware before each benchmark execution.
+You will also need to ensure that Krun is restarted once the machine has
+rebooted. You can do this by hand, or by using the boot configuration provided
```
ltratt: What's a "boot configuration"?

snim2: Good question, this was taking from an earlier commit, I guess it means rc.local?

ltratt: I sort-of guessed that... but it's a nonsense term, I think. Or, at least, I haven't heard it. So we probably need to expand it into more familiar English. I would personally suggest just saying "Depending on your OS, this might best be done by inserting a line into `/etc/rc.local`". I also wonder if we need to tell them *what* that line is, because I vaguely remember that you have to add some sort of "Krun is being started in resume mode" switch?

vext01: It's more than one line I'm afraid. The rc.local snippet deals with adding the switch you mention (`--resume`).

ltratt: OK, so this may suggest that we need to do a bit more rejigging of the instructions?

vext01: Here's the example rc.local btw: https://github.com/softdevteam/krun/blob/master/etc/rc.local.linux

vext01: Yeah. Basically the instructions don't mention `rc.local` at all.

ltratt: OK. We can either fix that in this PR or open another PR. Might as well do it here...

snim2: Maybe I'm being daft, but this conversation has confused me. We removed `--resume` and `--started-by-init` in a previous PR, when we added `--hardware-reboots` and simplified the CLI switches (at least, that's how I recall it, and I can't find those other switches in `--help`!). So, do you just want me to clarify the wording above, to remove "boot configuration", or is there much more to do here?

vext01: Sarah is right. Those flags are now gone. Krun now uses the existence of a manifest file to decide if it is resuming or not.

I don't think you need to mention these implementation details.

ltratt: Right, so all we need to do, it sounds like, is make it clearer that you need to add *something* to `/etc/rc.local` or equivalent. Since this is OS / distribution dependent, maybe we could add a bullet list or subsection saying something like "on OpenBSD you might do this: XXX; on Debian this: XXX; on systemd: XXX YYY ZZZ AAA BBB CCC DDD" [sorry, I couldn't resist].

snim2:
OK, is this ready to squash?

ltratt:
Works for me.

README.md
```diff
@@ -354,25 +354,32 @@ easier.
 
   * `--no-pstate-check`: Do not crash out if Intel P-states are not disabled.
 
-  * `--hardware-reboots`: Restart physical hardware before each benchmark
-    execution.
-
 Note that you should not collect results intended for publication with
 development switches turned on.
 
+We also recommend that for 'real' benchmarking you turn on the `--hardware-reboots`
```
vext01: We were going to move this explanation out of the `Development and Debug Switches` section, weren't we? I'd recommend moving it to where the user is running the example scenario.

snim2:
Howzat?

vext01:
I can live with that. Laurie?

ltratt:
Me too. If you squash, I'll merge.

snim2:
Squashed.

PR #330, opened 2017-08-31
Unprotect dmesg deb9

vext01:
As of Linux 4.8, the dmesg buffer is only available to root unless a procfs file contains 1.

This makes Krun set the file to 1 if it exists.

Tested (up until the crash you will get due to #327 not yet being merged) on bencher8 (before this change Krun would crash much earlier).

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -688,6 +688,7 @@ class LinuxPlatform(UnixLikePlatform):
     ASLR_MODE = 2
     CSET_CMD = "/usr/bin/cset"
     USER_CSET_DIR = "/cpusets/user"
+    RESTRICT_DMESG_FILE = "/proc/sys/kernel/dmesg_restrict"
 
     # Expected tickless kernel config
     #
@@ -829,6 +830,7 @@ class LinuxPlatform(UnixLikePlatform):
         if not self.no_tickless_check:
             self._check_tickless_kernel()
         self._check_aslr_enabled()
+        self._check_dmesg_unrestricted()
 
     def _find_virt_what(self):
         debug("Check virt-what is installed")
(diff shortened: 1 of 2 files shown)


snim2:
lgtm

vext01:
Any comment laurie?

krun/platform.py
```diff
@@ -875,6 +877,25 @@ def _check_cset_shield(self):
...
             debug(out)  # cset is quite chatty on stdout
 
+    def _check_dmesg_unrestricted(self):
+        debug("Checking if dmesg buffer is restricted")
+
+        if os.path.exists(LinuxPlatform.RESTRICT_DMESG_FILE):
+            cmd = "%s cat %s" % (self.change_user_cmd,
+                                 LinuxPlatform.RESTRICT_DMESG_FILE)
+            out, _, _ = run_shell_cmd(cmd)
+            if int(out) == 0:
```
ltratt: These two lines might be simpler as `if int(run_shell_cmd(cmd)[0])) == 0`? [I'm also curious if we should catch the "file doesn't exist and / or returns something other than an integer" case...]

vext01: The file must exist due to a earlier check. If it aint an int, an exception is raised, which is caught and emailed.

krun/platform.py
```diff
@@ -875,6 +877,25 @@ def _check_cset_shield(self):
             out, _, _ = run_shell_cmd(cmd)
             debug(out)  # cset is quite chatty on stdout
 
+    def _check_dmesg_unrestricted(self):
+        debug("Checking if dmesg buffer is restricted")
+
+        if os.path.exists(LinuxPlatform.RESTRICT_DMESG_FILE):
```
vext01: Here's the check @ltratt

ltratt:
OK, so this needs rebasing against master (and maybe squashing, but that's up to you), then I can merge.

vext01:
Synced and squashed.

PR #332, opened 2017-09-07
Tolerate one un-named sensor chip on Linux.

vext01:
This change allows Krun to tolerate at most one un-named sensor chip.

I've commented liberally, as the logic is a bit fiddly.

Tested on a machine with no un-named chips, and on Travis, which has no sensors at all. I don't have a machine with un-named sensors to test on.

Any comments?

@snim2 @fsfod  can you try this branch and report back?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -52,7 +52,7 @@ from krun import ABS_TIME_FORMAT
 from krun.util import (fatal, run_shell_cmd, log_and_mail,
                        PLATFORM_SANITY_CHECK_DIR)
 import krun.util as util
-from logging import warn, debug
+from logging import warn, debug, info
 from abc import ABCMeta, abstractmethod, abstractproperty
 from krun.env import EnvChangeSet, EnvChange, EnvChangeAppend
 from krun.vm_defs import BENCHMARK_USER
@@ -91,6 +91,8 @@ class BasePlatform(object):
         self._starting_temperatures = {}  # accessed via property
         self.temperature_thresholds = {}
         self.find_temperature_sensors()
+        if self.get_num_temperature_sensors() == 0 and not self.is_virtual():
+            fatal("No usable temperature sensors!")
 
         self.last_dmesg = None
 
@@ -405,6 +407,11 @@ class BasePlatform(object):
         remnants of any existing krun user."""
         pass
 
+    @abstractmethod
+    def get_num_temperature_sensors(self):
+        """Get the number of temperature sensors"""
+        pass
+
 
 class UnixLikePlatform(BasePlatform):
     """A UNIX-like platform, e.g. Linux, BSD, Solaris"""
@@ -540,6 +547,9 @@ class OpenBSDPlatform(UnixLikePlatform):
 
                 sensors.append(elems[0].strip())
         self.temp_sensors = sensors
+
+    def get_num_temperature_sensors(self):
+        return len(self.temp_sensors)
 
     def bench_env_changes(self):
         # Force malloc flags
@@ -706,6 +716,8 @@ class LinuxPlatform(UnixLikePlatform):
     CSET_CMD = "/usr/bin/cset"
     USER_CSET_DIR = "/cpusets/user"
     RESTRICT_DMESG_FILE = "/proc/sys/kernel/dmesg_restrict"
+    UNKNOWN_SENSOR_CHIP_NAME = "__krun_unknown_chip_name"
+    TEMP_SENSOR_INPUT_GLOB = "temp[0-9]*_input"
 
     # Expected tickless kernel config
     #
@@ -734,9 +746,9 @@ class LinuxPlatform(UnixLikePlatform):
 
     def __init__(self, mailer, config):
         self.temp_sensor_map = None
+        self.virt_what_cmd = self._find_virt_what()
         UnixLikePlatform.__init__(self, mailer, config)
         self.num_cpus = self._get_num_cpus()
-        self.virt_what_cmd = self._find_virt_what()
 
     def _fatal_kernel_arg(self, arg, prefix, suffix):
         """Bail out and inform user how to add a kernel argument"""
@@ -755,32 +767,77 @@ class LinuxPlatform(UnixLikePlatform):
               "Set `%s` in the kernel arguments.\n"
               "%s" % (prefix, arg, suffix))
 
-    def find_temperature_sensors(self):
-        """Detect sensors using the hwmon sysfs framework
+    @staticmethod
+    def _collect_temperature_sensor_globs():
+        """Collects the sysfs info needed to enumerate temperature sensors.
+        Separate to enable test mocking"""
 
-        The sensors move between reboots, so we have to roll our own sensor
-        identifier. We are not fussy, and use as many sensors as we can find."""
-
-        # maps our identifier to actual sysfs filename
-        sensor_map = {}
-
+        res = {}
         hwmons = glob.glob(LinuxPlatform.HWMON_CHIPS_GLOB)
         for chip in hwmons:
             name_path = os.path.join(chip, "name")
-            with open(name_path) as fh:
-                chip_name = fh.read().strip()
+            try:
+                with open(name_path) as fh:
+                    chip_name = fh.read().strip()
+            except IOError:
+                chip_name = None
 
-            inputs_glob = os.path.join(chip, "temp[0-9]*_input")
-            chip_sensors = glob.glob(inputs_glob)
+            # Check our super-obscure un-named chip string doesn't arise in the wild!
+            assert chip_name != LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME
 
-            for sensor in chip_sensors:
-                key = "%s:%s" % (chip_name, os.path.basename(sensor))
-                assert key not in sensor_map  # two chips with the same name?
-                sensor_map[key] = sensor
+            inputs_glob = os.path.join(chip, LinuxPlatform.TEMP_SENSOR_INPUT_GLOB)
+            res[chip] = chip_name, list(glob.glob(inputs_glob))
+        return res
+
+    def find_temperature_sensors(self):
+        """Detect sensors using the hwmon sysfs framework
+
+        It's unclear as to whether the sensors may move between reboots, so we
+        have to roll our own sensor identifiers. We are not fussy, and use as
+        many sensors as we can find"""
+
+        # maps our identifier to actual sysfs filename
+        sensor_tree = {}  # chip_id -> sensor_id -> sysfs_path
+        # chip IDs are pairs: (chip_name, num_sensors)
+        duplicate_chip_ids = set()
+
+        for chip, (chip_name, chip_sensors) in \
+                LinuxPlatform._collect_temperature_sensor_globs().iteritems():
+            if chip_name is None:
+                debug("Un-named temperature sensor chip found.")
+                debug("Naming the chip: '%s'" % LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME)
+                chip_name = LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME
+
+            chip_id = (chip_name, len(chip_sensors))
+            if chip_id in sensor_tree:
+                info("Found duplicate chips named '%s' with %s temperature sensor(s)" % chip_id)
+                # Ignore all chips with this ID
+                duplicate_chip_ids.add(chip_id)
+                del sensor_tree[chip_id]
+                continue
+            elif chip_id in duplicate_chip_ids:
+                info("Found another chip named '%s' with %s temperature sensor(s)" % chip_id)
+                continue
+
+            sensor_tree[chip_id] = {}
+            for sysfs_path in chip_sensors:
+                sensor_id = (chip_id[0], str(chip_id[1]), os.path.basename(sysfs_path))
+                assert sensor_id not in sensor_tree[chip_id]
+                sensor_tree[chip_id][sensor_id] = sysfs_path
+
+        # flatten the tree into a dict: sensor_id -> sysfs_path
+        sensor_map = {}
+        for chip_id, sensors in sensor_tree.iteritems():
+            for sensor_id, sysfs_path in sensors.iteritems():
+                assert sensor_id not in sensor_map
+                sensor_map[":".join(sensor_id)] = sysfs_path
 
         debug("Detected temperature sensors: %s" % sensor_map)
         self.temp_sensors = sensor_map.keys()
         self.temp_sensor_map = sensor_map
+
+    def get_num_temperature_sensors(self):
+        return len(self.temp_sensors)
 
     def _get_num_cpus(self):
         # most reliable method generic to all Linux
--- a/krun/tests/test_linuxplatform.py
+++ b/krun/tests/test_linuxplatform.py
@@ -234,8 +234,8 @@ class TestLinuxPlatform(BaseKrunTest):
 
         for sid in temps.iterkeys():
             elems = sid.split(":")
-            assert len(elems) == 2
-            assert elems[1].endswith("_input")
+            assert len(elems) == 3
+            assert elems[2].endswith("_input")
 
         # check temperature readings are within reasonable parameters
         assert all([type(v) == float for v in temps.itervalues()])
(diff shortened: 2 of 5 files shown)


fsfod:
This is what i get
[2017-09-07 16:32:22: DEBUG] Detected temperature sensors: {'coretemp:temp2_input': '/sys/class/hwmon/hwmon2/temp2_input', 'coretemp:temp3_input': '/sys/class/hwmon/hwmon2/temp3_input', 'acpitz:temp1_input': '/sys/class/hwmon/hwmon0/temp1_input', 'coretemp:temp1_input': '/sys/class/hwmon/hwmon2/temp1_input'}

vext01:
So looks like another case of "chip with no names, but no temp sensor either". Can you confirm @fsfod? The sensor dir with the missing `name` file, is it also missing a `temp` file?

fsfod:
I don't see any temp file in the hwmon directory for it.

vext01:
OK, so that explains it. Does this fix your crash also?

fsfod:
Yes it didn't crash for me.

vext01:
OK, so the only thing that worries me a little now is that we don't have a machine where it legitimately crashes. So we must check that my logic is good my hand.

I'll pull in @ltratt for another review just in case.

ltratt:
I think this should also be reviewed by a Linux user.

vext01:
Totally. Did you have any comments on the code @snim2 @fsod?

krun/platform.py
```diff
@@ -761,23 +762,61 @@ def find_temperature_sensors(self):
...
+            filtered_sensor_map = {}
+            for key, val in sensor_map.iteritems():
+                key_elems = key.split(":")
+                if key_elems[0] == LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME:
+                    continue
+                filtered_sensor_map[key] = val
+            sensor_map = filtered_sensor_map
+
+        # Check that there is at least one usable temperature sensor
+        if len(sensor_map) == 0 and not self.is_virtual():
```
vext01: We should add a similar check in `OpenBSDPlatform`.

krun/platform.py
```diff
@@ -761,23 +762,61 @@ def find_temperature_sensors(self):
         The sensors move between reboots, so we have to roll our own sensor
         identifier. We are not fussy, and use as many sensors as we can find."""
 
+        ignore_unnamed_chips = False
```
ltratt: I don't understand why this variable is present a) because I think it's meant to be set to `True` b) I don't see why anyone would want it to be set to `False`. Am I missing something?

krun/platform.py
```diff
@@ -761,23 +762,61 @@ def find_temperature_sensors(self):
...
+            try:
+                with open(name_path) as fh:
+                    chip_name = fh.read().strip()
+            except IOError:
+                # The chip does not have a name.
+                # That's fine as long as there is only one chip like this.
+                debug("Un-named temperature sensor chip found.")
+                chip_name = LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME
+
+            if chip_name in chips_seen:
```
ltratt: Shouldn't this `if` clause be in the `except` clause?

krun/platform.py
snim2: It is set to true on line 793 and checked on 807, so I think this is OK?

ltratt: OK, so maybe the variable is badly name? Should it be something like `found_unnamed_chips'?

krun/platform.py
```diff
@@ -761,23 +762,61 @@ def find_temperature_sensors(self):
...
             inputs_glob = os.path.join(chip, "temp[0-9]*_input")
             chip_sensors = glob.glob(inputs_glob)
 
             for sensor in chip_sensors:
                 key = "%s:%s" % (chip_name, os.path.basename(sensor))
-                assert key not in sensor_map  # two chips with the same name?
+                assert key not in sensor_map
                 sensor_map[key] = sensor
 
+        # In the case where we saw more than a single un-named chip, remove the
```
snim2: s/more than a single/more than one/g

krun/platform.py
vext01: We ignore un-named chips if we find more than one un-named chip.

krun/platform.py
vext01: I want to know if there is a duplicate chip name, whether it is "unknown" or otherwise.

krun/platform.py
ltratt: OK, so I think this variable would then be better as `found_unnamed_chips` (or similar; shorter might be better!) then.

krun/platform.py
ltratt: OK, I get it. I'm not totally keen on the `assert chip_name == LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME` though, because given all the weirdness we've seen, it seems likely that some poor soul will have duplicates that aren't unnamed.  Maybe we should have something like:

```python
if chip_name != LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME:
    fatal("Duplicated chip named '%s'")
```
so that it's a bit clearer what happened?

krun/platform.py
```diff
@@ -706,6 +706,7 @@ class LinuxPlatform(UnixLikePlatform):
     CSET_CMD = "/usr/bin/cset"
     USER_CSET_DIR = "/cpusets/user"
     RESTRICT_DMESG_FILE = "/proc/sys/kernel/dmesg_restrict"
+    UNKNOWN_SENSOR_CHIP_NAME = "unknown"
```
ltratt: I suggest we make this name a *lot* weirder, because if a real chip has a name "unknown" (which isn't *that* unlikely) our later test will trigger.

vext01:
I was thinking about this last night, and I think I can come up with something better and more general. Watch this space.

vext01:
The code now allows for chips of the same name, so long as the clashing chip names expose a differing number of sensors (in other words the number of sensors becomes part of the unique ID).

I shuffled the data structures about a bit to try and make clearer code.

Please test and past the relevant debug log lines.

ltratt:
@snim2 and @fsfod I think Edd is asking for testing from you guys.

vext01:
Ignore the test failures. That's just that the mock is missing the new abstract method.

fsfod:
```
[2017-09-08 20:00:52: DEBUG] Un-named temperature sensor chip found.
[2017-09-08 20:00:52: DEBUG] Naming the chip: '__krun_unknown_chip_name'
[2017-09-08 20:00:52: DEBUG] Detected temperature sensors: {'acpitz:1:temp1_input': '/sys/class/hwmon/hwmon0/temp1_input', 'coretemp:3:temp1_input': '/sys/class/hwmon/hwmon2/temp1_input', 'coretemp:3:temp3_input': '/sys/class/hwmon/hwmon2/temp3_input', 'coretemp:3:temp2_input': '/sys/class/hwmon/hwmon2/temp2_input'}
```

vext01:
Thanks.

Since none of ouur machine will hit the problem cases, I'd like to add some tests.

snim2:
This runs to completion for me:

```
[2017-09-11 11:38:30: DEBUG] Un-named temperature sensor chip found.
[2017-09-11 11:38:30: DEBUG] Naming the chip: '__krun_unknown_chip_name'
[2017-09-11 11:38:30: DEBUG] Detected temperature sensors: {'pch_skylake:1:temp1_input': '/sys/class/hwmon/hwmon1/temp1_input', 'coretemp:3:temp1_input': '/sys/class/hwmon/hwmon4/temp1_input', 'iwlwifi:1:temp1_input': '/sys/class/hwmon/hwmon3/temp1_input', 'coretemp:3:temp3_input': '/sys/class/hwmon/hwmon4/temp3_input', 'acpitz:1:temp1_input': '/sys/class/hwmon/hwmon0/temp1_input', 'coretemp:3:temp2_input': '/sys/class/hwmon/hwmon4/temp2_input'}
```

krun/platform.py
```diff
@@ -762,26 +773,55 @@ def find_temperature_sensors(self):
...
+                with open(name_path) as fh:
+                    chip_name = fh.read().strip()
+            except IOError:
+                debug("Un-named temperature sensor chip found.")
+                debug("Naming the chip: '%s'" % LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME)
+                chip_name = LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME
 
             inputs_glob = os.path.join(chip, "temp[0-9]*_input")
-            chip_sensors = glob.glob(inputs_glob)
+            chip_sensors = list(glob.glob(inputs_glob))
```
snim2: Does it matter that the glob is unsorted (I'm guessing not)?

krun/platform.py
```diff
@@ -762,26 +773,55 @@ def find_temperature_sensors(self):
...
+            chip_id = (chip_name, len(chip_sensors))
+
+            if chip_id in sensor_tree:
+                info("Found duplicate chips named '%s' with '%s' temperature sensors")
```
snim2: Should this call to `info` not be `info("Found duplicate chips named '%s' with '%s' temperature sensors" % (NAME, NUMBER))`?

krun/platform.py
```diff
@@ -762,26 +773,55 @@ def find_temperature_sensors(self):
...
+                # Ignore all chips with this ID
+                duplicate_chip_ids.add(chip_id)
+                del sensor_tree[chip_id]
+                continue
+            elif chip_id in duplicate_chip_ids:
+                info("Found another chip named '%s' with '%s' temperature sensors")
```
snim2: Likewise...

krun/platform.py
vext01: no, these will become keys in an unordered dict anyways.

krun/platform.py
vext01: Yes! My tests found this too!

krun/platform.py
vext01: Indeed

vext01:
I've added tests. @snim2 could you try running the tests and the example again (sorry).

Any comments on the testing code?

snim2:
OK, the example runs to completion for me, but the tests still fail.

vext01:
Can you show me the crash?

snim2:
Sure:

```
libkrun/test/test_libkruntime.py FF..F.FFF..F

===================================================================================== FAILURES =====================================================================================
_________________________________________________________________________ TestLibKrunTime.test_cycles_u64 __________________________________________________________________________

self = <test_libkruntime.TestLibKrunTime object at 0x7fb34e640590>

    @pytest.mark.skipif(not CORECYCLES_SUPPORT,
                        reason="no performance counters")
    def test_cycles_u64(self):
        rv, out, _ = invoke_c_prog("cycles_u64")
>       assert rv == 0
E       assert 1 == 0

libkrun/test/test_libkruntime.py:85: AssertionError
________________________________________________________________________ TestLibKrunTime.test_cycles_double ________________________________________________________________________

self = <test_libkruntime.TestLibKrunTime object at 0x7fb34e640350>

    @pytest.mark.skipif(not CORECYCLES_SUPPORT,
                        reason="no performance counters")
    def test_cycles_double(self):
        rv, out, _ = invoke_c_prog("cycles_double")
>       assert rv == 0
E       assert 1 == 0

libkrun/test/test_libkruntime.py:94: AssertionError
___________________________________________________________________ TestLibKrunTime.test_cycles_u64_double_ratio ___________________________________________________________________

self = <test_libkruntime.TestLibKrunTime object at 0x7fb34e653850>

    @pytest.mark.skipif(not CORECYCLES_SUPPORT, reason="would divide by zero")
    def test_cycles_u64_double_ratio(self):
        rv, out, _ = invoke_c_prog("cycles_u64_double_ratio")
>       assert rv == 0
E       assert 1 == 0

libkrun/test/test_libkruntime.py:111: AssertionError
_________________________________________________________________________ TestLibKrunTime.test_aperf_mperf _________________________________________________________________________

self = <test_libkruntime.TestLibKrunTime object at 0x7fb34e640ad0>

    @pytest.mark.skipif(not APERF_MPERF_SUPPORT,
                        reason="no performance counters")
    def test_aperf_mperf(self):
        rv, out, _ = invoke_c_prog("aperf_mperf")
>       assert rv == 0
E       assert 1 == 0

libkrun/test/test_libkruntime.py:127: AssertionError
____________________________________________________________________________ TestLibKrunTime.test_aperf ____________________________________________________________________________

self = <test_libkruntime.TestLibKrunTime object at 0x7fb34e653e90>

    @pytest.mark.skipif(not APERF_MPERF_SUPPORT,
                        reason="no performance counters")
    def test_aperf(self):
        rv, out, _ = invoke_c_prog("aperf")
>       assert rv == 0
E       assert 1 == 0

libkrun/test/test_libkruntime.py:140: AssertionError
____________________________________________________________________________ TestLibKrunTime.test_mperf ____________________________________________________________________________

self = <test_libkruntime.TestLibKrunTime object at 0x7fb34e653ed0>

    @pytest.mark.skipif(not APERF_MPERF_SUPPORT,
                        reason="no performance counters")
    def test_mperf(self):
        rv, out, _ = invoke_c_prog("mperf")
>       assert rv == 0
E       assert 1 == 0

libkrun/test/test_libkruntime.py:148: AssertionError
__________________________________________________________________ TestLibKrunTime.test_read_everything_all_cores __________________________________________________________________

self = <test_libkruntime.TestLibKrunTime object at 0x7fb34e694610>

    def test_read_everything_all_cores(self):
        rv, out, err = invoke_c_prog("read_everything_all_cores")
        assert rv == 0
        dct = parse_keyvals(out, doubles=True)
    
        # Two wallclock measurements
        expect = 2
    
        # Two more for measurements for each core
        if CORECYCLES_SUPPORT:
            expect += PLATFORM.num_cpus * 2
    
        # Two more for measurements for each aperf and mperf on each core
        if APERF_MPERF_SUPPORT:
            expect += 2 * PLATFORM.num_cpus * 2
    
>       assert len(dct) == expect
E       assert 2 == 26
E        +  where 2 = len({'wallclock_0': 94355.869442, 'wallclock_1': 94355.869442})

libkrun/test/test_libkruntime.py:178: AssertionError
```

also I'm getting a lot of this:

```
WL1 krun/tests/test_genericplatform.py:79 'caplog.text()' syntax is deprecated, use 'caplog.text' property instead
```

which I don't think it really important for us.

krun/tests/test_linuxplatform.py
```diff
@@ -268,6 +268,128 @@ def fake_read_zone(self, zone):
...
+            'coretemp:5:temp3_input': '/sys/class/hwmon/hwmon0/temp3_input',
+            'coretemp:5:temp2_input': '/sys/class/hwmon/hwmon0/temp2_input',
+            'coretemp:5:temp1_input': '/sys/class/hwmon/hwmon0/temp1_input',
+            '__krun_unknown_chip_name:1:temp1_input':
+                '/sys/class/hwmon/hwmon1/temp1_input',
+            'coretemp:5:temp5_input': '/sys/class/hwmon/hwmon0/temp5_input'}
+        assert "Un-named temperature sensor chip found" in caplog.text()
+
+    def test_find_temperature_sensors0003(self, monkeypatch, caplog):
+        """ Test two un-named chips with the same number of sensors"""
```
snim2: Space between `"""` and `Test`

vext01:
OK, this is an unrelated issue.

I can fix that, but should it be another PR?

krun/tests/test_linuxplatform.py
vext01: oops

krun/tests/test_linuxplatform.py
```diff
@@ -268,6 +268,128 @@ def fake_read_zone(self, zone):
...
+            'coretemp:5:temp2_input': '/sys/class/hwmon/hwmon0/temp2_input',
+            'coretemp:5:temp1_input': '/sys/class/hwmon/hwmon0/temp1_input'
+        }
+        log = caplog.text()
+        assert "Un-named temperature sensor chip found" in log
+        assert ("Found duplicate chips named '%s' with 1 temperature sensor(s)"
+                % LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME) in log
+
+    def test_find_temperature_sensors0004(self, monkeypatch, caplog):
+        """ Test three un-named chips with the same number of sensors"""
```
snim2: Space here also

vext01:
@snim2 can you check libkrun is built and did you build with or without MSR support?

snim2:
Well, there's an `libkruntime.so`, how should I build it?

vext01:
`make clean && NO_MSRS=1 make`

snim2:
huh, so running those commands made no difference.

vext01:
Hrm, i just wonder why i can't reproduce on bencher8...

snim2:
And it makes no difference whether I make clean and remake in the libkrun directory or the top-level krun directory. Weird.

vext01:
Ah, got it.

snim2:
OK, that one passes for me too

vext01:
It was a fix which should have gone in with #327. My bad.

OK to squash?

vext01:
Typo in the commit name, will fix during squash

snim2:
So many details it's easy to miss something. Sure, do you want the others to take a look first or are we good?

vext01:
@ltratt isn't about today, but @fsfod may want to have a quick look?

ltratt:
I have a dumb question. If I have two unnamed sensors, each with different numbers of sensors, will they be included? If so, will both be named `__krun_unknown_chip_name`.

I also think we should add a specific check that no *real* sensor is called `__krun_unknown_chip_name` (I know it will never happen in practise, but it's sometimes nice to be sure!).

vext01:
@ltratt, if there are two sensors with no name, but with different numbers of sensors, then this will work (by design). I will now write a test to prove it to you!

Will also add an assertion guarding the sensor name.

vext01:
There you go Laurie :)

ltratt:
OK, so then next question: I suspect we should give each (distinct) unnamed sensor a new name. Just increment an integer on the end perhaps?

vext01:
We don't need to, they generated key takes care of that already.

vext01:
I'd also be against trying to assign sequential keys, as we don't know if the same sensor would get the same number after a reboot.

The code currently assigns a numberic part using the number of sensors. This strategy is immune to this problem.

Look at the expected result in the test I pushed earlier:
```
+        assert platform.temp_sensor_map == {
+            # First un-named chip
+            '%s:1:temp1_input' % LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME:
+                '/sys/class/hwmon/hwmon0/temp1_input',
+            # Second un-named chip
+            '%s:2:temp1_input' % LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME:
+                '/sys/class/hwmon/hwmon1/temp1_input',
+            '%s:2:temp2_input' % LinuxPlatform.UNKNOWN_SENSOR_CHIP_NAME:
+                '/sys/class/hwmon/hwmon1/temp2_input',
+        }
```

Once you have expended the unknown chip part, you will end up with sensors whose chip parts are disambiguated by the number of sensors:

```
(Pdb++) pprint.pprint(platform.temp_sensor_map)
{'__krun_unknown_chip_name:1:temp1_input': '/sys/class/hwmon/hwmon0/temp1_input',
 '__krun_unknown_chip_name:2:temp1_input': '/sys/class/hwmon/hwmon1/temp1_input',
 '__krun_unknown_chip_name:2:temp2_input': '/sys/class/hwmon/hwmon1/temp2_input'}
```

The above map has two sensor un-named chips involved.

Makes sense?

ltratt:
Good point.

OK, how about if we give the string name `__krun_blah_%d` where `%d` is the number of sensors? Basically, something that humans can use to debug things later.

vext01:
It's already there ;)

`__krun_unknown_chip_name:1:temp1_input` is `chip_name:num_sensors:sensor_name`

I'm open to reformat it, if that is the issue though.

ltratt:
OK, I'm fine with that. Sorry for the noise!

vext01:
@snim2 you ok for me to squash?

snim2:
Yes, please go ahead!

vext01:
splat.

PR #337, opened 2017-09-14
Make reboot docs more concise.

vext01:
So, turns out there are already docs for reboot mode.

I felt they could be more concise, and I don't know what the statement referring to emails is about.

OK?


snim2:
Looks OK to me. That comment on emails was probably from some old switch we removed. If @ltratt is happy I will merge

ltratt:
Fine with me.

PR #345, opened 2017-09-19
Daemonise Krun

### LINKED ISSUE #336: Krun must be started from tmux or screen (or maybe nohup).
The reason I think we lost bencher6 this week...

I think I invoked Krun outside of a tmux (I usually invoke in tmux).

When run outside of a tmux, I speculate that, the SSH connection can drop once the network card is off, thus forcefully killing Krun. Because the kill is forceful, the post-exec commands can't run, thus the network card stays down.

Pure speculation. Seems plausible?

The fix is either:
  * Invoke in tmux
  * Invoke by reboot (once `rc.local` is in place -- should work, right?)

vext01:
This change daemonises Krun. Not ready for merge, but ready for initial review.

`daemonise()` is very loosely based on the example here:
http://code.activestate.com/recipes/278731/

There's a lot of code in the above example I think we probably don't need, but I'd be interested to hear your thoughts.

If I start Krun (not in a tmux),  close the terminal, log back in, then Krun is still running :) I waited a while and looked at the log file. Krun seems to have completed OK.

Needs to be tested more thoroughly, and with reboots (will test on my Linux machine at home when I am back)

TODO:
 * Does is work with reboots.
 * Does the daemon stop the system rebooting if krun is on a mount point (as suggested in a comment in the linked recipe)
 * Update docs

Fixes #336

### DIFF
--- a/krun/util.py
+++ b/krun/util.py
@@ -528,3 +528,33 @@ def check_audit_unchanged(results, platform):
             "identical to the one on which the last results were "
             "gathered, which is not the case.")
         fatal(error_msg)
+
+
+def daemonise():
+    """Daemonise Krun"""
+
+    debug("daemonising...")
+    try:
+        pid = os.fork()
+    except OSError:
+        fatal("failed to daemonise: first fork")
+
+    if pid != 0:
+        # parent
+        os._exit(0)
+
+    os.setsid()
+    try:
+        pid = os.fork()
+    except OSError:
+        fatal("failed to daemonise: second fork")
+
+    if pid != 0:
+        # parent
+        os._exit(0)
+
+    # Redirect stdin/stdout/stderr to /dev/null
+    null = os.open("/dev/null", os.O_RDWR)
+    for fd in xrange(3):
+        # Since these fds will be in-use, dup2 will close them first
+        os.dup2(null, fd)
--- a/README.md
+++ b/README.md
@@ -378,8 +378,10 @@ easier.
 You should not collect results intended for publication with development switches
 turned on.
 
-We also recommend that for 'real' benchmarking you turn on the `--hardware-reboots`
-switch to restart physical hardware before each benchmark execution.
+We also recommend that for 'real' benchmarking you turn on the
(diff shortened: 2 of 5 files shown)


krun/util.py
```diff
@@ -528,3 +528,35 @@ def check_audit_unchanged(results, platform):
...
+
+    if pid != 0:
+        # parent
+        os._exit(0)
+
+    # Redirect stdin/stdout/stderr to /dev/null
+    null = os.open("/dev/null", os.O_RDWR)
+    for fd in xrange(3):
+        # Since these fds will be in-use, dup2 will close them first
+        os.dup2(null, fd)
```
vext01: When invoked from init on OpenBSD, these dup2 calls fail to redirect the log output to /dev/null.
Hrm...

vext01: ah, silly mistake, got it.

vext01:
I've tested this on OpenBSD and Linux. All is well.

If it's OK with you @ltratt , I will sync the branch and squash.

krun/util.py
```diff
@@ -528,3 +528,35 @@ def check_audit_unchanged(results, platform):
...
+    try:
+        pid = os.fork()
+    except OSError:
+        fatal("failed to daemonise: first fork")
+
+    if pid != 0:
+        # parent
+        os._exit(0)
+
+    # child becomes the session leader
```
ltratt: Not exactly true -- the child *creates* a new session (according to the man page). I think it's probably best to remove this comment (`man setsid` is more informative).

vext01: Eh?

OpenBSD manual:
> The calling process is the session leader of the new session

Linux manual:
> The calling process is the leader  of  the  new  session

right?

ltratt: Yes, but the current comment can be read to suggest it's taking over an *existing* session rather than a *new* session. But ultimately the comment isn't really useful. Anyone who needs to know this should look at the man page.

vext01:
OK?

ltratt:
Looks good. Squash and I'll merge?

vext01:
Splat.

PR #346, opened 2017-09-20
Openbsd fixes

vext01:
Fixes various bits on OpenBSD.

You probably want to look at individual commits.

OK?

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -572,8 +572,12 @@ class OpenBSDPlatform(UnixLikePlatform):
 
     def _check_apm_state(self):
         debug("Checking APM state is geared for high-performance")
-        adjust = False
 
+        # apmd needs to be enabled for this code to work
+        args = self.change_user_args("root") + ["rcctl", "start", "apmd"]
+        run_shell_cmd(" ".join(args))
+
+        adjust = False
         out = self._get_apm_output()
         lines = out.split("\n")
 
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -524,6 +524,7 @@ class UnixLikePlatform(BasePlatform):
 class OpenBSDPlatform(UnixLikePlatform):
     FIND_TEMP_SENSORS_CMD = "sysctl -a | grep -e 'hw\.sensors\..*\.temp'"
     GET_SETPERF_CMD = "sysctl hw.setperf"
+    VIO_DMESG_PATTERN = "virtio[0-9]+ at"
 
     # flags to set OpenBSD MALLOC_OPTIONS to. We disable anything that could
     # possibly introduce non-determinism or extra computation.
@@ -696,8 +697,13 @@ class OpenBSDPlatform(UnixLikePlatform):
         return []  # not supported on OpenBSD
 
     def is_virtual(self):
-        """Not yet implemented on this platform, assume not virtualised"""
+        """If we see a vio(4) disk in the dmesg, this is a virtual machine"""
 
+        dmesg_lines = self._collect_dmesg_lines()
+        comp_pat = re.compile(OpenBSDPlatform.VIO_DMESG_PATTERN)
+        for line in dmesg_lines:
+            if comp_pat.match(line):
+                return True
         return False
 
     def get_allowed_dmesg_patterns(self):
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -701,10 +701,7 @@ class OpenBSDPlatform(UnixLikePlatform):
 
         dmesg_lines = self._collect_dmesg_lines()
         comp_pat = re.compile(OpenBSDPlatform.VIO_DMESG_PATTERN)
-        for line in dmesg_lines:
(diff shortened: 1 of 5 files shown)


krun/platform.py
```diff
@@ -696,8 +697,13 @@ def pin_process_args(self):
         return []  # not supported on OpenBSD
 
     def is_virtual(self):
-        """Not yet implemented on this platform, assume not virtualised"""
+        """If we see a vio(4) disk in the dmesg, this is a virtual machine"""
 
+        dmesg_lines = self._collect_dmesg_lines()
+        comp_pat = re.compile(OpenBSDPlatform.VIO_DMESG_PATTERN)
+        for line in dmesg_lines:
```
ltratt: Might be simpler as something like `return any(comp_pat.match(l) for l in dmesg_lines)`?

ltratt:
One minor comment but otherwise looks OK to me.

ltratt:
We have a build failure, although it's not immediately obvious to me what's failed.

krun/platform.py
vext01: Simpler to read, much less efficient. Thoughts?

vext01:
a sleep(1) waited for 0.999626 seconds. I had asserted that we always wait at-least a second.

I didn't think this was possible!

krun/platform.py
ltratt: Is this a performance critical part of things? [I think not, but I might be wrong!]

vext01: Well, not hugely, it's just a waste of cycles, and regex aint exactly efficient.

My dmesg is 528 lines long. That's a lot of regex.

ltratt: I'd rather have easy to read code: you'd be amazed at how quick that code will execute on a modern machine!

vext01: I'm well aware of how fast a machine is :P 

I don't feel strongly enough to argue more :) Fixed.

vext01:
Can I squash this?

krun/platform.py
ltratt: Oh, I see what you were implying. You needn't worry :)

```python
>>> type(x for x in [1])
<type 'generator'>
>>> type([x for x in [1]])
<type 'list'>
```

ltratt:
I've always assumed that `sleep` is an approximation that one should not rely on (e.g. does it use a monotonic timer?). I'm not convinced that asserting anything about its duration is sensible to be honest...

krun/platform.py
vext01: Um. Not sure what you mean. I just didn't want to do load of regex and allocate a big list if it aint necessary. The vio entry, if it exists, will be early in the dmesg.

If it makes me look cleverer, i'll take credit for your code snippet though :)

vext01:
It is the monotonic timer, fwiw.

krun/platform.py
ltratt: What I mean is that the version I'm supplying lazily evaluates the list, so its performance is the same as the `for` loop. So, I believe, it's smaller code that's (more or less) the same performance.

You're welcome to take the credit though -- you'll also get the blame ;)

vext01: Ah, the generator doesn't allocate the list?

ltratt:
[I also don't see any reference to a monotonic timer in `sleep` or `nanosleep` on OpenBSD.]

vext01:
OK, I see where the confusion is.

This test is like this:
```
t1 = monotonic_time()  # monotonic
sleep(1)  # who knows how it works...
t2 = montonic_time()  # monotonic

interval = t2 - t1
```

I was talking about the readings, you are talking about the sleep.

So about exiting early:
```
     The sleep() function suspends execution of the calling process until
     either seconds seconds have elapsed or a signal is delivered to the
     process and its action is to invoke a signal-catching function or to
     terminate the process.  The suspension time may be longer than requested
     due to the scheduling of other activity by the system.
```

So as i see it, unless the process is killed then the wait should be longer than the requested interval.

We know the process wasn't killed, as it exited with 0.

This is weird.

vext01:
Oh, the process doesn't have to be dying... OK. Well, I don't know who'd be sending us a signal...

Shall we merge this then? I don't want to dwell on this too much.

ltratt:
There are other signals than kill that might be received. Whatever, I don't think we can rely on `sleep` sleeping for a given period: a test that relies on it is going to be fragile.

ltratt:
Agreed. I suggest we remove this particular part of the test.

vext01:
Yeah, the comment in the test pretty much says this.

Let me make a better test.

vext01:
This is more robust IMHO.

ltratt:
Ok, I'm fine with that. Squash and then I'll merge?

vext01:
splat.

ltratt:
Ah, you also need to rebase against origin/master please.

vext01:
Oops. And, like so. Go ahead.

PR #348, opened 2017-09-21
Kill log timestamp

### LINKED ISSUE #344: Remove timestamp from log file name.
There's a racy timestamp on the log file name created by Krun. Since the other files we create are not timestamped, I suggest we just kill the timestamp, thus making the code simple.

vext01:
Fixes #344 

I like this one, as it removes more code than it adds :)

Assigning Sarah, as it's her code I'm killing :smiling_imp: 

OK?

### DIFF
--- a/krun.py
+++ b/krun.py
@@ -220,6 +220,11 @@ def main(parser):
     on_first_invocation = not (os.path.isfile(manifest_filename) and
                                os.stat(manifest_filename).st_size > 0)
 
+    log_file = config.log_filename()
+    if on_first_invocation and os.path.exists(log_file):
+        util.fatal("Log file '%s' already exists. "
+                   "Move the file away before running Krun." % log_file)
+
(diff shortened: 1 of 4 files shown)


snim2:
*sniff* and I worked so hard on that code! 
Do we want to check that the user does not have an extant `.log` file before Krun starts?

vext01:
It's probably not a big deal, but we could if you like.

snim2:
Well, do we do the same for the env logs and so on, or do we just check the manifest?

vext01:
I think you are right. I'll do that tomorrow now :)

vext01:
@snim2 here's the check. OK to squash?

snim2:
Looks good to me, please squash away!

vext01:
splat

snim2:
Thanks

PR #351, opened 2017-09-28
Check for spaces in a couple of other config fields.

### LINKED ISSUE #343: Config items with spaces in names
PR #342 fixed an error where VM names with spaces in caused a crash. It would be good to have a test case with a benchmark name containing a space (and maybe other config items).

vext01:
Here are some more basic space checks.

We could really go to town with checking the config file (validating email addresses etc), but let's keep it simple for now.

Fixes #343

### DIFF
--- a/krun/config.py
+++ b/krun/config.py
@@ -107,6 +107,14 @@ class Config(object):
             if " " in vm_name:
                 fatal("VM names must not contain spaces")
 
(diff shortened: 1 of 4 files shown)


snim2:
Looks sensible to me. Email addresses are difficult to validate, and email is not really on the critical path of the benchmarking, it's just for user feedback, so I think we want to stick with the essentials here.

PR #352, opened 2017-09-28
Move masking

### LINKED ISSUE #325: Move masking of the core cycle counter outside the timed section.
This loop:
https://github.com/softdevteam/krun/blob/9c38f01a7e6292669d7c9f83780d44e2abae8279/libkrun/libkruntime.c#L401

Can be moved into the getter function, thus moving the masking operations out of the timed section.

vext01:
This moves the core cycle counter masking out of the critical section, instead doing the masking later when extracting the value from libkrun.

While we are here, make calling `krun_get_{aperf,mperf,core_cycles}()` abort when libkrun is built without MSR support (I also considered always returning a 0 core cycle count, but opted for the more explicit of the two).

(One C warning also fixed).

Needs testing on a machine with the Krun linux kernel, but can be reviewed nonetheless.

Fixes #325

### DIFF
--- a/libkrun/libkruntime.c
+++ b/libkrun/libkruntime.c
@@ -93,10 +93,10 @@ static int krun_num_cores = 0;
 
 // Private prototypes
 #ifdef __linux__
-static void     krun_core_bounds_check(int core);
 static void     krun_mdata_bounds_check(int mdata_idx);
 #ifndef NO_MSRS
 static int      krun_get_fixed_pctr1_width(void);
+static void     krun_core_bounds_check(int core);
 #endif // NO_MSRS
 #endif // __linux__
 void            krun_check_mdata(void);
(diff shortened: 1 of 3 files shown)


vext01:
This needs re-testing on Linux. I will be back later to do this.

ltratt:
OK.

vext01:
OK, this is working a treat. You OK for me to squash?

ltratt:
Works for me. I'm not super keen on the comments after `#endif` but I don't feel strongly about them either. Otherwise, I have no comments (no pun intended).

vext01:
We can kill them all at once in another PR if you like.

ltratt:
Up to you. I think they're pointless, but I'm not going to argue strongly for it.

Why don't you squash (and rebase!) this branch so I can merge?

vext01:
splat

ltratt:
A thought (not for this PR): we should run LLVM's `scan-build` over this. It's a low effort way of getting some surprisingly useful static analysis. You never know...

PR #357, opened 2017-10-04
Heavily rework the README.

ltratt:
**DO NOT MERGE**

Unfortunately, when I looked at the README as a whole, it had become a mess. I suspect this happened bit-by-bit, when PRs made changes to parts of the README without considering the overall structure. Because of that, the README was illogically structured, internally inconsistent, and contained multiple weird things (including sentences which stopped in the middle of nowhere). All in all, that was a slightly sobering experience (which, as a reviewer of many PRs, I feel responsible for).

This PR moves things around, and heavily edits, the README to start getting it back into shape. However, in a couple of cases, I couldn't work out what the right thing to do is, so there are `\laurie` comments which need addressing.

I suggest that the best way to review this PR is not to review the commit *per se*, but to read the whole file -- without that, you'll only have a local view of the changes and the problems with the overall structure may not become apparent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,37 +1,20 @@
 # Krun
 
-Krun is a framework for running software benchmarking experiments.
+Krun is a framework for running high-quality software benchmarking experiments.
+Krun experiments consist of a configuration file, a carefully configured
+benchmarking machine, and the benchmarks themselves.
 
-**Krun uses sudo to elevate privileges! Please read these instructions in
-full.**
 
-The `examples/` directory contains a simple experiment using Krun.
-This is a good starting point for setting up your own Krun configuration.
+## Step 1: Initial installation
 
-The example here contains two benchmark programs (*nbody* and *dummy*),
-executed on two VMs (*cPython* and native code with no VM). There is a separate
-example with Java (using *JVM* such as HotSpot), which requires you to set up
-some environment variables before you compile Krun.
-Each benchmark is run for 5 iterations on the same VM, then the VM is
-restarted and the benchmark is re-run for another 5 iterations.
-We say that the experiment runs 2 *process executions* and 5 *in-process iterations* of
-each benchmark.
-
-This configuration can be found in the file `examples/example.krun`. The
-example with Java can be found in `examples/java.krun`.
-
-## Step 1: prepare the benchmarking machine
-
-Krun currently only runs on Unix-like environments.
-To run this example experiment, you need superuser rights to the machine you are
-using, e.g. on Linux you should be able to run `sudo`.
+Krun currently only runs on Debian Linux and OpenBSD. Porting it to other
+Unix variants is unlikely to be difficult, and we welcome patches.
 
 ### Dependencies
 
-Krun currently runs on (Debian) Linux and OpenBSD.
+You need to have the following programs installed:
 
-You need to have the following installed:
-
+  * sudo
   * Python2.7 (pre-installed in Debian)
   * GNU make, a C compiler and libc (`build-essential` package in Debian)
   * cpufrequtils (Linux only. `cpufrequtils` package in Debian)
@@ -48,86 +31,168 @@ You need to have the following installed:
 If you want to benchmark Java, you will also need:
   * A Java SDK 7 (`openjdk-7-jdk` package in Debian)
 
-Note that to use pinning on Linux, `cset shield` must be in a working state.
-Some Linux distributions have been known to ship with this functionality
-broken. See the `cset` tutorial for information on how to test `cset shield`:
-https://rt.wiki.kernel.org/index.php/Cpuset_Management_Utility/tutorial
 
-### Kernel arguments
+## Step 2 (Linux only): kernel and OS setup
 
-If you are using a Linux system, you will need to set a kernel arguments to
-disable Intel P-states. If your Linux bootloader is Grub, you can follow these
-steps:
+### P-states
 
-  * Edit /etc/default/grub (e.g. `sudo gedit /etc/default/grub`)
-  * Add `intel_pstate=disable` to `GRUB_CMDLINE_LINUX_DEFAULT`
+Benchmarking is at its most accurate when Intel p-states are disabled in
+the kernel. If you are using Grub, this can be achieved as follows:
+
+  * Edit `/etc/default/grub` so that the `GRUB_CMDLINE_LINUX_DEFAULT`
+    variable includes `intel_pstate=disable`.
   * Run `sudo update-grub`
 
-You can disable Krun's P-state check with `--disable-pstate-check`, however
-this is strongly discouraged for real benchmarking.
+If you are unable to do this, you can disable Krun's P-state check with
+`--disable-pstate-check`, but be aware that this degrades the quality of the
+resulting benchmarking numbers.
 
-### The Krun Linux Kernel
+### Performance counters
 
-When Krun is run on Linux it requires a custom Linux Kernel which offers low
-latency access to the `IA32_APERF`, `IA32_MPERF` and `IA32_PERF_FIXED_CTR1`
-MSRs (sadly `IA32_APERF` or `IA32_MPERF` cannot be read from user-space via
-`rdpmc` and `rdmsr` is strictly a ring 0 operation). The kernel must also be
-configured to be tickless on all CPU cores except the boot core.
+We recommend using our custom Linux kernel found at:
 
-Instructions and source code can be found here:
-https://github.com/softdevteam/krun-linux-kernel
+  https://github.com/softdevteam/krun-linux-kernel
 
-#### Benchmarking on a Stock Linux Kernel
+which provides low latency access to the following counters:
 
-You can run Krun on a stock Linux Kernel, but Krun will be unable to
-collect data from:
+  * `IA32_PERF_FIXED_CTR1` (the core cycle counter)
+  * `IA32_APERF` counts
+  * `IA32_MPERF` counts
 
-  * IA32_PERF_FIXED_CTR1 (the core cycle counter)
-  * IA32_APERF counts
-  * IA32_MPERF counts
+If you are unable to do this, you can set `NO_MSRS=1` in your Unix environment
+when building Krun (see later), but be aware that this degrades the quality of
+the resulting benchmarking numbers.
 
-Since these are highly useful metrics, we strongly advise against using a stock
-Linux kernel for real benchmarking.
 
-With the above warning in mind, to run on a stock Linux kernel, when building
-Krun, include `NO_MSRS=1` in your environment.
+## Step 3: Fetch and build Krun
 
-## Step 2: Fetch the Krun source
+First fetch Krun:
 
-```bash
+```sh
 $ git clone --recursive https://github.com/softdevteam/krun.git
 $ cd krun
 ```
 
-## Step 3: Build Krun
+Then run `make` (or `gmake` on OpenBSD). The Krun Makefile honours the standard
+variables: `CC`, `CPPFLAGS`, `CFLAGS` and `LDFLAGS`.
 
-The Krun Makefile honours the standard variables: `CC`, `CPPFLAGS`, `CFLAGS`
-and `LDFLAGS`. For example, if you wish to use `clang` rather than `gcc` you
-can append `CC=clang` to the `make` command below. You should build Krun
-by invoking GNU make:
+If you want to benchmark Java programs, you need to set the
+`JAVA_HOME` environment variable to point to your JDK installation, and
+set several other flags:
 
-```bash
-$ pwd
-.../Krun
-$ make  # gmake on non-Linux platforms.
-```
-
-If you want to benchmark Java programs, you will also need to set the
-`JAVA_HOME` environment variable, and build Krun with some extra flags. The
-invocation below comes from a Ubuntu Linux machine, you may need to replace
-some paths and invoke `gmake` on other platforms:
-
-```bash
-$ pwd
-.../Krun
-$ env JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/ make  \
+```sh
+$ JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/ make  \
     JAVA_CPPFLAGS='"-I${JAVA_HOME}/include -I${JAVA_HOME}/include/linux"' \
-    JAVA_LDFLAGS=-L${JAVA_HOME}/lib ENABLE_JAVA=1
+    JAVA_LDFLAGS=-L${JAVA_HOME}/lib ENABLE_JAVA=1 make
 ```
 
-## Step 4: Build the benchmarks
 
-```bash
+## Step 4: Audit system services
+
+Background services (e.g. `cron` or `sendmail`) can interfere with benchmarking.
+The more services that you are able to switch off, the less interference is
+likely to occur. Some services are best disabled at boot and/or permanently
+(depending on your OS) and must be done manually. However, you may wish
+to disable some services only during benchmarking (e.g. you may wish to have a mail server
+running before and after benchmarking to inform you of benchmarking progress),
+which can be specified in the `PRE_EXECUTION_CMDS` and `POST_EXECUTION_CMDS`
+settings in your Krun config file.
+
+Commands in each list are run, in order, using the `krun` user's shell (e.g.
+`/bin/sh`). If a command fails, Krun stops execution immediately without running
+subsequent commands. If you wish execution to continue even if a command fails
+you can use standard shell idioms: e.g. `cmd || true` guarantees that the
+overall command succeeds even if `cmd` fails. 
+
+For example on a systemd Linux you may turn daemons off before execution with:
+
+```
+PRE_EXECUTION_CMDS = [
+    "sudo systemctl stop cron",
+    "sudo systemctl stop atd",
+    ...
+]
+```
+
+and turn them back on with:
+
+```
+POST_EXECUTION_CMDS = [
+    "sudo systemctl start cron || true",
+    "sudo systemctl start atd || true",
+    ...
+]
+```
+
+In general it is best practise to turn things back on explicitly, because after
+Krun runs the final benchmark it will not reboot the machine. If, for example,
+you put network interfaces down in `PRE_EXECUTION_CMDS`, you should put them
+back up in `POST_EXECUTION_CMDS` so that you can login to the machine after the
+final benchmark has been run. We urge you to check such commands carefully:
+small oversights can easily lead to you locking yourself out of the system.
+
+Krun can also copy intermediate results to a remote host and query that host to
+see whether it should suspend benchmarking. See
+`https://github.com/softdevteam/warmup_experiment/blob/master/warmup.krun` for
+more advanced options.
+
+
+### Linux
+
+Note that Debian has, from a benchmarking perspective, the unfortunate habit of
+automatically starting daemons which get pulled in by dependencies.
+
+On Linux, list services with:
+
+```sh
+# systemctl | grep running
+```
+
+Permanently disable services (including after system reset) with:
+
+```sh
+# systemctl stop <service>
+# systemctl disable <service>
+```
+
+Commonly enabled services that you may wish to disable:
+
+ * apache2
+ * memcached
+ * nfs-common
+
+
+### OpenBSD
+
+On OpenBSD, list services with:
+
+```sh
+# doas rcctl ls started
+```
+
+Permanently disable services (including after system reset) with:
+```sh
+# rcctl stop <service>
+# rcctl disable <service>
+```
+
+Commonly enabled services that you may wish to disable:
+
+ * pflogd
+ * sndiod
+
+
+## Step 5: Build and run the example
+
+The `examples` directory contains the `example.krun` experiment. This contains
+two benchmark programs (*nbody* and *dummy*), both of which are run on *C* and
+*PyPy*. Each benchmark is run for 5 *in-process iterations* (where the
+benchmark is repeated 5 times within a for loop within a single process) across
+2 *process executions* (where the entire VM is restarted).
+
+First build the examples:
+
+```sh
 $ cd examples/benchmarks
 $ pwd
 .../krun/examples/benchmarks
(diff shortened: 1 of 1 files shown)


ltratt:
[A small part of this PR also fixes https://github.com/softdevteam/krun/issues/333]

vext01:
I'd actually argue that the docs need to be a multi-chapter (one markdown per chapter) manual at this point. But we can use this as a springboard.

Reviewing this now.

README.md
```diff
@@ -1,36 +1,19 @@
...
+Krun is a framework for running high-quality software benchmarking experiments.
+Krun experiments consist of a configuration file, a carefully configured
+benchmarking machine, and the Krun program itself.
 
-**Krun uses sudo to elevate privileges! Please read these instructions in
-full.**
 
-The `examples/` directory contains a simple experiment using Krun.
-This is a good starting point for setting up your own Krun configuration.
+## Step 1: Initial installation
```
vext01: Title case all titles.

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
 
-If you are using a Linux system, you will need to set a kernel arguments to
-disable Intel P-states. If your Linux bootloader is Grub, you can follow these
-steps:
+## Step 2 (Linux only): kernel and OS setup
 
-  * Edit /etc/default/grub (e.g. `sudo gedit /etc/default/grub`)
-  * Add `intel_pstate=disable` to `GRUB_CMDLINE_LINUX_DEFAULT`
-  * Run `sudo update-grub`
+The kernel must also be
```
vext01: This is part of the krun-linux-kernel setup. You need to pull that section back up here I'm afraid.

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
+configured to be tickless on all CPU cores except the boot core.
+\laurie{shouldn't we tell people how to do this?}
 
-You can disable Krun's P-state check with `--disable-pstate-check`, however
-this is strongly discouraged for real benchmarking.
+### P-states
 
-### The Krun Linux Kernel
+Benchmarking is at its most accurate when Intel p-states are disabled in
+the kernel. If you are using Grub, this can be achieved as follows:
```
vext01: I'm not even sure that is true. I'd not even mention it.

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
 
-When Krun is run on Linux it requires a custom Linux Kernel which offers low
-latency access to the `IA32_APERF`, `IA32_MPERF` and `IA32_PERF_FIXED_CTR1`
-MSRs (sadly `IA32_APERF` or `IA32_MPERF` cannot be read from user-space via
-`rdpmc` and `rdmsr` is strictly a ring 0 operation). The kernel must also be
-configured to be tickless on all CPU cores except the boot core.
+  * Edit `/etc/default/grub` so that the `GRUB_CMDLINE_LINUX_DEFAULT`
+    variable has `intel_pstate=disable` added to it.
```
vext01: "variable includes `intel_pstate=disable`"

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
+  * Run `sudo update-grub`
+
+If you are unable to do this, you can disable Krun's P-state check with
+`--disable-pstate-check`, but be aware that this lessens the quality of the
```
vext01: s/lessens/degrades/

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
-Instructions and source code can be found here:
-https://github.com/softdevteam/krun-linux-kernel
+We recommend using our custom Linux kernel found at:
 
-#### Benchmarking on a Stock Linux Kernel
+  https://github.com/softdevteam/krun-linux-kernel
 
-You can run Krun on a stock Linux Kernel, but Krun will be unable to
-collect data from:
+which provides low latency access to the following counters:
```
vext01: May aswell backtick the MSR names while we are here.

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
 
   * IA32_PERF_FIXED_CTR1 (the core cycle counter)
   * IA32_APERF counts
   * IA32_MPERF counts
 
-Since these are highly useful metrics, we strongly advise against using a stock
-Linux kernel for real benchmarking.
+This kernel should be built and installed as any other version of the Linux
```
vext01: It would be better to point the user to the docs of the krun-linux-kernel.

This is hard in itself, as github now suggests you don't change the default repo branch any more.

I wonder if we can make a wiki for the krun-linux-kernel docs and point to that?

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
+kernel.
 
-With the above warning in mind, to run on a stock Linux kernel, when building
-Krun, include `NO_MSRS=1` in your environment.
+If you are unable to do this, you can set `NO_MSRS=1` in your Unix environment
```
vext01: NO_MSRS is a libkrun build-time knob.

README.md
```diff
@@ -48,76 +31,68 @@ You need to have the following installed:
...
-.../Krun
-$ make  # gmake on non-Linux platforms.
-```
+by invoking GNU make (`make` on Linux, but `gmake` on OpenBSD).
 
 If you want to benchmark Java programs, you will also need to set the
 `JAVA_HOME` environment variable, and build Krun with some extra flags. The
-invocation below comes from a Ubuntu Linux machine, you may need to replace
-some paths and invoke `gmake` on other platforms:
+invocation below comes from a Ubuntu Linux machine, and may need to be
```
vext01: We probably shouldn't mention ubuntu as it isn't officially supported.

README.md
```diff
@@ -127,7 +102,7 @@ $ env JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/ make  \
 
 ## Step 4: Build the benchmarks
 
-```bash
```
vext01: I think we could do with some blurb here about "you will probably be adding your own benchmarks, but for the sake of example, we have some examples you can use".

README.md
```diff
@@ -137,73 +112,80 @@ $ make
...
-Commonly enabled services you probably don't want include:
+Commonly enabled services that you may wish to disable:
 
  * pflogd
  * sndiod
 
+
 ## Step 6: Run the example experiment
 
-```bash
```
vext01: Shall we merge "build the experiments" into this section?

README.md
```diff
@@ -137,73 +112,80 @@ $ make
...
-$ cd ../
-$ pwd
-.../krun/examples
+The `examples` directory contains the `example.krun` experiment. This contains
+two benchmark programs (*nbody* and *dummy*), both of which are run on *C* and
+*CPython* (both of which we'll consider to be "VMs"). Each benchmark is run for
```
vext01: PyPy not CPython.

README.md
```diff
@@ -214,97 +196,25 @@ If you want to try the example Java benchmarks, there is a separate
...
-specified config file, such as the total number of process executions and which
-benchmark keys will be skipped etc.
 
 ## Creating your own experiments
 
-The configuration file `examples/example.krun` controls the experiment here.
-To create your own experiments, you can start by expanding on this example.
-The directory structure for the
+It is easiest to use `examples/example.krun` as a template for your own,
+new, experiments. The directory structure \laurie{"must be"?}:
```
vext01: * Benchmarks must be in a `benchmarks` subdir.

https://github.com/softdevteam/krun/blob/cf1ee595c2b07e19790f750f9a650422a815e0ec/krun/vm_defs.py#L59

 * Each benchmark then has a subdir (to match the name of the benchmark in the config file)

 * Each benchmark variant then optionally has a subdirectory (we use on variant per language for our experiments):

https://github.com/softdevteam/krun/blob/cf1ee595c2b07e19790f750f9a650422a815e0ec/krun/vm_defs.py#L121-L132

The name of the benchmark file is also determined by the variant's entry point.

README.md
```diff
@@ -317,17 +227,16 @@ experiment/
     ...
 ```
 
-The `Makefile` (or similar build configuration) should perform any necessary
-compilation or preprocessor steps.
-Each benchmark should expose a function (or method) called `run_iter`
-which is the entry point to the benchmark.
-In the case of compiled languages, it often makes sense to add an extra
-class to an existing benchmark, in order to provide an entry point.
+The `Makefile` should build the VMs and benchmarks needed for the experiment.
```
vext01: In our experiments we have two Makefiles and the top-level one recursively invokes `make` in the benchmarks dir. It doesn't need to be that way though.

README.md
```diff
@@ -317,17 +227,16 @@ experiment/
...
+
+Each benchmark should expose a function (or method) called `run_iter` which is
```
vext01: It's a little more complicated for Java because you can't have functions at the top-level:

https://github.com/softdevteam/krun/blob/cf1ee595c2b07e19790f750f9a650422a815e0ec/iterations_runners/iterations_runner.java#L230-L235

https://github.com/softdevteam/krun/blob/cf1ee595c2b07e19790f750f9a650422a815e0ec/iterations_runners/iterations_runner.java#L273

README.md
```diff
@@ -317,17 +227,16 @@ experiment/
...
+the entry point to the benchmark. To preserve source code history, it can
+be easiest to put this function in a new file, which then imports the benchmark.
 
-The examples here uses C, Java and cPython.
-The following VMs are currently supported:
+New VMs require some support in Krun. The following VMs are currently supported
+out of the box:
 
-  * Standard Java SDK (such as Hotspot)
+  * Standard JDK (i.e. Hotspot)
```
vext01: Some require patches :(

vext01: Later in this section we say for a third time which platforms are supported ;)

README.md
```diff
@@ -346,73 +255,64 @@ The following platforms are currently supported:
...
-    (with `useradd -m`).  This switch disables the use of a fresh user account,
-    meaning that `userdel` and `useradd` are not invoked, nor does Krun switch
-    user; the user Krun was invoked with is used for benchmarking.
+```sh
+$ /path/to/krun/krun.py --dry-run --quick --debug=INFO config.krun
+```
 
-  * `--dry-run`: Fakes actual benchmark processes, making them finish
-    instantaneously.
+\laurie{i think this should have `--no-reboots` on too?}
```
vext01: The logic is the other way around. We have `--hardware-reboots` to turn on rebooting.

README.md
```diff
@@ -346,73 +255,64 @@ The following platforms are currently supported:
...
+`crontab -e` and add a line similar to the following:
 
 ```
 @reboot /path/to/krun/scripts/start_krun_from_cron /path/to/your/config.krun
 ```
 
-The path to the config file must be an absolute path.
-
-Any arguments supplied after the config file path are passed to Krun unchanged.
+Note that the path to the config file must be an absolute path, otherwise `cron`
```
vext01: The absolute path constraint isn't really to do with cron. It's a Krun constraint.

README.md
```diff
@@ -346,73 +255,64 @@ The following platforms are currently supported:
...
+will be unable to run it. Any arguments supplied after the config file path are
+passed to Krun unchanged.
```
vext01: In the next para, might as well backtick `crontab(5)`.

README.md
```diff
@@ -346,73 +255,64 @@ The following platforms are currently supported:
...
+Krun config file:
+
+```python
 MAIL_TO = ["me@mydomain.com", "other_person@herdomain.com"]
 ```
 
-Krun uses `sendmail(8)` to send email, so you will need to make sure that this
-works prior to starting your experiment.
+Krun uses `sendmail(8)` to send email, so you will need to make sure that
+you have a functional SMTP server installed (and dno't forget to switch it off
```
vext01: dno't -> don't

README.md
```diff
@@ -423,7 +323,10 @@ PRE_EXECUTION_CMDS = [
 ```
 
 Similarly, you can use `POST_EXECUTION_CMDS` to turn daemons back on after each
-process execution.
+process execution. For example, if you put network interfaces down in
```
vext01: Duplication. We touched on this earlier, but this section is more detailed.

vext01: What has disabling daemons got to do with monitoring progress actually?

README.md
ltratt: They weren't title-cased before...

README.md
```diff
@@ -1,36 +1,19 @@
...
-
-This configuration can be found in the file `examples/example.krun`. The
-example with Java can be found in `examples/java.krun`.
-
-## Step 1: prepare the benchmarking machine
-
-Krun currently only runs on Unix-like environments.
-To run this example experiment, you need superuser rights to the machine you are
-using, e.g. on Linux you should be able to run `sudo`.
+Krun currently only runs on Unix-like environments: it has been tested
```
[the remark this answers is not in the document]
ltratt: It works on Ubuntu (somewhat) IIRC.

README.md
ltratt: What's not true? This hasn't really changed in a meaningful way from the old thing.

README.md
ltratt: I think Java users will realise that.

README.md
vext01: Oops. Well up to you if you change them.

README.md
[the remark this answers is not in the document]
vext01: So do we want to support their kernel customisations too? I'd argue that ubuntu is a desktop oriented distro not well suited to benchmarking anyway.

README.md
vext01: That "Benchmarking is at its most accurate when Intel p-states are disabled".

I'm not suggesting this is your fault. I guess I'm finding new bugs.

README.md
vext01: OK :)

vext01:
The problem is, a README is usually a little text file. Here we have a humongous manual in one big file. It's overwhelming even if you are the main author of Krun ;)

README.md
ltratt: Huh?

README.md
ltratt: Not sure I follow.

README.md
ltratt: We don't ask people to install PyPy... Should we? Otherwise this will fail. It also feels really weird to have C as an example. Could we change it to PyPy and CPython maybe?

README.md
ltratt: Right, so if some JDKs require patches, we don't support them: I think we only support HotSpot?

ltratt:
I have addressed most of the comments, but several still have questions from me that are waiting on you.

README.md
vext01: The tickless configuration is covered in the krun-linux-kernel docs.

README.md
vext01: Let's not detail how to build the kernel here as that is covered in the krun-linux-kernel repo.

The difficulty is pointing users to it. The docs for the kernel are in branches (not master). From this repo, we can either point them to a specific branch in the krun-linux-kernel repo and continually update the link as we track debian kernel versions, or we can point at the krun-linux-kernel repo landing page and update the default branch when a new version is released.

The problem with the former: It's annoying and will be forgotten.

The problem with the latter:
untitled

Github supports wikis. I wonder if we can add some basic info on how to choose the right branch (and look at the docs within) to a wiki page and link to that instead?

Hope that makes it clearer?

README.md
vext01: I think we ditched CPython becuase we wanted two fast languages in the example which the test suite runs.

vext01: Not sure why C isn't a valid example though.

README.md
ltratt: What does "pull ... back up" mean?

README.md
vext01: C compilers are ubiquitous, which makes it an example lots of people are likely to be able to run out of the box. PyPy, not so much, but CPython is slow.

README.md
vext01: Some VMs sorry. The JVM has a good enough FFI already.

There are comments at the top of the iterations runners for languages which we needed to patch for the warmup experiment. So that's:

* v8
* HHVM
* JRuby/Truffle or TruffleRuby

README.md
ltratt: That perhaps suggests that we should make the other repo do everything in the master branch? Anyway, whatever you choose, you just need to tell me what to write here.

README.md
ltratt: How slow is CPython? Is it bad enough to make this a big problem? If I'm a reader seeing "C" is pretty confusing -- confusing enough that the speed up probably may well not be worth it.

README.md
```diff
@@ -129,6 +103,32 @@ running before and after benchmarking to inform you of benchmarking progress),
...
+    "sudo systemctl start atd",
+    ...
+]
+```
+
+In general it is best practise to turn things back on explicitly, because after
+Krun runs the final benchmark it will not reboot the machine. If, for example,
+you put network interfaces down in `PRE_EXECUTION_CMDS`, you should put them
+back up in `POST_EXECUTION_CMDS` so that you can login to the machine after the
+final benchmark has been run.
```
vext01: Shall we say something like

> If you are going to take down SSH and network, you are going to get this wrong many times, and you will lock yourself out. We recommend machines with remote management so that locked out machines can we recovered

Call me a pessimist if you like :)

README.md
ltratt: OK, but I think we don't need to mention that here. I've removed the confusing stuff and been clearer that we support OpenJDK.

README.md
ltratt: I don't think it really helps people -- if they have remote management, they do, and if they don't, they don't. Telling them in the README isn't going to change their hardware...

ltratt: But I can make it a bit clearer that this is something people need to be really careful about.

README.md
```diff
@@ -129,6 +103,32 @@ running before and after benchmarking to inform you of benchmarking progress),
...
+    "sudo systemctl stop atd",
+    ...
+]
+```
+
+and turn them back on with:
+
+```
+POST_EXECUTION_CMDS = [
+    "sudo systemctl start cron",
```
vext01: Random thought: shall we mention the `|| true` hack to allow things to fail?

ltratt: Good idea.

README.md
vext01: Yeah, we just need to warn people that it's really easy to lock yourself out.

I do it on a regular basis!

vext01: There's the stop cookie hack, but it might be an information overload?

We might point users at the warmup experiment for an "involved example"?

ltratt: Good idea.

README.md
vext01: The example runs nbody with a parameter of 15. Rough timings on my laptop:
```
wilfred:python> time pypy bench.py
pypy bench.py  4.08s user 0.17s system 93% cpu 4.536 total
wilfred:python> time python bench.py
python bench.py  64.40s user 19.94s system 99% cpu 1:25.03 total
```

I don't get why C is confusing. It's a language like any other than you can benchmark. What am I missing here? :P

README.md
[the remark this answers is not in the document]
ltratt: Sarah has run things on Ubuntu; Tom on Arch. I don't think we can, or should, say "Debian or nothing" (though we can say "Debian is best for this task") as we'll put people off getting *some* improvements.

README.md
ltratt: C isn't a VM and most people won't think that its warmup is relevant. The original text even said "VMs like PyPy and C"!

Are those timings for one iteration or a complete run of the benchmark?

README.md
```diff
@@ -48,263 +31,213 @@ You need to have the following installed:
...
 
-### Kernel arguments
+## Step 2 (Linux only): kernel and OS setup
 
-If you are using a Linux system, you will need to set a kernel arguments to
-disable Intel P-states. If your Linux bootloader is Grub, you can follow these
-steps:
+The kernel must also be
+configured to be tickless on all CPU cores except the boot core.
+\laurie{shouldn't we tell people how to do this?}
```
ltratt: I think this comment has been forgotten!

README.md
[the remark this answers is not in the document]
vext01: It kinda works on Ubuntu because it is almost Debian, but we won't be making Linux kernel versions for Ubuntu.

I'd be amazed if Arch worked. Are you sure it worked? IIRC there's a check for `/etc/debian_version`:

https://github.com/softdevteam/krun/blob/cf1ee595c2b07e19790f750f9a650422a815e0ec/krun/platform.py#L1410-L1420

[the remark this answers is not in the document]
ltratt: Ah, I think Tom used Debian on his laptop so I was wrong that it works on Arch. OK, let me adjust things accordingly.

README.md
vext01: Is Krun just for warmup experiments then? I hoped it would be more general.

> Krun is a framework for running high-quality software benchmarking experiments.

The above timings are one pexec and we run two in the test suite.

ltratt: Point taken. Maybe I just need to rework the text to make it less confusing.

README.md
vext01: Again, this is not the role of this repo, but instead the krun-linux-kernel repo.

README.md
vext01: I think I was suggesting a re-ordering, but now I'm not sure.

Basically the krun-linux-kernel repo documents how to make the kernel tickless.

I wrote a big comment on this, but I can't seem to find it...

README.md
vext01: I'm not sure we can do that. The master branch is, well, the master branch from upstream.

Let's not let this block this PR. I am in the process of making binary releases of the kernel. We can point them at this, then configuring tickless mode is moot.

README.md
ltratt: OK, I'll remove it. [This stuff was all in the original README! Don't shoot me :) ]

README.md
ltratt: OK, so please tell me what to write here. I can't guess.

vext01: For now, kill the sentence:

> This kernel should be built and installed as any other version of the Linux...

README.md
vext01: ok

README.md
ltratt: Just that? Nothing else?

vext01: I'm not sure what the end result will be until I've finished experimenting with binary releases of the kernel and wikis/branches in the krun repo. Sorry.

I suggest I come back to this later. We've said elsewhere in the docs that you should use the Krun kernel, right?

ltratt: OK (and "yes").

ltratt:
If that new push works for you, I think I can squash?

README.md
```diff
@@ -103,6 +98,12 @@ running before and after benchmarking to inform you of benchmarking progress),
 which can be specified in the `PRE_EXECUTION_CMDS` and `POST_EXECUTION_CMDS`
 settings in your Krun config file.
 
+Commands in each list are run, in order, using the `krun` user's shell (e.g.
+`bin/sh`). If a command fails, Krun stops execution immediately without running
```
vext01: The krun user gets the default shell for the platform upon which we are running:
https://github.com/softdevteam/krun/blob/cf1ee595c2b07e19790f750f9a650422a815e0ec/krun/platform.py#L521-L523

(The benchmarks themselves are always run with dash though)

ltratt: That's what I said, right?

README.md
```diff
@@ -179,10 +186,9 @@ Commonly enabled services that you may wish to disable:
 
 The `examples` directory contains the `example.krun` experiment. This contains
 two benchmark programs (*nbody* and *dummy*), both of which are run on *C* and
-*CPython* (both of which we'll consider to be "VMs"). Each benchmark is run for
-5 *in-process iterations* (where the benchmark is repeated 5 times within a for
-loop within a single process) across 2 *process executions* (where the entire VM
-is restarted).
+*CPython*. Each benchmark is run for 5 *in-process iterations* (where the
```
vext01: You meant PyPy?

https://github.com/softdevteam/krun/blob/cf1ee595c2b07e19790f750f9a650422a815e0ec/examples/example.krun#L38-L42

ltratt: Oops...

README.md
vext01: You said `bin/sh`, but that's not necessarily the default shell.

The default shell on Debian is configured by `/etc/default/useradd`.

Perhaps we should specify /bin/sh as the user's shell at creation time?

ltratt: The "e.g." is crucial.

vext01:
OK, go ahead and squash.

ltratt:
Squashed.

PR #365, opened 2017-11-16
Re-add the wait for the system to come up.

vext01:
Here's the fix for Krun not waiting for the system to come up.

The original code that was removed looked like this:

```
-        if self.reboot and self.started_by_init and jobs_left > 0:
-            debug("Waiting %s seconds for the system to come up." %
-                 str(STARTUP_WAIT_SECONDS))
-            self.platform.sleep(STARTUP_WAIT_SECONDS)
-
         # Important that the dmesg is collected after the above startup wait.
         # Otherwise we get spurious dmesg changes.
         self.platform.collect_starting_dmesg()
```

This version is a little different due to quite a lot of code churn since.

I'm testing now, please don't merge yet.

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -483,6 +483,12 @@ class ExecutionScheduler(object):
     def run(self):
         """Benchmark execution starts here"""
 
+        # In reboot mode, wait for the system to come up before we proceed
+        if self.platform.hardware_reboots:
+            debug("Waiting %s seconds for the system to come up." %
+                  str(STARTUP_WAIT_SECONDS))
+            self.platform.sleep(STARTUP_WAIT_SECONDS)
+
         # Important that the dmesg is collected after the above startup wait.
         # Otherwise we get spurious dmesg changes.
         self.platform.collect_starting_dmesg()


krun/scheduler.py
```diff
@@ -483,6 +483,12 @@ def _make_post_cmd_env(self, results):
     def run(self):
         """Benchmark execution starts here"""
 
+        # In reboot mode, wait for the system to come up before we proceed
+        if self.platform.hardware_reboots:
```
snim2: Do we need to check whether there are jobs left in the manifest, or has that already happened by this point?

vext01: There's an assertion a few lines down which tells me that it is assured that there are jobs left if we get here:

https://github.com/softdevteam/krun/blob/bfa452a688899b488bb4b17314293d504e259382/krun/scheduler.py#L486-L496

snim2: OK, cool

snim2:
lgtm

ltratt:
Looks good to me too. When the tests pass, let me know, and I'll merge.

vext01:
Here's a debug log from an example run on bencher7:
http://theunixzoo.co.uk/random/example.log

```
...
[2017-11-16 16:15:24: DEBUG] Reading status cookie from /home/vext01/research/krun/examples/example.manifest
[2017-11-16 16:15:24: DEBUG] Waiting 120 seconds for the system to come up.
[2017-11-16 16:17:25: DEBUG] execute shell cmd: dmesg
...
```

So, looks pretty good to me. Shall we merge?

ltratt:
The paper says 3 minutes, but I think that's OTT, so the 2 minutes here is probably still more than sufficient. Merge as-is?

vext01:
I think so.

PR #371, opened 2018-02-09
Put back rc.local-style startup.

### LINKED ISSUE #370: Cron @reboot doesn't do what it says on the tin.
`@reboot` runs as cron starts, whether the system is coming up or not.

This means that if you have a post exec command that starts `cron` back up, then you have two Kruns running. Usually the first will reboot before the second starts, apart from after the last pexec, when the first Krun exits. Then the second hits an assertion:

```
Fatal Krun error:
  File "/home/kruninit/warmup_experiment/krun/scripts/../krun.py", line 238, in main
  File "/home/kruninit/warmup_experiment/krun/scripts/../krun.py", line 362, in inner_main
    sched.run()
  File "/home/kruninit/warmup_experiment/krun/krun/scheduler.py", line 496, in run
    assert self.manifest.num_execs_left > 0
```

Since we now have `--daemonise`, it's harder to mess up a `rc.local`. We've agreed to rename the current `start_krun_from_cron` to a better name and to call this from `rc.local`.

vext01:
These changes revert the "run from cron" way of starting Krun.

Tested on Linux, only OpenBSD to follow. Please wait before merging.

Fixes #370

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,23 +300,27 @@ switches.
 Achieving the highest possible benchmarking quality requires more care. First,
 none of Krun's debug or development switches must be used. Second, Krun needs to
 run in "reboot mode" where each process execution will be run after the machine
-has (automatically) rebooted. The recommended way to do this is via `cron(8)`
-and the `start_krun_from_cron` script supplied with Krun.
+has (automatically) rebooted. The simplest way to do this is to have your init
+system invoke `scripts/run_krun_at_boot` via an `rc.local` script.
 
-As a regular user (who has access to the experiment and the Krun code), run
-`crontab -e` and add a line similar to the following:
-
+Your `/etc/rc.local` should look like this:
 ```
-@reboot /path/to/krun/scripts/start_krun_from_cron /path/to/your/config.krun
+#!/bin/sh
+/usr/bin/sudo -u someuser /path/to/krun/scripts/run_krun_at_boot /path/to/your/config.krun
+exit 0
 ```
 
-Note that: the path to `Krun` must be an absolute paths for `cron` to be
-able to run it; and the config file must be an absolute path for `Krun` to
-be able to use it. Any arguments supplied after the config file path are
-passed to Krun unchanged.
+Make sure you replace the paths as appropriate and substitute `someuser` with
+the name of a normal unprivileged user that you wish to use to kick off Krun.
 
-You can then start the experiment with the same command (i.e. your `crontab(5)`
-line without the `@reboot`).
+Make sure `/etc/rc.local` is executable and that it only contains absolute
+paths. Note that `sudo(8)` is installed in different places on different
+operating systems (for OpenBSD, it's `/usr/local/bin/sudo`).
+
+Any arguments supplied after the config file path are passed to Krun unchanged.
+
+You can then start the experiment by manually running the command from your new
+`rc.local` (i.e. `sudo -u ...`).
 
 
 ### Monitoring progress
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -300,23 +300,25 @@ switches.
...
+sudo -u someuser /path/to/krun/scripts/run_krun_at_boot /path/to/your/config.krun
+exit 0
 ```
 
-Note that: the path to `Krun` must be an absolute paths for `cron` to be
-able to run it; and the config file must be an absolute path for `Krun` to
-be able to use it. Any arguments supplied after the config file path are
-passed to Krun unchanged.
+Make sure you replace the paths as appropriate and substitute `someuser` with
+the name of a normal unprivileged user that you wish to use to kick off Krun.
```
snim2: Should this not be the `krun` user?

README.md
```diff
@@ -300,23 +300,25 @@ switches.
...
+
+Make sure `/etc/r.local` is executable and that it only contains absolute paths.
+
+Any arguments supplied after the config file path are passed to Krun unchanged.
 
-You can then start the experiment with the same command (i.e. your `crontab(5)`
-line without the `@reboot`).
+You can then start the experiment by running manually the command from you new
```
snim2: "by manually running" (not running manually) ... "from your" (not you)

README.md
vext01: No, that gets created and removed automatically. For real experiments I actually use a `kruninit` user account with just the experiments in.

README.md
vext01: oops :)

README.md
```diff
@@ -300,23 +300,25 @@ switches.
...
+`rc.local` (i.e. `sudo -u ...`).
```
snim2: Should we also say something about `systemd`?

README.md
snim2: OK, cool

README.md
vext01: What kind of thing?

snim2: Well, it is possible to start rc.local services with systemd I think ( https://askubuntu.com/questions/886620/how-can-i-execute-command-on-startup-rc-local-alternative-on-ubuntu-16-10 ) but I'm not sure whether that would be the best way of starting Krun on a systemd system. Does that make sense?

vext01: Yes, you can use `rc.local` on systemd (for now). The pushed instructions will work on systemd.

vext01:
Comments addressed.

README.md
```diff
@@ -300,23 +300,25 @@ switches.
 Achieving the highest possible benchmarking quality requires more care. First,
 none of Krun's debug or development switches must be used. Second, Krun needs to
 run in "reboot mode" where each process execution will be run after the machine
-has (automatically) rebooted. The recommended way to do this is via `cron(8)`
-and the `start_krun_from_cron` script supplied with Krun.
-
-As a regular user (who has access to the experiment and the Krun code), run
-`crontab -e` and add a line similar to the following:
+has (automatically) rebooted. The simplest way to do this is to have your init
+system invoke `scripts/run_krun_at_boot` via a `rc.local` script.
```
snim2: "an `rc.local`" (not a)

snim2:
Sorry, one more...

snim2:
OK, time to squash, unless you are waiting on more tests?

vext01:
I need to test on BSD. I'll do that now.

vext01:
OK, one small change was needed. Can I squash?

snim2:
Please do!

vext01:
splat

PR #372, opened 2018-03-14
Make sure the libkruntime is always kept in the Lua iterations_runner

fsfod:
I changed the call to krun_done in the Lua to use a direct reference to libkruntime to make sure it is always kept alive until the end of the benchmark. I also changed the call to krun_init the same way for consistency and since it is only called once.

### DIFF
--- a/iterations_runners/iterations_runner.lua
+++ b/iterations_runners/iterations_runner.lua
@@ -79,7 +79,6 @@ ffi.cdef[[
 local libkruntime = ffi.load("kruntime")
 
 local krun_init = libkruntime.krun_init
-local krun_done = libkruntime.krun_done
 local krun_measure = libkruntime.krun_measure
 local krun_get_num_cores = libkruntime.krun_get_num_cores
 local krun_get_wallclock = libkruntime.krun_get_wallclock
@@ -164,7 +163,11 @@ for BM_i = 1, BM_iters, 1 do
     end
 end
 
-krun_done()
+-- In LuaJIT, FFI functions are cdata values that are unable to reference any other object owned by
+-- the garbage collector. Calling an FFI function which has been cached on the stack (as we've done above) 
+-- might fail because the parent FFI clib object may have been GC'd . By explicitly accessing krun_done via libkruntime
+-- here we guarantee libkruntime to live until this point.
+libkruntime.krun_done()
 
 io.stdout:write("{")
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_install:
     - sudo apt-get install oracle-java8-installer
     - sudo apt-get install virt-what
     - sudo apt-get install pypy
+    - sudo apt-get install luajit
 
 install:
     - pip install -r requirements.txt
--- a/examples/benchmarks/fannkuch_redux/lua/bench.lua
+++ b/examples/benchmarks/fannkuch_redux/lua/bench.lua
@@ -1,0 +1,64 @@
+-- The Computer Language Benchmarks Game
+-- http://shootout.alioth.debian.org/
+-- contributed by Mike Pall
+
+local EXPECT_CKSUM = 1616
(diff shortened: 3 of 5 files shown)


iterations_runners/iterations_runner.lua
```diff
@@ -164,7 +162,8 @@ for BM_i = 1, BM_iters, 1 do
     end
 end
 
-krun_done()
+-- Note we want a reference to libkruntime here so it's always kept alive to this point.
```
vext01: I'd recommend a small tweak to this comment to make it more clear. How about:

> In LuaJIT, FFI functions are not considered by the garbage collector. Calling an FFI function which has been cached on the stack (as we've done above) might fail because the parent FFI object may have been GC'd . By explicitly accessing `krun_done` via `libkruntime` here we guarantee `libkruntime` to live until this point.

iterations_runners/iterations_runner.lua
```diff
@@ -103,7 +101,7 @@ end
 
 dofile(BM_benchmark)
 
-krun_init()
```
vext01: Let's cache this one too and only use `krun_done` directly.

iterations_runners/iterations_runner.lua
vext01: Is this true of regular Lua too btw? If so, revise as required.

fsfod: Normal Lua doesn't have cdata types

iterations_runners/iterations_runner.lua
fsfod: I didn't think there was much reason to cache it?

fsfod:
I thought the Lua runner would be tested in travis? so didn't try to setup some test of the original runner. I did use the same changes in my openresty runner though.

iterations_runners/iterations_runner.lua
fsfod: @vext01 what about this for the first sentence
> In LuaJIT, FFI functions are cdata values that are unable to reference any other object owned by the garbage collector.

iterations_runners/iterations_runner.lua
vext01: Well, there isn't, but it makes clear that the `krun_done` case is the only special case.

iterations_runners/iterations_runner.lua
vext01: Works for me.

vext01:
We should at least import it to check for any syntax errors.

fsfod:
I could add a Lua VM entry to examples.krun so travis tests it?

fsfod:
@vext01 the code is tested now

vext01:
Yes, I'd be OK with that. Make sure you install luajit into travis.

fsfod:
I did and it all passed. I had to add the LuaJIT VM to the example krun though file since you were already doing several layers of overlay krun files idk if it matters much

iterations_runners/iterations_runner.lua
```diff
@@ -164,7 +163,11 @@ for BM_i = 1, BM_iters, 1 do
     end
 end
 
-krun_done()
+-- In LuaJIT, FFI functions are cdata values that are unable to reference any other object owned by
```
vext01: One last question: In the Lua world, wha tis the accepted line length?

fsfod: idk i just based it the line lengths you already had in the file that long and just went with that

vext01: OK, I might do a formatting pass on this file at some point.

vext01:
Ready to squash?

fsfod:
@vext01 yes

vext01:
Go ahead.

fsfod:
@vext01 squashed and ready to merge

vext01:
Just waiting for tests...

PR #375, opened 2018-05-24
Fix cpu pinning breaking taskset commands

fsfod:
Change cpuset commands to run after all the task set commands are run. Also add clearing of cpuset when the last benchmark is finished and as a precaution when we startup in case we crashed before leaving cpuset shield still set.

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -383,6 +383,9 @@ class BasePlatform(object):
         debug("Save power")
         self._save_power()
 
+    def clear_cpu_pinning(self):
+        pass
+
     @abstractmethod
     def _save_power(self):
         pass
@@ -914,8 +917,9 @@ class LinuxPlatform(UnixLikePlatform):
 
         if self.config.ENABLE_PINNING:
             self._check_cset_installed()
+            # Clear any leftover cpu pinning since it will break taskset
+            self.clear_cpu_pinning()
         self._check_isolcpus()
-        self._check_cset_shield()
         self._check_cpu_governor()
         self._check_cpu_scaler()
         self._check_perf_samplerate()
@@ -923,6 +927,8 @@ class LinuxPlatform(UnixLikePlatform):
             self._check_tickless_kernel()
         self._check_aslr_enabled()
         self._check_dmesg_unrestricted()
+        if self.config.ENABLE_PINNING:
+            self._check_cset_shield()
 
     def _find_virt_what(self):
         debug("Check virt-what is installed")
@@ -936,6 +942,12 @@ class LinuxPlatform(UnixLikePlatform):
         if exe is None:
             fatal("virt-what is not installed")
         return exe
+
+    def clear_cpu_pinning(self):
+        debug("Clearing cpuset")
+        args = self.change_user_args() + [LinuxPlatform.CSET_CMD, "shield", "-r"]
+        out, _, _ = run_shell_cmd(" ".join(args))
+        debug(out)
 
     # separate for testing
     def _configure_cset_shield_args(self):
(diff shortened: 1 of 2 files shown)


fsfod:
@vext01 ping

krun/platform.py
```diff
@@ -937,6 +942,13 @@ def _find_virt_what(self):
             fatal("virt-what is not installed")
         return exe
 
+    def clear_cpu_pinning(self):
+         if os.path.exists(LinuxPlatform.USER_CSET_DIR):
+            debug("Clearing cpuset")
+            args = self.change_user_args("root") + [LinuxPlatform.CSET_CMD, "shield", "-r"]
```
vext01: I think you can remove the "root" argument here. IIRC it's the default.

krun/platform.py
```diff
@@ -937,6 +942,13 @@ def _find_virt_what(self):
             fatal("virt-what is not installed")
         return exe
 
+    def clear_cpu_pinning(self):
+         if os.path.exists(LinuxPlatform.USER_CSET_DIR):
```
vext01: This only exists if a shield is in place?

vext01:
Couple of comments.

krun/platform.py
vext01: We should probably find out to avoid nasty surprises. Can you look into it?

fsfod: I didn't really realize it stored the info of active shields. What i really wanted is does the cpuset binary exist. Also how do i tell if the cpuset was created by us?

vext01: I think a safer bet would be to only do the pinning bits if pinning is enabled in the user's config. We already check for the userspace tools:

https://github.com/softdevteam/krun/blob/2ef1e77703d0356dccefb84132f7643bc223e021/krun/platform.py#L912-L916

You can use `if self.config.ENABLE_PINNING:` just like we have done above.

fsfod: i just changed it to if self.config.ENABLE_PINNING

krun/platform.py
```diff
@@ -914,15 +917,17 @@ def check_preliminaries(self):
 
         if self.config.ENABLE_PINNING:
             self._check_cset_installed()
+        # Clear any leftover cpu pinning since it will break taskset
+        self.clear_cpu_pinning()
```
vext01: Should we remove the `ENABLE_PINNING` check in `clear_pinning` and move this code into the above branch?

krun/platform.py
```diff
@@ -914,15 +917,17 @@ def check_preliminaries(self):
...
         self._check_isolcpus()
-        self._check_cset_shield()
         self._check_cpu_governor()
         self._check_cpu_scaler()
         self._check_perf_samplerate()
         if not self.no_tickless_check:
             self._check_tickless_kernel()
         self._check_aslr_enabled()
         self._check_dmesg_unrestricted()
+        self._check_cset_shield()
```
vext01: And this will need to be guarded, as it doesn't check for `ENABLE_PINNING` itself. When `ENABLE_PINNING` is false there is no guarantee the cset userspace tool is installed.

krun/platform.py
```diff
@@ -937,6 +942,13 @@ def _find_virt_what(self):
             fatal("virt-what is not installed")
         return exe
 
+    def clear_cpu_pinning(self):
+         if self.config.ENABLE_PINNING:
```
vext01: And I'd remove this check, and have the call-sites responsible for the check.

krun/platform.py
fsfod: i guess so now.

vext01:
OK, this looks good. Shall we squash?

fsfod:
@vext01 squashed and ready to merge.

Issue #379, opened 2018-07-18
Clock lists should be type stable


ltratt:
Some VMs will optimise things like:

```
x = [-1] * 10
for i in range(10):
  x[i] = 0.0
```

in a way that allocates a list of unboxed ints on line 1, and then reallocs the list on line 3 to be a list of unboxed floats. We should create an array whose members never change type.

ltratt:
The Python case is now fixed in https://github.com/softdevteam/krun/pull/380#discussion_r204104472; the JS case remains to be done.

vext01:
In terms of JS, we had hoped to use typed arrays, however:
 * There's no unsigned 64-bit type in the API (or 64-bit integer for that matter)
 * Support is not universal.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays

So we will use NaN instead.

ltratt:
The JavaScript case was fixed by #382. Closing.

PR #380, opened 2018-07-19
Allocate arrays of a fixed type in Python.

ltratt:
On VMs like PyPy, this code:

```python
  x = [0]
  x[0] = 0.0
```

causes a reallocation (from list of unboxed ints to list of unboxed floats), even though the number of elements doesn't change -- which is precisely the opposite of what we wanted to do.

This PR moves things to use the array module with arrays of fixed types that can't possibly reallocate.

This addresses the Python part of https://github.com/softdevteam/krun/issues/379 though note I've only lightly tested it by using `example.krun`.

### DIFF
--- a/iterations_runners/iterations_runner.py
+++ b/iterations_runners/iterations_runner.py
@@ -45,7 +45,7 @@ usage: iterations_runner.py <benchmark> <# of iterations> <benchmark param>
 
 Arguments in [] are for instrumentation mode only."""
 
-import cffi, sys, imp, os
+import array, cffi, sys, imp, os
 
 
 ffi = cffi.FFI()
@@ -105,14 +105,13 @@ if __name__ == "__main__":
     num_cores = krun_get_num_cores()
 
     # Pre-allocate result lists
-    wallclock_times = [0] * iters
-    cycle_counts = [None] * num_cores
-    aperf_counts = [None] * num_cores
-    mperf_counts = [None] * num_cores
-    for core in xrange(num_cores):
-        cycle_counts[core] = [0] * iters
-        aperf_counts[core] = [0] * iters
-        mperf_counts[core] = [0] * iters
+    wallclock_times = array.array("d", [0.0] * iters)
+    # Although we can't be sure what size "L" actually is, if we generate ints
+    # it can't store, an OverflowError results, so there's no chance of silent
+    # truncation.
+    cycle_counts = [array.array("L", [0] * iters) for _ in range(num_cores)]
+    aperf_counts = [array.array("L", [0] * iters) for _ in range(num_cores)]
+    mperf_counts = [array.array("L", [0] * iters) for _ in range(num_cores)]
 
     # Main loop
     for i in xrange(iters):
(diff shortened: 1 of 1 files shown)


snim2:
lgtm

iterations_runners/iterations_runner.py
```diff
@@ -105,14 +105,14 @@ def usage():
...
-    wallclock_times = [0] * iters
-    cycle_counts = [None] * num_cores
-    aperf_counts = [None] * num_cores
-    mperf_counts = [None] * num_cores
-    for core in xrange(num_cores):
-        cycle_counts[core] = [0] * iters
-        aperf_counts[core] = [0] * iters
-        mperf_counts[core] = [0] * iters
+    def fill(x, n):
+        return [x for _ in range(n)]
```
vext01: Do we need the fill function? You can just use the `__mul__` of a Python list:

```
floats_fill = array.array([0.0] * iters)
```

iterations_runners/iterations_runner.py
```diff
@@ -105,14 +105,14 @@ def usage():
...
+    floats_fill = fill(0.0, iters)
+    wallclock_times = array.array("d", floats_fill)
+    ints_fill = fill(0, iters)
+    cycle_counts = [array.array("L", ints_fill) for _ in range(num_cores)]
+    aperf_counts = [array.array("L", ints_fill) for _ in range(num_cores)]
+    mperf_counts = [array.array("L", ints_fill) for _ in range(num_cores)]
```
vext01: The `L` type code is an unsigned integer with minimum size 4 bytes. The types of these arrays work by chance of x86_64 because an `unsigned long` happens to be the same size as a `uint64_t` (confirmed with a C program here).

Is there a way to have an array explicitly backed by a `uint64_t` instead? It doesn't look like it :(

vext01: (our core cycle and a/mperf metrics are `uint64_t`)

ltratt:
OK, see if that new commit works for you.

iterations_runners/iterations_runner.py
vext01: It bothers me a bit that someone can port Krun to another platform and get bogus results. This will be forgotten I reckon.

Can we either:

 * Fill a regular list with the largest value expressible by a u64.
 * Make the code crash somehow if the assumption is invalid

Neither is pretty, I know.

ltratt: `assert(array.array("L", [1<<64-1])[0] == 1<<64-1)`?

ltratt: Actually, we needn't worry:

```
>>> array.array("L", [1<<64])
  File "<stdin>", line 1, in <module>
OverflowError: long int too large to convert
```

vext01: Maybe, but I'd like to check it fails on a 32-bit arch. Do you have one around?

I notice python3 has a `Q` which is `unsigned long long` of min size 8. Sadly, we are Python 2.

vext01:
```
>>> import ctypes
>>> ctypes.sizeof(ctypes.c_ulong)
8
```

What do you think of that instead?

ltratt: The point is: it'll fail as soon as you put a value in it can't handle. So if it has fewer bytes, it'll refuse a value bigger than it can handle. I'm pretty sure we're OK here.

[And no, I don't have anything that old lurking around ;)]

vext01: Ah, yeah, OK.

Can you add a comment explaining this, and I then I'm happy.

ltratt: OK, pushed.

ltratt:
OK, squashed.

Issue #340, opened 2017-09-18
Krun does not check if turbo mode is disabled on OpenBSD.


vext01:
This is difficult because there is no access to MSRs in OpenBSD.

ltratt:
This seems unresolvable to me. Maybe we can document it as a limitation, but it doesn't sound like we can fix anything?

vext01:
At the time, I couldn't find any way of knowing if turbo mode was enabled on OpenBSD.

At the very least, there should be a line in `README.md`.

PR #381, opened 2018-07-23
Warn that Krun can't check for turbo mode on OpenBSD.

ltratt:
Fixes https://github.com/softdevteam/krun/issues/340

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,11 @@ Commonly enabled services that you may wish to disable:
  * pflogd
  * sndiod
 
+Note that Krun is unable to check whether turbo mode is enabled on OpenBSD or
+not, and is also unable to use APERF/MPERF ratios to indirectly check whether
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -185,6 +185,11 @@ Commonly enabled services that you may wish to disable:
  * pflogd
  * sndiod
 
+Note that Krun is unable to check whether turbo mode is enabled on OpenBSD or
+not, and is also unable to use APERF/MPERF ratios to indirectly check whether
```
vext01: Kill "or not"?

vext01:
This concludes my review.

README.md
ltratt: I don't think the sentence parses without "or not" ("whether" requires two clauses).

vext01: Both are fine for me, so as you wish.

ltratt: Let's keep it as-is. I think we can then merge?

PR #386, opened 2018-08-02
Check over-commit is totally off when running on Linux.

### LINKED ISSUE #385: Check for disabled overcommit
https://www.kernel.org/doc/Documentation/vm/overcommit-accounting

We want memory overcommit OFF, yet the default on Debian 8 and 9 the default is 0 (partial).

Krun should set the policy to 2 as part of it's sanity checks.

vext01:
Query the over-commit policy and set it to 2 if it is not already 2.

Tested on Debian 9 under VMM.

OK?

Fixes #385

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,6 +497,7 @@ than one leads to undefined behaviour).
  * Detect virtualised hosts.
  * Unrestrict the dmesg buffer (Linux Kernel 4.8+)
  * Turn off "turbo boost" (Linux only)
+ * Turn off memory over-commit (Linux only).
 
 Please make sure you understand the implications of this.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -497,6 +497,7 @@ than one leads to undefined behaviour).
  * Detect virtualised hosts.
  * Unrestrict the dmesg buffer (Linux Kernel 4.8+)
  * Turn off "turbo boost" (Linux only)
+ * Set the memory over-commit policy (Linux only).
```
ltratt: Don't we need to say "turns off memory over-commit" or similar? i.e. tell people *what* policy we're choosing?

vext01: Fixed. Try that.

vext01:
splat.

PR #394, opened 2019-12-19
Tickless fixes

vext01:
The first commit message should help to explain the need for this change.

The second commit is just test housekeeping. Also note that the test suite explodes badly using newer py.test versions. I had to install the old versions listed in `requirements.txt`.

### DIFF
--- a/krun/platform.py
+++ b/krun/platform.py
@@ -754,17 +754,10 @@ class LinuxPlatform(UnixLikePlatform):
         # Omit scheduler ticks when CPU is idle.
         "CONFIG_NO_HZ_IDLE": False,
 
-        # Omit scheduler ticks when an adaptive tick CPU has only one
-        # runnable process. This enables the tickless functionality, but the
-        # system admin must manually specify on adaptive-tick CPUs via
-        # the kernel command line. By default, no CPUs are adaptive tick,
-        # so we insist also upon CONFIG_NO_HZ_FULL_ALL.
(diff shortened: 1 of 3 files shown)


ltratt:
I am not worried about Python 2. It will be around for a fair while yet and porting Krun would take more effort than I think we can justify at the moment.

krun/platform.py
```diff
@@ -1036,9 +1029,22 @@ def _check_tickless_kernel(self):
 
         debug("Checking linux kernel is tickless")
 
+        if self.num_cpus == 1:
+            # Tickless mode doesn't make sense on a single-cpu system, as you
+            # need at least on non-adaptive ticks CPU to do various
```
ltratt: "at least on non-adaptive" doesn't parse. [I can guess the overall meaning, but only because I know the area!]

vext01: ah, `on` -> `one`. Good catch.

ltratt: Even then it's still a bit hard to parse. Maybe "non-adaptive-ticks" helps?

vext01:
> I am not worried about Python 2. It will be around for a fair while yet and porting Krun would take more effort than I think we can justify at the moment.

I think you are right. I just checked Debian testing and that is surprisingly still Python 2 by default. So that means Debian 11 (when it comes) will still use Python 2 as default.

vext01:
Pushed a fix.

ltratt:
Please squash.

vext01:
splat. also updated PR description.

PR #395, opened 2019-12-19
Refurbish test suite

vext01:
This makes the Krun test suite work on newer versions of things found in Debian 10.

I don't think we've updated the deps since Debian 7 days!

I'll squash this all into one at the end.

### DIFF
--- a/krun/tests/test_linuxplatform.py
+++ b/krun/tests/test_linuxplatform.py
@@ -1,10 +1,11 @@
 import pytest
+import logging
 import krun.platform
 from distutils.spawn import find_executable
 from krun.platform import LinuxPlatform
 from krun.tests import BaseKrunTest, subst_env_arg
 from krun.util import FatalKrunError, run_shell_cmd
-from krun.vm_defs import  PythonVMDef
+from krun.vm_defs import PythonVMDef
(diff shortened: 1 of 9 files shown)


krun/tests/test_linuxplatform.py
```diff
@@ -439,6 +448,8 @@ def test_check_dmesg_filter0001(self, platform, mock_manifest):
             platform.get_dmesg_whitelist(), old_lines, new_lines, mock_manifest)
 
     def test_aslr0001(self, platform, caplog):
+        caplog.set_level(logging.DEBUG)
```
ltratt: Is this intentional?

vext01: Yeah. caplog used to capture all messages by default, but now it doesn't capture debug-level by default. Some tests check the debug level messages, so we have to turn them on.

ltratt: OK, we should tweak the commit message a bit?

vext01: Sure. I'll include it at squash time.

ltratt:
Please squash / force push.

vext01:
Give that a go.

ltratt:
Works for me!

PR #396, opened 2020-05-12
Implement timeouts for process executions.

### LINKED ISSUE #26: Iterations runner output is printed at the end of the process.
The iterations runners print status messages to the effect:

```
iteration 1/2000
iteration 2/2000
...
```

However, due to stderr buffering, these messages are all printed at once when the execution ends. It would be nice if we could have these messages printed as they arrive.

vext01:
This is required because sometimes you need to benchmark VMs and/or
benchmark suites which freeze up. We've seen this numerous times with
Java VMs and benchmarks. Without having a timeout to kill frozen process
executions, you can't make progress and have to restart the experiment
(possibly skipping that benchmark).

Further, by restarting the experiment you are biasing your experiment:
giving benchmarks that have the worst performance (that's what a
non-terminating benchmark is) another chance to redeem themselves.

Timed-out process executions now show as a 'T' flag in the manifest file.
Flags for finished process executions are carried through to the results
file in a new `pexec_flags` field. This field can be one of {C, E, T}
(completed, errored, timed-out).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ You need to have the following programs installed:
   * GNU make, a C compiler and libc (`build-essential` package in Debian)
   * cpufrequtils (Linux only. `cpufrequtils` package in Debian)
   * cffi (`python-cffi` package in Debian)
+  * subprocess32 Python module.
   * cset (for pinning on Linux only. `cpuset` package in Debian)
   * virt-what (Linux only. `virt-what` package in Debian)
   * Our custom Linux kernel (see below).
@@ -464,6 +465,10 @@ The structure of the JSON results is as follows:
                                 # (structure same as 'core_cycle_counts')
     'mperf_counts': {...}       # Per-core MPERF deltas
                                 # (structure same as 'core_cycle_counts')
+    'pexec_flags': {...}        # A flag for each process execution:
+                                # 'C' completed OK.
+                                # 'E' benchmark crashed.
+                                # 'T' benchmark timed out.
     'eta_estimates': {u"bmark:VM:variant": [t_0, t_1, ...], ...} # A dict mapping
                   # benchmark keys to rough process execution times. Used internally:
                   # users can ignore this.
--- a/krun/results.py
+++ b/krun/results.py
@@ -32,6 +32,9 @@ class Results(object):
         self.aperf_counts = dict()
         self.mperf_counts = dict()
 
+        # Record the flag for each process execution.
+        self.pexec_flags = dict()
+
         # Record how long execs are taking so we can give the user a rough ETA.
         # Maps "bmark:vm:variant" -> [t_0, t_1, ...]
         self.eta_estimates = dict()
@@ -79,6 +82,7 @@ class Results(object):
                     self.core_cycle_counts[key] = []
                     self.aperf_counts[key] = []
                     self.mperf_counts[key] = []
+                    self.pexec_flags[key] = []
                     self.eta_estimates[key] = []
 
     def read_from_file(self, results_file):
@@ -104,6 +108,7 @@ class Results(object):
             cycles_len = len(self.core_cycle_counts[key])
             aperf_len = len(self.aperf_counts[key])
             mperf_len = len(self.mperf_counts[key])
+            pexec_flags_len = len(self.pexec_flags[key])
 
             if eta_len != wct_len:
                 fatal("inconsistent etas length: %s: %d vs %d" % (key, eta_len, wct_len))
@@ -116,6 +121,9 @@ class Results(object):
 
             if mperf_len != wct_len:
                 fatal("inconsistent mperf length: %s: %d vs %d" % (key, mperf_len, wct_len))
+
+            if pexec_flags_len != wct_len:
+                fatal("inconsistent pexec flags length: %s: %d vs %d" % (key, pexec_flags_len, wct_len))
 
             # Check the length of the different measurements match and that the
             # number of per-core measurements is consistent.
@@ -167,6 +175,7 @@ class Results(object):
             "core_cycle_counts": self.core_cycle_counts,
             "aperf_counts": self.aperf_counts,
             "mperf_counts": self.mperf_counts,
+            "pexec_flags": self.pexec_flags,
             "audit": self.audit.audit,
             "eta_estimates": self.eta_estimates,
             "error_flag": self.error_flag,
@@ -189,17 +198,22 @@ class Results(object):
                 self.core_cycle_counts == other.core_cycle_counts and
                 self.aperf_counts == other.aperf_counts and
                 self.mperf_counts == other.mperf_counts and
+                self.pexec_flags == other.pexec_flags and
                 self.audit == other.audit and
                 self.eta_estimates == other.eta_estimates and
                 self.error_flag == other.error_flag)
 
-    def append_exec_measurements(self, key, measurements):
+    def append_exec_measurements(self, key, measurements, flag):
         """Unpacks a measurements dict into the Results instance"""
+
+        # Only a subset of flags can arise at this time.
+        assert flag in ("C", "E", "T")
 
         # Consistently format monotonic time doubles
         wallclock_times = format_raw_exec_results(
             measurements["wallclock_times"])
 
+        self.pexec_flags[key].append(flag)
         self.wallclock_times[key].append(wallclock_times)
         self.core_cycle_counts[key].append(measurements["core_cycle_counts"])
         self.aperf_counts[key].append(measurements["aperf_counts"])
--- a/krun/util.py
+++ b/krun/util.py
@@ -4,11 +4,12 @@ import re
 import select
 import shutil
 import sys
-import subprocess
 import pwd
 import grp
 import getpass
-from subprocess import Popen, PIPE
+import time
+from datetime import datetime, timedelta
+import subprocess32  # For timeout support.
 from logging import error, debug, info, warn, root as root_logger
 from bz2 import BZ2File
 from krun.amperf import check_amperf_ratios
@@ -56,6 +57,7 @@ class RerunExecution(Exception):
 class FatalKrunError(Exception):
     pass
 
+
 def fatal(msg):
     error(msg)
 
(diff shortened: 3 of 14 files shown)


krun/util.py
```diff
@@ -192,16 +207,28 @@ def read_popen_output_carefully(process, platform, print_stderr=True):
...
                     stderr_consumer.send(d)
 
+        if timeout_at is not None and datetime.now() >= timeout_at:
+            raise ExecutionTimeout()
+
     # We know stderr and stdout are closed.
     # Now we are just waiting for the process to exit, which may have
     # already happened of course.
+    while True:
+        process.poll()
```
ltratt, line 217: Why not `wait` (or its equivalent) if stdout/stderr are closed at this point? Or does `poll` call `wait` underneath?

krun/util.py
```diff
@@ -192,16 +207,28 @@ def read_popen_output_carefully(process, platform, print_stderr=True):
...
+        if process.returncode is not None:
+            break
+
+        if timeout_at is not None:
+            debug("exit: %s" % datetime.now())
+        if timeout_at is not None and datetime.now() >= timeout_at:
+            raise ExecutionTimeout()
+        time.sleep(0.1)  # A small sleep to prevent us hogging too much CPU.
```
ltratt, line 225: Wouldn't it be better on the `poll`/`wait` to specify the remaining timeout and leave it to the OS to do this?

krun/util.py
vext01: It's a silly function name. [`poll()`](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.poll) checks if the process has exited and sets the returncode if so. It does not block and is distinct from `poll(2)`.

There's also a [`wait()`](https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait), but it blocks forever.

In Python 3 (and `subprocess32`) there's [`wait()`](https://docs.python.org/3.5/library/subprocess.html#subprocess.Popen.wait) which has an optional timeout.

In the description (the EDIT part) I'm suggesting we perhaps use `subprocess32.wait()` with a timeout. I only found that half way through typing the PR description :)

krun/util.py
vext01: That's what I proposed in the PR description I think.

ltratt:
Why don't you make the changes you proposed in the description and then I can review that? I think it might be easier.

vext01:
Sure thing.

vext01:
The last push implements a better way of doing the timeout.

Also documented the function better and used `subprocess32` everywhere to avoid confusion.

If it looks good, I'll carry on integrating this into the config and results file.

krun/util.py
```diff
@@ -186,10 +199,17 @@ def read_popen_output_carefully(process, platform, print_stderr=True, timeout=No
...
-
         if timeout_at is not None:
-            debug("select done: %s" % datetime.now())
+            time_left = (timeout_at - datetime.now()).total_seconds()
+            if time_left <= 0:
+                raise ExecutionTimeout()
+            ready = select.select(open_fds, [], [], time_left)
+        else:
+            # The `select()` documentation doesn't specify what to pass for
+            # ulimited timeout. It only says that the parameter should be
```
ltratt, line 209: Surely we don't want unlimited (I assume "ulimited" is a typo!) timeout? We only want to wait until at most the end of the timeout period?

vext01:
I'm going to have to re-test this on a VirtualBox VM, as the time is still wonky on VMM VMs :(

krun/util.py
ltratt: If it's unlimited, then this PR doesn't do anything in the common case. At the moment we only apply the timeout after stdout/stderr have been closed, but generally they're closed only at the point a process is going to exit anyway, so that doesn't really buy us much. What you normally want to do is run the process for at most `timeout`, bearing in mind that `poll`/`select` can return in the interim so you need to keep updating the timeout. Something like this:

```python
timeout = <some point in the future>
while time.time() < timeout:
    remaining = min(timeout - time.time(), 0)
    poll(fds, remaining)

poll.wait(min(timeout - time.time(), 0))
```

For example, [here's lang_tester doing this sort of thing](https://github.com/softdevteam/lang_tester/blob/master/src/tester.rs#L710) (I admit that it's a bit hard to read because it's using the C API).

vext01: > If it's unlimited, then this PR doesn't do anything in the common case.

This function defaults to no timeout, if that's what you mean? We use this function for all kinds of other things too, not just for process executions, so we don't want to affect those.

> At the moment we only apply the timeout after stdout/stderr have been closed

Eh? I'm not sure how you've arrived at that. 

```
> /home/edd/research/krun/krun/util.py(207)read_popen_output_carefully()
-> ready = select.select(open_fds, [], [], time_left)
(Pdb) open_fds
[4, 6]
(Pdb) timeout
30
(Pdb) time_left
30.0
```

stderr and stdout are both open and select is being called with a 30 second timeout.

ltratt: > This function defaults to no timeout, if that's what you mean?

Isn't the point of this PR to enforce a timeout? I feel like I'm missing out on something!

krun/vm_defs.py
```diff
@@ -261,9 +261,9 @@ def _run_exec_popen(self, args, stderr_file=subprocess.PIPE):
         # command with. Command line arguments will have been appended *inside*
         # to adjust the new user's environment once the user switch has
         # occurred.
-        child_pipe = subprocess.Popen(args, stdout=subprocess.PIPE,
+        child_pipe = subprocess32.Popen(args, stdout=subprocess32.PIPE,
                                       stderr=stderr_file, env={})
-        return read_popen_output_carefully(child_pipe, platform=self.platform)
+        return read_popen_output_carefully(child_pipe, platform=self.platform, timeout=30)  # XXX
```
vext01, line 266: @ltratt  See here, we only specify a timeout when running a process execution. This value will need to eventually come from a config file.

krun/util.py
vext01: Does this clarify?
https://github.com/softdevteam/krun/pull/396#discussion_r423865174

vext01: Yeah, because some other unrelated bits of code (for which it wouldn't make sense to have a timeout) use this routine too.

So if this all seems reasonable, I'll hook this into the config file and results file. I just wanted to check the overall approach looks sane!

ltratt:
OK, I think you should go ahead with this. I guess you're not ready to squash yet though?

vext01:
We are quite a ways off yet: I think there will need to be a `warmup_stats` change too, as they won't know how to handle timed out pexecs.

I'm dreading that part to be honest.

ltratt:
> I think there will need to be a warmup_stats change too, as they won't know how to handle timed out pexecs.

That is separate from this PR, and I don't think we should wait on it. We might decide not to even augment warmup_stats in that regard, depending on what we find.

vext01:
> That is separate from this PR, and I don't think we should wait on it. We might decide not to even augment warmup_stats in that regard, depending on what we find.

I think we should do them at the same time. If `warmup_stats` doesn't understand timed out pexecs in results files, it's going to either crash, or do the wrong thing with them (e.g. treat them as crashed pexecs or skipped pexecs).

Perhaps if this was all private, but it's not unheard of for people to try and use our tools.

ltratt:
I disagree: krun and warmup_stats do have an independent existence. We can try krun with a timeout, see if it tells us anything useful and then alter warmup_stats. Unless people explicitly turn the timeout option on, the output remains the same as before and thus compatible with warmup_stats, right? So this is a backwards-compatible extension.

vext01:
> I disagree: krun and warmup_stats do have an independent existence.

Kind of. The default input to warmup stats is a krun results file, isn't it?

> Unless people explicitly turn the timeout option on, the output remains the same as before and thus compatible with warmup_stats, right?

I was planning on having a default timeout, but we could *not* do that and then, yes, what you said is correct.

ltratt:
> The default input to warmup stats is a krun results file, isn't it?

It also takes CSV from any arbitrary non-Krun benchmark runner you can dream up.

> I was planning on having a default timeout, but we could not do that and then, yes, what you said is correct.

OK, let's not do it by default then we save ourselves work (at least for now) :)

vext01:
Roger that.

vext01:
With your permission, I'd like to squash and rebase to make this easier to review. May I?

ltratt:
Squashing and rebasing means I have to start reviewing from scratch. Is there a good reason to do that?

vext01:
I was hoping to present this better (rebasing only the merge of master and squashing only the my new commits), but I've been battling with git and failed.

What I could do is leave the original commits after the rebase and squash only the new ones?

ltratt:
Why don't I just review the commits as-is? At first glance they don't look *that* bad!

krun/results.py
```diff
@@ -117,6 +122,9 @@ def integrity_check(self):
             if mperf_len != wct_len:
                 fatal("inconsistent mperf length: %s: %d vs %d" % (key, mperf_len, wct_len))
 
+            if pexec_flags_len != wct_len:
+                fatal("inconsistent mperf length: %s: %d vs %d" % (key, mperf_len, wct_len))
```
ltratt: I think the error message is wrong (it's not mperf length, right?).

krun/results.py
```diff
@@ -123,7 +123,7 @@ def integrity_check(self):
                 fatal("inconsistent mperf length: %s: %d vs %d" % (key, mperf_len, wct_len))
 
             if pexec_flags_len != wct_len:
-                fatal("inconsistent mperf length: %s: %d vs %d" % (key, mperf_len, wct_len))
+                fatal("inconsistent pexec flags length: %s: %d vs %d" % (key, pexec_flags_len, wct_len))
```
ltratt, line 126: Ah, fixed here.

README.md
```diff
@@ -464,6 +465,10 @@ The structure of the JSON results is as follows:
                                 # (structure same as 'core_cycle_counts')
     'mperf_counts': {...}       # Per-core MPERF deltas
                                 # (structure same as 'core_cycle_counts')
+    'pexec_flags': {...}        # A flag for each process execution:
```
ltratt, line 468: Do we need to document `O` too?

ltratt:
Reviewed, and I only really have 1 comment.

krun/results.py
vext01: fixed in a later commit.

ltratt: Yes, I realised that later.

README.md
vext01: `O` cannot make it into the results file because an outstanding pexec doesn't have a result yet. Same applies for `S`.

vext01:
If you are satisfied with the response re: pexec flags, then I can squash.

ltratt:
Please squash.

vext01:
Squashed. If you like the commit message, I'll update the PR title and description too.

ltratt:
Please go ahead! [And mark the PR non-draft.]

vext01:
I think we are ready for bors.

ltratt:
This isn't a bors'd repo ;)

PR #397, opened 2020-05-13
Fix tests which are broken on a uniprocessor machine.

vext01:
These failures were exposed by development in a Debian VM.

Tickless kernels are irrelevant on a uniprocessor machine and this
caused some tests which are expected to fail to not.

Also add a test which checks that a kernel without tickless mode works
if the machine has only one CPU.

This is unrelated to #396 -- just I found these failures while doing that.

### DIFF
--- a/krun/tests/test_linuxplatform.py
+++ b/krun/tests/test_linuxplatform.py
@@ -42,12 +42,14 @@ class TestLinuxPlatform(BaseKrunTest):
         monkeypatch.setattr(krun.platform.LinuxPlatform,
                             "_open_kernel_config_file",
                             mock_open_kernel_config_file)
+        platform.num_cpus = 2
 
         def wrap_get_kernel_cmdline(self):
             return "nohz_full=1-%s" % (platform.num_cpus - 1)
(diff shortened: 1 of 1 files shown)


ltratt:
What does `platform.num_cpus = 2` mean/do? I initially thought it was an assertion of sorts, but then realised it was an assignment, and I'm not sure how you assign a number of CPUs?

vext01:
It's tricking the platform abstractions into thinking that there are two CPUs. This is needed to prevent the platform being OK with a non-tickless kernel on account of the test host being uni-processor.

Think of it as a poor-man's mock.

ltratt:
OK, got it!

vext01:
> This is needed to prevent the platform being OK with a non-tickless kernel on account of the test host being uni-processor.

I should elaborate. A "fully tickless" kernel needs one "non-tickless" CPU to manage the other tickless cores. If you only have one CPU, then you can't have any tickless cores and Krun is fine with that.

In reality I doubt we will ever see a bare-metal uniprocessor machine ever again anyway! :P

PR #399, opened 2020-05-19
Log which benchmark failed at the point the error is reported.

vext01:
This makes scanning the logs much, much easier for me.

### DIFF
--- a/krun/scheduler.py
+++ b/krun/scheduler.py
@@ -383,7 +383,7 @@ class ExecutionJob(object):
             try:
                 self.sched.platform.check_throttle_counts(self.sched.manifest)
                 measurements = util.check_and_parse_execution_results(
-                    stdout, stderr, rc, self.sched.config, instrument=vm_def.instrument)
+                    stdout, stderr, rc, self.sched.config, self.key, instrument=vm_def.instrument)
                 flag = "C"
             except util.RerunExecution as e:
                 subject = ("Benchmark needs to be re-run: %s (exec_idx=%s)" %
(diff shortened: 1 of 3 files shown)


krun/util.py
```diff
@@ -248,7 +248,7 @@ def check_and_parse_execution_results(stdout, stderr, rc, config,
     if json_exn or rc != 0:
         # Something went wrong
         rule = 50 * "-"
-        err_s = ("Benchmark returned non-zero or emitted invalid JSON.\n")
+        err_s = ("Benchmark '%s' returned non-zero or emitted invalid JSON.\n" % key)
```
vext01, line 251: This hunk is the crux of this PR. The rest is updating call-sites.

krun/scheduler.py
```diff
@@ -383,7 +383,7 @@ def run(self, mailer, dry_run=False):
             try:
                 self.sched.platform.check_throttle_counts(self.sched.manifest)
                 measurements = util.check_and_parse_execution_results(
-                    stdout, stderr, rc, self.sched.config, instrument=vm_def.instrument)
+                    stdout, stderr, rc, self.sched.config, self.key, instrument=vm_def.instrument)
```
ltratt, line 386: I might be missing out on something, but this is only the assignment of `key` to a field -- nothing seems to read from it?

vext01: It's passing `self.key` as a function argument so that we can print it in the log.

ltratt: I'm an idiot.

vext01: Don't sweat it. Functions with that many args become hard to scan :P

PR #402, opened 2020-10-26
Add support for Som++ VM

jacob-hughes:


krun/vm_defs.py
```diff
@@ -757,7 +757,11 @@ def run_exec(self, entry_point, iterations, param, heap_lim_k,
         script_path = self._get_benchmark_path(benchmark, entry_point,
                                                force_dir=force_dir)
 
-        args = [self.vm_path, '--cp', ':'.join(self.classpath)] + self.extra_vm_args + \
+        cp = '--cp'
+        if self.vm_path.endswith('SOM++'):
```
ltratt, line 761: I suspect that yksom is the odd one out here, so maybe it would be better as:

```python
if self.vm_path.endswith("yksom"):
    cp = "--cp"
else:
    cp = "-cp"
```

?

jacob-hughes: I can do that. This is a hack really. I think the "proper" fix is to specify a separate yksom VM def, but this feels a bit overkill for such a small detail. What do you think?

ltratt: I'm all for the hack above, at least for the time being!

jacob-hughes: Fixed: 8278ab4d72cb0695625173902a680444ec644756

ltratt:
Please squash.

jacob-hughes:
Squashed

jacob-hughes:
Think this one has to be done by hand

ltratt:
Mea culpa!

ltratt (approved):

